PR #1, opened 2022-03-16
Set up beginning CI and add YAML check script

CoryMartin-NOAA:
This PR adds:
- pycodestyle CI test
- unit test for YAML checker
- YAML checker
- basic cmake structure


guillaumevernieres (commented):
That was quick @CoryMartin-NOAA ! 
Looks good to me. Could the python norm check be added or moved to the list of ctests?

CoryMartin-NOAA:
> That was quick @CoryMartin-NOAA ! Looks good to me. Could the python norm check be added or moved to the list of ctests?

@guillaumevernieres of course! It's probably good to keep a separate norms test for quick edits, but we can add a ctest for the norms also

guillaumevernieres:
> > That was quick @CoryMartin-NOAA ! Looks good to me. Could the python norm check be added or moved to the list of ctests?
> 
> @guillaumevernieres of course! It's probably good to keep a separate norms test for quick edits, but we can add a ctest for the norms also

It works on my desktop ... Going to approve, that norm check can be implemented later.

guillaumevernieres (dismissed):
:+1:  @CoryMartin-NOAA

test/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+add_test(NAME test_check_yaml_keys
+         COMMAND ${Python3_EXECUTABLE} ${CMAKE_SOURCE_DIR}/ush/check_yaml_keys.py ${CMAKE_SOURCE_DIR}/test/testinput/check_yaml_keys_ref.yaml ${CMAKE_SOURCE_DIR}/test/testinput/check_yaml_keys_test.yaml
```
aerorahul: Looks good. Though, I would suggest using `${PROJECT_SOURCE_DIR}` instead.  The use of which will become clear when GDASApp is included as a project inside another project.

Same for `${PROJECT_BINARY_DIR}`

aerorahul (approved):
LGTM

guillaumevernieres (approved):
SLGTM!

PR #3, opened 2022-03-18
Add build script, bundle CMakeLists.txt and modify .gitignore files

CoryMartin-NOAA:
This PR adds in from JEDI-T2O/UFS-DA modulefiles, scripts, and the bundle file for building the GDASApp with JEDI repositories. This will include all repos needed for both FV3-JEDI and SOCA. 

NOTE: the modulefiles will need modified soon once there are python virtual environments set up with the correct R2D2 hashes.

PR #5, opened 2022-03-21
Build and Install

aerorahul:
This PR:
- builds the GDASApp as a single package (building in PR #3)
- Does not conform to NCO vertical structure due to the way `ecbuild_bundle` works.  This 0ec3fb4 tried that, but failed due to `jedi-cmake` not being found.  `jedicmake` is not an ecbuild project.
- Updates the `build.sh` to single handedly build and install the entire applicaiton.  These should ideally be instructions in a `README.md`.  Usage:  `./build.sh -h`
- updates `.gitignore` to ignore a few more files.


CMakeLists.txt
```diff
@@ -1,10 +1,149 @@
...
+
+# Initialize bundle
+# -----------------
+ecbuild_bundle_initialize()
+
+# Build bundle source code.
+if(BUILD_GDASBUNDLE)
+
+# jedi-cmake
+  ecbuild_bundle( PROJECT jedicmake GIT "https://github.com/jcsda/jedi-cmake.git" BRANCH develop UPDATE )
```
CoryMartin-NOAA, line 45: Is there a different way to include `jedicmake` that won't lead to the error? Can we just do `add_subdirectory` or something like that?

aerorahul: I tried `add_subdirectory(src)` and then retain all this in `src/CMakeLists.txt`.  See https://github.com/NOAA-EMC/GDASApp/commit/0ec3fb472a7f77abd1a376a40ee6da1dbd1349ce .  But, I couldn't get it to work. 
This will need a resolution eventually, since NCO will not want to see all the bundle specific repos at the top-level.

guillaumevernieres:
Looks good @aerorahul , thanks. Testing on orion and my desktop, I'll finish the review after the build test have completed.

build.sh
```diff
@@ -0,0 +1,102 @@
...
+set +x
+
+# Build
+echo "Building ..."
+set -x
+make -j ${BUILD_JOBS:-6} VERBOSE=$BUILD_VERBOSE
+set +x
+
+# Install
+if [[ -n ${INSTALL_PREFIX:-} ]]; then
```
CoryMartin-NOAA, line 95: Only comment here, in the scope of global-workflow, will the workflow build script be responsible then for assigning an install prefix?

guillaumevernieres (approved):
:+1:

build.sh
aerorahul: yes, I think that would be the case.

CoryMartin-NOAA (approved):

PR #6, opened 2022-03-21
Update Orion modulefile to use new python venv

CoryMartin-NOAA:
Update the Orion modulefile to 1) load each library/module explicitly rather than the JEDI metamodule and 2) load a GDASApp python virtual environment.

A later PR will do the same for Hera once I have set up the venv there.

@guillaumevernieres please ensure when these modules are loaded that soca r2d2 scripts work as intended. I *think* the correct commits were checked out by pip when the env was installed.

### DIFF
--- a/modulefiles/GDAS/orion.lua
+++ b/modulefiles/GDAS/orion.lua
@@ -1,4 +1,4 @@
--- NOAA HPC Orion Modulefile for UFS-DA
+-- NOAA HPC Orion Modulefile for GDASApp 
 help([[
 ]])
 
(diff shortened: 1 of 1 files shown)


modulefiles/GDAS/orion.lua
```diff
@@ -11,19 +11,50 @@ setenv('JEDI_OPT', jedi_opt)
...
+load("nccmp/1.8.7.0")
+load("pio/2.5.1-debug")
+
+load("ecbuild/ecmwf-3.6.1")
+load("eckit/ecmwf-1.16.0")
+load("fckit/ecmwf-0.9.2")
+load("atlas/ecmwf-0.24.1")
+
+load("hpc")
+load("miniconda3")
```
aerorahul, line 45: Think this should be `hpc-miniconda3/4.6.14`

guillaumevernieres (approved):
Tested the the r2d2 install with soca, seems fine.

CoryMartin-NOAA (commented):
This is a test adding a comment from Orion using GitHub CLI tools

PR #14, opened 2022-03-23
Add automated CI testing on Hera

CoryMartin-NOAA:
Closes #10 

This is building on top of PR #12  but adds Hera specific things.
Doing this now for testing since Orion is down today.

PR #19, opened 2022-03-23
Add in ufsda python tools and example j-jobs/ex-scripts

CoryMartin-NOAA:
This PR adds:
*ufsda python*
ush/ufsda python scripts for creating YAML and other cool utilities. This name may change in the future, and I do expect eventually it to live outside of this repo (to share with RRFS, HAFS, 3DRTMA, etc.) but is fine here for now.

Note: for now, this is just some dummy cmake to copy the files over, we will want to install this as a python library eventually (possibly even as part of the pyvenv soon once things have stabilized).

*sample j-jobs and ex-scripts*
I've basically copied directly some sample j-jobs and ex-scripts from our old UFS-DA to here, purely as examples for @guillaumevernieres . They will not work as-is because of old path names, missing config, etc. 

One other thing to note, would be the necessary changes needed to drive these from a high level script that would replace the configuration stored in global-workflow. All of these things (such as `CDATE` or `GESDIR` would need to be set by the driver/test script if not called by the workflow.

Another note: this builds off of PR #14


guillaumevernieres (dismissed):
Thanks @CoryMartin-NOAA

CoryMartin-NOAA:
ok @guillaumevernieres I propose we merge this in so you have the python tools and I will open a separate PR once I've solved the case of the disappearing test files.

guillaumevernieres (approved):

PR #22, opened 2022-03-25
Add action to add green checkmark when passed label is applied

CoryMartin-NOAA:


CoryMartin-NOAA:
@guillaumevernieres @aerorahul do you think this is necessary/useful or overkill? This adds a checkmark or X based on label, a 'poor man's GitHub CI' approach.

aerorahul:
You seem to be having too much fun with this @CoryMartin-NOAA

aerorahul (approved):

PR #32, opened 2022-03-28
Add ioda-converters to build and add test for BUFR2IODA

### LINKED ISSUE #31: Add ioda converters to build and add a BUFR2IODA test
Need ioda converters built to read BUFR files.

Also add a new test to generate ADPSFC IODA file from BUFR.

CoryMartin-NOAA:
Closes #31

This PR adds the `iodaconv` repo to the full GDASApp bundle build.

Also, a new unit test is added if the bundle is built that will test the BUFR2IODA.x software and convert a BUFR file to IODA netCDF format.

### DIFF
--- a/test/testinput/bufr_adpsfc.yaml
+++ b/test/testinput/bufr_adpsfc.yaml
@@ -1,8 +1,3 @@
-# (C) Copyright 2020 NOAA/NWS/NCEP/EMC
-#
-# This software is licensed under the terms of the Apache Licence Version 2.0
-# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
-
 observations:
   - obs space:
       name: bufr
--- a/test/testinput/bufr_adpsfc.yaml
+++ b/test/testinput/bufr_adpsfc.yaml
@@ -2,7 +2,7 @@ observations:
   - obs space:
       name: bufr
 
-      obsdatain: "./testinput/gdas.t06z.adpsfc.tm00.bufr_d"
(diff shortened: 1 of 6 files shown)


aerorahul (approved):
LGTM

test/testinput/bufr_adpsfc.yaml
```diff
@@ -0,0 +1,168 @@
+# (C) Copyright 2020 NOAA/NWS/NCEP/EMC
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
```
aerorahul: We are under LGPL license.
Not everything copy/paste is correct 😉

aerorahul (changes requested):
looks good except the copy paste error.

test/testinput/bufr_adpsfc.yaml
CoryMartin-NOAA: I'll do you one better and get rid of all of it.

aerorahul: Y E S

aerorahul (dismissed):
LGTM.
I cannot comment on the yaml or the test, but they look familiar and correct.

CoryMartin-NOAA:
The tests 'pass' but no IODA file is created... This is a BUFR2IODA issue. Not sure why it isn't reading the symlink but will copy the file instead (could be too many layers of links).

guillaumevernieres (approved):
:+1:

Issue #33, opened 2022-03-28
Develop an executable python script to create YAML files for JEDI executables


CoryMartin-NOAA:
We need a generic, executable python tool that can create YAML from environment variables and/or an input YAML file for use in JEDI executables.

Something like this:
`genJEDIyaml --yaml $PSLOT.yaml`

Things can either be set from environment variables:
`export CDATE=2022032818`
or in that YAML:
`CDATE: 2022032818`

I suspect the best option is to read env first, and then read YAML (so the YAML can overwrite if need be)

Additionally, everything called in the main script should also be callable from a python function and passed a `dict` rather than read a YAML.

Thoughts on features, design, etc. @RussTreadon-NOAA @guillaumevernieres?

CoryMartin-NOAA:
Starting work on https://github.com/NOAA-EMC/GDASApp/tree/feature/yamlgen

Current idea would be:
`genYAML --yaml myexp.yaml --output 3dvar.yaml`
Is `yaml` too generic? Should I say `input`? This YAML would be optional and just overwrite certain ENV vars.

Edit: changed yaml to config:
`genYAML --config myexp.yaml --output 3dvar.yaml`

guillaumevernieres:
> Starting work on https://github.com/NOAA-EMC/GDASApp/tree/feature/yamlgen
> 
> Current idea would be: `genYAML --yaml myexp.yaml --output 3dvar.yaml` Is `yaml` too generic? Should I say `input`? This YAML would be optional and just overwrite certain ENV vars.
> 
> Edit: changed yaml to config: `genYAML --config myexp.yaml --output 3dvar.yaml`

:+1:  on ``--config`` , more meaningful that ``--yaml`. I like ``--input`` as well.
is ``myexp.yaml`` a template?

guillaumevernieres:
Or given, a set of necessary keys and a type of application we want to use (jedi or otherwise), generate the yaml for that?
``genYAML --config myexp.yaml --type 3dvar --output 3dvar.yaml``

CoryMartin-NOAA:
after chatting with @aerorahul, we think it's best to stick with a template approach for now.

`genYAML --config myexp.yaml`

myexp.yaml:
```
template: /path/to/3dvar.yaml
output: /path/to/my/exp/3dvar_out.yaml
config:
  user_var1: foo
  user_var2: bar
```

Anything in config can replace placeholders in the templates.

CoryMartin-NOAA:
Closed through #35

PR #34, opened 2022-03-28
Add a prep script and associated test for the marine app

### LINKED ISSUE #28: Test the marine/soca prep step as part of the ci testing

guillaumevernieres:
## Description

This PR adds the `exgdas_global_marine_analysis_prep.py`  (shamelessly copied from its atmospheric counterpart) and dependencies. 
This script does the following:
- setup R2D2 from the user's configuration
- fetch obs for the DA window from a specified R2D2 database
- stage the backgrounds required by the pseudo model
- generate the `soca` variational yaml file. Testing the validity of the `yaml` file will be subject to a future PR.

## Features added
- A skeleton of an R2D2 setup step, which will require more options in the imediate future. See here: https://github.com/NOAA-EMC/GDASApp/blob/21e09856b3e0ac807535bc1bfcf52257569edd83/ush/ufsda/r2d2.py#L14
Along with this, some refactoring of the R2D2 obs fetch was done 
- Use of the `ufsda.r2d2.store` for obs (instead of defining a separate obs store function within the script)
- Storing of the MOM6 background in R2D2, along with it's associated ctest
- Added a ctest for `exgdas_global_marine_analysis_prep.py`. It is currently using `ecbuild_add_test` but I'm open to reverting back to a pure cmake implementation if it makes sence. This test (is intended to) simulates the global workflow environment 

## Testing
Tested on my laptop! (no, not the brick ...)

If you want to test that branch of your own, you do not need to build the bundle (yet):
```console
./build.sh -c -DBUILD_GDASBUNDLE=OFF
```

added ctests shown below:
```console
...
      Start  8: test_gdasapp_soca_obsdb
 8/10 Test  #8: test_gdasapp_soca_obsdb ..............   Passed    0.19 sec
      Start  9: test_gdasapp_soca_bkgdb
 9/10 Test  #9: test_gdasapp_soca_bkgdb ..............   Passed    0.22 sec
      Start 10: test_gdasapp_soca_ana_prep
10/10 Test #10: test_gdasapp_soca_ana_prep ...........   Passed    0.22 sec

100% tests passed, 0 tests failed out of 10

Label Time Summary:
gdasapp    =   0.64 sec*proc (3 tests)
script     =   0.64 sec*proc (3 tests)
```

## Issue addressed
- closes #28


parm/templates/stage.yaml
```diff
@@ -0,0 +1,18 @@
+window begin: '{{window_begin}}'
```
CoryMartin-NOAA, line 1: Now that we aren't beholden to the JCSDA EWOK, I think we want to move away from so many 'template' YAMLs. For stage, I think we should do most/all of it in a python script. See #33

test/soca/CMakeLists.txt
```diff
@@ -11,3 +11,79 @@ ecbuild_add_test(
...
+endforeach(FILENAME)
+
+# install
+install(FILES ${test_input}
+        DESTINATION "test/testinput/")
+
+# Set runtime environement variables
+# ----------------------------------
+set( CDATE 2018041512 )        # Analysis time
+set( GDATE 2018041509 )        # What is that?
```
CoryMartin-NOAA: guess date

test/soca/CMakeLists.txt
```diff
@@ -11,3 +11,79 @@ ecbuild_add_test(
...
+set( assim_freq 6 )            # DA window
+set( OBS_LIST "adt_j3 sst_noaa19_l3u icec_EMC icefb_GDR temp_profile_fnmoc salt_profile_fnmoc" )
+set( COMOUT ${CMAKE_CURRENT_BINARY_DIR}/3dvar )
+set( COMIN_OBS ${CMAKE_CURRENT_BINARY_DIR}/obs/r2d2-shared )
+set( COMIN_GES ${CMAKE_CURRENT_BINARY_DIR}/bkg )  # TODO: stage the background somewhere
```
CoryMartin-NOAA: @RussTreadon-NOAA and I have been talking about this offline, I think we have decided we want R2D2 to stage things to the COMROT directory as part of the prep step.

test/soca/CMakeLists.txt
```diff
@@ -11,3 +11,79 @@ ecbuild_add_test(
...
+set( CASE "12345" )                               # TODO: No idea what this is, figure it out!
```
CoryMartin-NOAA: C768 for ops, case is 'FV3 resolution'

CoryMartin-NOAA: A lot of these things will be redundant once I make a more generic YAML creator now that we don't need to follow EWOK.

test/soca/CMakeLists.txt
```diff
@@ -11,3 +11,79 @@ ecbuild_add_test(
...
+set( CASE_ENKF "123456")                          # TODO: No idea what this is, figure it out!
+set( LEVS "75" )                                  # TODO: I can sort of guess ...
+set( OBS_YAML ${CMAKE_CURRENT_BINARY_DIR}/testinput/exp_obs.yaml ) # TODO: Is that the list of obs
```
CoryMartin-NOAA: This is a list of YAML files that contain all the obs, qc, etc. stuff that will be concatenated together. The old prep job would use the EWOK yamls to get the `obsdatain` path, but we could agree to some convention and just make it a list of ob types to simplify things

test/soca/CMakeLists.txt
```diff
@@ -11,3 +11,79 @@ ecbuild_add_test(
...
+                                                                   #       for the experiment?
+set( HOMEgfs ${PROJECT_SOURCE_DIR} )                                      # TODO: Where the bundle src is?
```
CoryMartin-NOAA: HOMEgfs is the workflow home; the GDASApp would be $HOMEgfs/sorc/gdasapp.fd

CoryMartin-NOAA (commented):
I think it's clear we need a more generic YAML creator ASAP. Too bad I have 4.5 hours of new hire training this afternoon...

guillaumevernieres:
Don't waste your time looking at the draft PR just yet. I still need to sort out a few issues.

test/CMakeLists.txt
```diff
@@ -27,6 +27,8 @@ list(APPEND test_data
   ${CMAKE_CURRENT_BINARY_DIR}/testdata/prof.nc
   ${CMAKE_CURRENT_BINARY_DIR}/testdata/icec.nc
   ${CMAKE_CURRENT_BINARY_DIR}/testdata/icefb.nc
+  ${CMAKE_CURRENT_BINARY_DIR}/testdata/MOM.res.nc
+  ${CMAKE_CURRENT_BINARY_DIR}/iced.2019-04-15-43200.nc
```
guillaumevernieres: testdata/.....

test/soca/CMakeLists.txt
```diff
@@ -12,3 +12,109 @@ ecbuild_add_test(
...
+endforeach(FILENAME)
+
+# install
+install(FILES ${test_input}
+        DESTINATION "test/testinput/")
+
+# Set runtime environement variables
+# ----------------------------------
+set( CDATE 2018041512 )        # Analysis date
+set( GDATE 2018041509 )        # Guess date ... Still ... what's that?
```
guillaumevernieres: gdate should be 2018041506
gdate=cdate-assim_feq

test/soca/CMakeLists.txt
```diff
@@ -12,3 +12,109 @@ ecbuild_add_test(
...
+set( assim_freq 6 )            # DA window
+set( COMOUT ${CMAKE_CURRENT_BINARY_DIR}/3dvar )
+set( COMIN_OBS ${CMAKE_CURRENT_BINARY_DIR}/obs/r2d2-shared )
+set( COMIN_GES ${CMAKE_CURRENT_BINARY_DIR}/bkg )  # Backgrounds from previous forecast
+set( CASE "12345" )                               # TODO: No idea what this is, figure it out!
```
guillaumevernieres: CASE should be replaced by OCNRES (400)

test/soca/CMakeLists.txt
```diff
@@ -12,3 +12,109 @@ ecbuild_add_test(
...
+set( CASE_ENKF "123456")                          # TODO: No idea what this is, figure it out!
```
guillaumevernieres: CASE_ENFK="C192" # resolution of emem of enkf

test/soca/testinput/exp_obs.yaml
```diff
@@ -0,0 +1,29 @@
+observations:
```
guillaumevernieres: remove that file

ush/ufsda/r2d2.py
```diff
@@ -22,14 +53,15 @@ def store(config):
...
     for time in times:
         year = Hour(time).format('%Y')
         month = Hour(time).format('%m')
         day = Hour(time).format('%d')
         hour = Hour(time).format('%H')
         inputs['date'] = time
-        window_begin = Hour(time) - DateIncrement(f'PT{assim_freq/2}H')
+        # TODO: Sort it out with Cory? Below doesn't seem right, and doesn't
+        #       work with the marine obs
+        window_begin = Hour(time)  # - DateIncrement(f'PT{assim_freq/2}H')
```
guillaumevernieres: remove L64 and L67 ... and make it work!

ush/ufsda/stage.py
```diff
@@ -168,9 +168,9 @@ def obs(config):
             type='ob',
             provider=config['r2d2_obs_src'],
             experiment=config['r2d2_obs_dump'],
-            date=config['window begin'],
+            date=config['r2d2 window begin'],
```
guillaumevernieres: should not be user defined

parm/soca/obs/config/adt_j3.yaml
```diff
@@ -0,0 +1,11 @@
+obs space:
+  name: adt_j3
+  obsdatain:
+    obsfile: $(experiment_dir)/{{current_cycle}}/adt_j3.{{window_begin}}.nc4
```
CoryMartin-NOAA: FYI I'm not following the 'ewok' convention for these paths/filenames, suggest you don't do so either. But it's up to you.

parm/templates/stage.yaml
CoryMartin-NOAA: This comment still stands :-) but yes I know it falls on me to get feature/cycling merged in to develop

scripts/exgdas_global_marine_analysis_prep.py
```diff
@@ -0,0 +1,93 @@
...
+import os
+import sys
+import yaml
+import glob
+import dateutil.parser as dparser
+
+# get absolute path of ush/ directory either from env or relative to this file
+my_dir = os.path.dirname(__file__)
+my_home = os.path.dirname(os.path.dirname(my_dir))
+ufsda_home = os.path.join(os.getenv('HOMEgfs'), 'sorc', 'ufs_da.fd', 'UFS-DA')
```
CoryMartin-NOAA, line 30: A lot of these paths are not valid anymore, but yes, again, on me to fix. This is mainly a reminder that I need to fix develop with feature/cycling ASAP after this is merged.

scripts/exgdas_global_marine_analysis_prep.py
```diff
@@ -0,0 +1,93 @@
...
+import ufsda
+
+
+def gen_bkg_list(bkg_path='.', file_type='MOM', yaml_name='bkg.yaml'):
+    # generate a YAML of the list the backgrounds for the pseudo model
+    # TODO (Guillaume): Move somewhere else ...
+    files = glob.glob(bkg_path+'/*'+file_type+'*')
+    files.sort()
+    bkg_list = []
+    for bkg in files:
```
CoryMartin-NOAA, line 46: can SOCA do the date templating that FV3 JEDI can do? If so, this is not needed, right?

test/soca/CMakeLists.txt
```diff
@@ -12,3 +12,108 @@ ecbuild_add_test(
...
 )
+
+# Prepare backgrounds
+# -------------------
+# 1 - create fake 3 hourly restarts under RESTART
+# 2 - store them with R2D2
+set( TESTDATA ${PROJECT_BINARY_DIR}/test/testdata )
+file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/bkg)
+file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/bkg/RESTART)
+ecbuild_add_test(
```
CoryMartin-NOAA: I don't think we should be using ecbuild, see https://github.com/NOAA-EMC/GDASApp/blob/develop/test/atm/CMakeLists.txt on how to use `add_test` and `set_tests_properties`

test/soca/CMakeLists.txt
```diff
@@ -12,3 +12,108 @@ ecbuild_add_test(
...
+           ${PROJECT_BINARY_DIR}/test/soca/testinput/${filename} )
+endforeach(FILENAME)
+
+# install
+install(FILES ${test_input}
+        DESTINATION "test/testinput/")
+
+# Set runtime environement variables
+# ----------------------------------
+set( CDATE 2018041512 )        # Center of current cycle date
```
CoryMartin-NOAA: Rather than all this cmake, why not just make the test run a shell script that has all of this defined in it? Wouldn't that be cleaner? What I do is just pass like `${CMAKE_CURRENT_BINARY_DIR}` to the shell script as `$1` and then define everything else inside the script. See https://github.com/NOAA-EMC/GDASApp/blob/develop/test/atm/test_convert_gsi_satbias.sh

test/soca/CMakeLists.txt
```diff
@@ -12,3 +12,108 @@ ecbuild_add_test(
...
+file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/3dvar)
+
+# Prep
+ecbuild_add_test(
+      TARGET    test_gdasapp_soca_ana_prep
+      TYPE      SCRIPT
+      COMMAND   ${Python3_EXECUTABLE}
+      ARGS      ${PROJECT_SOURCE_DIR}/scripts/exgdas_global_marine_analysis_prep.py
+      WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/3dvar
+      ENVIRONMENT
```
CoryMartin-NOAA: Yeah if this was a test wrapped in a .sh, this would make this command much simpler.

test/soca/create_bkgdb.py
```diff
@@ -0,0 +1,36 @@
...
+
+    # Setup the shared R2D2 databases
+    ufsda.r2d2.setup(r2d2_config_yaml='r2d2_config.yaml', shared_root='./r2d2-shared')
+
+    # Invent backgrounds
+    # MOM6 format: MOM.res.2019-04-15-01-00-00.nc MOM.res.2019-04-15-02-00-00.nc  MOM.res.2019-04-15-03-00-00.nc ...
+    # CICE6 format: iced.2019-04-15-00000.nc iced.2019-04-15-03600.nc  iced.2019-04-15-07200.nc ...
+
+    bkgdir = os.getenv('TESTDATA')
+    shutil.copyfile(os.path.join(bkgdir, 'MOM.res.nc'), './RESTART/MOM.res.2018-04-15-09-00-00.nc', follow_symlinks=True)
```
CoryMartin-NOAA: create a datetime object, loop with an interval of 3 hours instead of writing out 4 nearly duplicate lines?

ush/ufsda/r2d2.py
```diff
@@ -2,13 +2,44 @@
...
+import yaml
+import os
 
 possible_args = [
     'provider', 'experiment', 'database', 'type', 'file_type',
     'resolution', 'model', 'user_date_format', 'fc_date_rendering', 'tile',
 ]
 
 
+def setup(shared_root='', r2d2_config_yaml='r2d2_config.yaml'):
```
CoryMartin-NOAA, line 14: I think I'd rather have a `r2d2_config_ctest.yaml` or something like that and then set `R2D2_CONFIG=/path/to/r2d2_config_ctest.yaml` at test runtime. I think that's neater than this, especially given the issue of this approach not being viable for different platforms/filesystems for 'real' databases.

ush/ufsda/r2d2.py
```diff
@@ -22,27 +53,35 @@ def store(config):
...
         year = Hour(time).format('%Y')
         month = Hour(time).format('%m')
         day = Hour(time).format('%d')
         hour = Hour(time).format('%H')
         inputs['date'] = time
-        window_begin = Hour(time) - DateIncrement(f'PT{assim_freq/2}H')
         if r2d2_type in ['bc', 'ob']:
             if r2d2_type == 'ob':
-                inputs['date'] = window_begin
+                inputs['date'] = time
```
CoryMartin-NOAA, line 63: Note to me to fix this for atm

CoryMartin-NOAA (changes requested):
Mostly looks good but I think 1) we need to not use `ecbuild_add_test` and 2) wrap the test in a `.sh` file if setting 10+ env vars so that the cmake is cleaner.

scripts/exgdas_global_marine_analysis_prep.py
guillaumevernieres: ... Yes, I knew I was forgetting something. I'll test but it should work.

ush/ufsda/r2d2.py
guillaumevernieres: so keep a setup step that would export R2D2_CONFIG?

parm/soca/obs/config/adt_j3.yaml
guillaumevernieres: I'll keep this for now a change when I actually do something with these yaml's.

ush/ufsda/r2d2.py
CoryMartin-NOAA: the runtime env would export it, either through 'sourcing' or module load, etc.

test/soca/CMakeLists.txt
guillaumevernieres: will do

test/soca/create_bkgdb.py
guillaumevernieres: sure, might save a line or 2 ...

CoryMartin-NOAA: Probably not but cleaner :-) up to you to change if you'd like

ush/ufsda/r2d2.py
guillaumevernieres: I definitely don't like that approach and have been trying to avoid it for a while. How about I keep it that way for now and revisit when closer to actually cycling? It doesn't have to be used. We can remove it later, but I do like an explicit step to setup r2d2. The default could be coming from the runtime env.

CoryMartin-NOAA: Yeah we need to discuss this more I think, particularly with @aerorahul. You can keep it for now, yes, but I feel like defining it explicitly leads to 'separate' databases, when we really should only have 1 shared DB per platform and 1 per user for output/archival of experiments.

scripts/exgdas_global_marine_analysis_prep.py
guillaumevernieres: @CoryMartin-NOAA , turns out no, I can't do what you are doing in 4D h(x). `PSEUDO` is a factory in [fv3-jedi](https://github.com/JCSDA-internal/fv3-jedi/blob/a553b461ced0bd6810bfef4774889a6bb660f95f/src/fv3jedi/Model/pseudo/ModelPseudo.cc#L24). We should modify`PseudoModel` at some point ...

CoryMartin-NOAA: Darn. Yes, OOPS should be smart enough to look for files valid at whatever time step it is requesting based on a templated path.

guillaumevernieres:
@CoryMartin-NOAA et al. : 
I made some of the changes that you requested
1 - switched ecbuild to cmake tests
2 - added a shell script to test the prep step and clean the CMakeLists.txt
3 - realized that we need to feed in the ``window begin`` when fetching from a database that has a storage step that isn't a factor of the DA window ... Oops
4 - added a loop for the bkg, although I don't think that makes it any cleaner ... Just sayin'

parm/soca/obs/config/adt_j3.yaml
```diff
@@ -1,9 +1,9 @@
 obs space:
   name: adt_j3
   obsdatain:
-    obsfile: $(experiment_dir)/{{current_cycle}}/adt_j3.{{window_begin}}.nc4
+    obsfile: ./obs/adt_j3.{{window_begin}}.nc4
```
CoryMartin-NOAA, line 4: Inconsistency between `./obs` and `obs_out`, both are probably the same but should be consistent with path formats

parm/soca/variational/ufsda_global_ocn_3dvarfgat.yaml
```diff
@@ -1,29 +1,32 @@
...
     fields metadata: ./fields_metadata.yml
 
   model:
     name: PseudoModel
-    tstep: PT1H
+    tstep: PT3H
     state variables: &model_vars [socn, tocn, ssh, hocn, uocn, vocn, mld, layer_depth]
-    states: $<< ${BKG_LIST}
+    basename: ./bkg/
+    ocn_filename: MOM.res.%yyyy-%mm-%dd-%hh-00-00.nc
```
CoryMartin-NOAA: I thought you said this didn't work?

parm/soca/variational/ufsda_global_ocn_3dvarfgat.yaml
```diff
@@ -54,9 +57,6 @@ cost function:
         dsdzmin: 3.0e-6
         dtdzmin: 1.0e-6
         nlayers: 10
-        dcdt:
```
CoryMartin-NOAA: You don't need this mask?

CoryMartin-NOAA (commented):
Couple of small questions/comments before approval.

CoryMartin-NOAA (dismissed):

CoryMartin-NOAA (approved):

PR #35, opened 2022-03-30
Add modified YAML generation tool, new unit tests, and start to YAML templates

CoryMartin-NOAA:
This is a lengthy pull request, so apologies in advance :-)

Two new, one modified unit tests:

- test_gdasapp_generate_yaml_func has replaced test_gdasapp_generate_yaml; it is to test the generation of YAML from python by creating a dictionary and calling the necessary function
- test_gdasapp_genYAML_prep just uses `cat` to create a sample input YAML file such as a developer (or workflow) would create that contains high level options
- test_gdasapp_genYAML_run uses the output YAML from the previous test to run `genYAML --config file/yaml` and produce a YAML file.

There are three keys in this input YAML file:
```
template:
output:
config:
  var1:
  var2:
  var3:
  ...
```
`template`: path to high level template YAML file, this should match what the output is expected to look like
`output`: path to output YAML file after template, concatenation, includes, etc. are done
`config`: dict of additional variables to configure for this (intended to replace templated variables in the templates)

New ush files:
`genYAML`: exe script that is intended to be called from the command line; calls:
`ufsda.yamltools`: mostly pulled from ufsda/ufs_yaml, but made a bit more generic (no hard requirements for certain variables)

The YAML files:
[parm/atm/berror/staticb_bump.yaml](https://github.com/NOAA-EMC/GDASApp/compare/feature/yamlgen?expand=1#diff-47bde31346da4ab2bc7bd65a304d6d76595e37d93b4640d1c1c9bada9b926e11)
Copied from JEDI-T2O from @RussTreadon-NOAA, all that is changed here is `datadir: &staticb_dir $(fv3jedi_staticb_dir)`

[parm/atm/common/c384.yaml](https://github.com/NOAA-EMC/GDASApp/compare/feature/yamlgen?expand=1#diff-2d91d3180266a0dc731357184fee060de7bcd3ee68406050f4381f44ce620841)
The idea here is to have a YAML file for each resolution to define things like npx, npy, layout, etc.

[parm/atm/common/paths.yaml](https://github.com/NOAA-EMC/GDASApp/compare/feature/yamlgen?expand=1#diff-0805fcfde47c563aed6abb6b865c0e28d6ebc1046976cc322e58db8296876888)
The intention was to have a place to define paths that are used in other places but may not need to be user configurable

[parm/atm/geom/analysis.yaml](https://github.com/NOAA-EMC/GDASApp/compare/feature/yamlgen?expand=1#diff-58b20939680f00515c73f38e39c6e762f8ee5002283e3c4400d605a26445ddb4)
This (and the background) are included in the variational DA template

*obs*

Obs are split into config and lists. First config, the traditional QC/UFO YAMLs:

[parm/atm/obs/config/amsua_n19.yaml](https://github.com/NOAA-EMC/GDASApp/compare/feature/yamlgen?expand=1#diff-e0982f3f0058aa1e0f577031f702736834c405ffe9a8611b880bddd216d392a6) and [parm/atm/obs/config/sondes.yaml](https://github.com/NOAA-EMC/GDASApp/compare/feature/yamlgen?expand=1#diff-597525375d020ada2feb558a183f4614f21e3349f4786e31dfa0349c24ff8db7)
Copied from EWOK, just changed some paths/names. We probably need discussion on these variable names such as 
` obsfile: $(OBS_DIR)/$(OBS_PREFIX).sondes.$(OBS_DATE).nc4`
Ideally these YAMLs, the paths are as generic as possible so that they can work in all use cases (cycled, single analysis, etc.)

For lists:
[parm/atm/obs/lists/gdas_prototype.yaml](https://github.com/NOAA-EMC/GDASApp/compare/feature/yamlgen?expand=1#diff-c36a7df4f58f1b13d0ee079e3a477a8ec130c4100570b7f9bb3997487851dc94)
This is to have different lists of UFO YAML files. So that we can control (such as version or date) what obs are to be used.

[parm/atm/variational/3dvar_dripcg.yaml](https://github.com/NOAA-EMC/GDASApp/compare/feature/yamlgen?expand=1#diff-dfc4f9b321af6f5ceae1698ae85936a90bd2d1e6fe990b74526c12c9fd762c0c)
This is the output YAML template. This is what the final product will resemble, but this also can include templated variables and other YAML files to 'include'

I've marked this a PR, not draft, but would like some discussion before I mark it for final testing.


CoryMartin-NOAA:
One other thing to note:
in SOLO:
`${var}` means it is an env var
`$(var)` means it comes from YAML/config dictionary

guillaumevernieres (dismissed):
Looks good.

test/CMakeLists.txt
```diff
@@ -37,8 +37,7 @@ install(FILES ${test_data}
 list(APPEND test_input
   ${PROJECT_SOURCE_DIR}/test/testinput/check_yaml_keys_ref.yaml
   ${PROJECT_SOURCE_DIR}/test/testinput/check_yaml_keys_test.yaml
-  ${PROJECT_SOURCE_DIR}/test/testinput/generate_yaml.yaml
-  ${PROJECT_SOURCE_DIR}/test/testinput/generate_yaml_include.yaml
+  ${PROJECT_SOURCE_DIR}/test/testinput/genYAML_example.yaml
```
guillaumevernieres: Remove that line, that file doesn't exist and isn't needed. 
or
remove the test and script that generates it and add the template yaml?
Your choice!

CoryMartin-NOAA: Yep, thanks, that is an erroneous add, removed.

parm/atm/common/c384.yaml
```diff
@@ -0,0 +1,9 @@
+io_layout_x: 1
```
RussTreadon-NOAA: I do not know how this file will be used.  Do we want all the parameters in this file hardwired or should some be variables?   For example, the anl and ges np variables won't be the same for a dual resolution setup (e.g., operational gdas).

parm/atm/geom/background.yaml
```diff
@@ -0,0 +1,18 @@
...
+  field table filename: '$(fv3jedi_fix_dir)/field_table'
+akbk: '$(fv3jedi_fix_dir)/akbk.nc4'
+# input.nml
+layout:
+  - $(layout_x)
+  - $(layout_y)
+io_layout:
+  - $(io_layout_x)
+  - $(io_layout_y)
+npx: $(npx_ges)
```
RussTreadon-NOAA: This file is identical to parm/atm/geom/analysis.yaml except for npx and npy.   The analysis yaml references _anl variables while the background yaml references _ges variables.   Can we combine these two yamls into one with a python script inserting them in the correct location with the correct values in the monolithic yaml passed to fv3jedi_var.x?    The motivation for this is to reduce redundancy.

parm/atm/obs/config/amsua_n19.yaml
```diff
@@ -0,0 +1,323 @@
...
+  name: amsua_n19
+  obsdatain:
+    obsfile: $(OBS_DIR)/$(OBS_PREFIX).amsua_n19.$(OBS_DATE).nc4
+  obsdataout:
+    obsfile: $(DIAG_DIR)/diag_amsua_n19_$(OBS_DATE).nc4
+  simulated variables: [brightness_temperature]
+  channels: &amsua_n19_channels 1-15
+obs operator:
+  name: CRTM
+  Absorbers: [H2O,O3,CO2]
```
RussTreadon-NOAA: Does CO2 work in the variational analysis?   It didn't before.

parm/atm/obs/config/amsua_n19.yaml
```diff
@@ -0,0 +1,323 @@
...
+  Clouds: [Water, Ice]
+  Cloud_Fraction: 1.0
+  obs options:
+    Sensor_ID: amsua_n19
+    EndianType: little_endian
+    CoefficientPath: $(CRTM_COEFF_DIR)
+obs bias:
+  input file: $(BIAS_DIR)/$(BIAS_PREFIX).amsua_n19.satbias.$(BIAS_DATE).nc4
```
RussTreadon-NOAA, line 19: We should add a line for the updated bias correction file generated by fv3jedi_var.x.   The output bias correction file would have a date of OBS_DATE.

parm/atm/obs/config/amsua_n19.yaml
```diff
@@ -0,0 +1,323 @@
...
+  variational bc:
+    predictors:
+    - name: constant
+    - name: lapse_rate
+      order: 2
+      tlapse: &amsua_n19_tlapse $(BIAS_DIR)/$(BIAS_PREFIX).amsua_n19.tlapse.$(BIAS_DATE).txt
```
RussTreadon-NOAA, line 26: Based on comments in ufo issues tlapse is usually treated as a constant.   We can leave it as a cycled file for the time being but the file contents probably won't be evolving with time.

parm/atm/obs/config/amsua_n19.yaml
```diff
@@ -0,0 +1,323 @@
...
+      tlapse: *amsua_n19_tlapse
+    - name: emissivity
+    - name: scan_angle
+      order: 4
+    - name: scan_angle
+      order: 3
+    - name: scan_angle
+      order: 2
+    - name: scan_angle
+obs filters:
```
RussTreadon-NOAA, line 48: We may want to add a bias correction covariance section before obs filters.   This corresponds to the abias_pc file in the GSI.

parm/atm/common/c384.yaml
CoryMartin-NOAA: The idea was to have this fill in resolution-specific variables to then put in other places in the final YAML. But based on this and other comments, I am feeling that you are suggesting some sort of pre-processing script that takes `$CASE`, `$CASE_ENKF` and `$DOHYBVAR` (among others) and produce this and the analysis/background YAML sections on demand.

parm/atm/obs/config/amsua_n19.yaml
CoryMartin-NOAA: Good point, I forgot about this, I bet it still doesn't work properly without the 'linear obs operator' part added.

parm/atm/obs/config/amsua_n19.yaml
CoryMartin-NOAA: So does that require an `output file` section here? I'm not very familiar with how that works.

parm/atm/obs/config/sondes.yaml
```diff
@@ -0,0 +1,309 @@
+obs space:
+  name: sondes
+  obsdatain:
+    obsfile: $(OBS_DIR)/$(OBS_PREFIX).sondes.$(OBS_DATE).nc4
+    obsgrouping:
+      group variables: ["station_id", "LaunchTime"]
```
RussTreadon-NOAA, line 6: Is LaunchTime now in the obs files?   It was absent in early tests.

CoryMartin-NOAA: That depends on the source of the observations. Which is something we should probably think about. If using GSI ncdiags, to get LaunchTime, the GSI observer has to be re-run. Otherwise, we can just grab the operational diags and convert them.

parm/atm/obs/config/sondes.yaml
```diff
@@ -0,0 +1,309 @@
...
+     - name: air_temperature
+#    - name: virtual_temperature
+     - name: specific_humidity
+     - name: eastward_wind
+     - name: northward_wind
+   - name: SfcPCorrected
+     variables:
+     - name: surface_pressure
+     da_psfc_scheme: UKMO
+#
```
RussTreadon-NOAA, line 38: Tests with ufs_da added the prior to the obs filters section
```
linear obs operator:
  name: Composite
  components:
   - name: VertInterp
     variables:
     - name: air_temperature
     - name: specific_humidity
     - name: eastward_wind
     - name: northward_wind
   - name: Identity
     variables:
     - name: surface_pressure
```
Not sure if this is still needed.

parm/atm/obs/config/sondes.yaml
```diff
@@ -0,0 +1,309 @@
...
+  - filter: Perform Action
+    filter variables:
+    - name: air_temperature
+    action:
+      name: inflate error
+      inflation variable:
+        name: ObsErrorFactorConventional@ObsFunction
+        options:
+          test QCflag: PreQC
+          inflate variables: [air_temperature]
```
RussTreadon-NOAA, line 232: ufs_da tests add the line 
`pressure: MetaData/air_pressure
`
between PreQC and inflate.   Not sure if this is still needed.   This comment applies to other PreQC sections in this yaml.

parm/atm/obs/config/sondes.yaml
CoryMartin-NOAA: Yes, thanks for this and the other. This is a great example of why, going forward, it will be much better for us to maintain our own YAMLs and not rely on the EWOK ones.

parm/atm/obs/config/sondes.yaml
CoryMartin-NOAA: Yes, I will add this one and the other in a following commit.

parm/atm/variational/3dvar_dripcg.yaml
```diff
@@ -0,0 +1,44 @@
...
+    diagnostics:
+      departures: bkgmob1
+final:
+  diagnostics:
+    departures: anlmob
+  increment:
+    output:
+      filetype: latlon
+      filename: $(ANL_DIR)/atminc_latlon.
+    geometry: $<< $(GEOM_ANL)
```
RussTreadon-NOAA: ufs_da also includes yaml lines to write out tiled increments since we might update the model to directly read tiles.   This need not be added in this PR

RussTreadon-NOAA (commented):
Changes looks good.   More questions than suggestions in review.

parm/atm/common/c384.yaml
RussTreadon-NOAA: For the time being it's ok to set variables here.  I'm looking ahead to when we release this to developers.   They will prefer setting variables in a central high level location and then having pre-processing scripts set up lower level stuff based on their high level input.   `$CASE`, `$CASE_ENKF`, `$DOYBVAR` currently exist at the g-w level.   These and other variables are both g-w variables and GDASapp variables.   We don't want to require developers install and build g-w if they only want to run GDASapp.   We also don't want to require developers set some variables for parallels at the g-w level and other variables at the GDASapp level.   Can GDASapp setup scripts perform a hierarchal search of a dictionary to set variables which may be defined at the g-w or GDASapp level?   Maintaining such a dictionary could be problematic since we need to  keep tabs of multiple repos.

CoryMartin-NOAA: The only thing I'm really worried about is `layout_x` and `layout_y`. Those are easily set as part of `config.resources` in the g-w. Since that has to correspond with MPI tasks, is it reasonable to expect a user running the scripts to set their own `layout_x` and `layout_y` based on how many tasks they request?

For the `npx`, `npy`, etc. it should be easy enough to just pass `CASE`, `CASE_ENKF` and `DOHYBVAR` to the python `genYAML` code and let it compute all of these YAML keys.

CoryMartin-NOAA: My mistake,  I mean in `config.fv3` . So those can easily come from there. But where are those stored in the GDASApp? I guess we could have a C96.yaml that just contains `layout_x` and `layout_y` and only looks for that if `$layout_x` is not set?

RussTreadon-NOAA: Users will need to set their own `layout_x` and `layout_y` when submitting jobs at the GDASapp level.   If they are running the g-w, `config.fv3` can provide the definition.    As I have learned if `layout_x` and `layout_y` don't match with what I specify in my driver script `fv3jedi_var.x` notes this and aborts.

CoryMartin-NOAA: Ok that works for me, I will work on a solution that removes some of these redundant YAML entries and push as part of a new commit to this PR. I have an idea of how to do so that I think should work fairly well.

parm/atm/obs/config/amsua_n19.yaml
RussTreadon-NOAA: Here's what was added to a `amsua_n19.yaml` in a working copy of ufs_da

```
    - name: scan_angle
+  covariance:
+    minimal required obs number: 20
+    variance range: [1.0e-6, 10.0]
+    step size: 1.0e-4
+    largest analysis variance: 10000.0
+    prior:
+      input file: $(experiment_dir)/{{current_cycle}}/amsua_n19.{{previous_cycle}}.satbias_cov.nc4
+      inflation:
+        ratio: 1.1
+        ratio for small dataset: 2.0
+    output file: $(experiment_dir)/{{current_cycle}}/amsua_n19.{{current_cycle}}.satbias_cov.nc4
obs filters:
```

CoryMartin-NOAA:
Marking this as draft until I have a better solution for geometry YAML generation for all atm cases.

ush/ufsda/yamltools.py
```diff
@@ -0,0 +1,241 @@
...
+    # now recursively update config dict
+    config_out = update_config(config_out)
+    # clean up if specified
+    if clean and template is not None:
+        config_out = clean_yaml(config_out, config_temp)
+
+    return config_out
+
+
+def atmanl_case(config):
```
CoryMartin-NOAA, line 67: @RussTreadon-NOAA this function and the other one it calls replace the resolution specific YAML files.

This will use `CASE`, `CASE_ENKF`, `DOHYBVAR`, `layout_x` and `layout_y` from either the env or the input YAML to decide what the geometry values will be. Is this what you had in mind to replace redundant YAML files?

RussTreadon-NOAA: Thanks, @CoryMartin-NOAA , for adding this flexibility.

parm/atm/obs/config/amsua_n19.yaml
CoryMartin-NOAA: Thanks, added.

parm/atm/obs/config/sondes.yaml
RussTreadon-NOAA: Thanks for the explanation.

parm/atm/obs/config/amsua_n19.yaml
RussTreadon-NOAA: This point need not hold up this PR.   Mark as resolved and address in subsequent PRs.

parm/atm/obs/config/sondes.yaml
CoryMartin-NOAA: Done.

RussTreadon-NOAA (approved):
Thank you @CoryMartin-NOAA  for addressing all my comments.

CoryMartin-NOAA:
Thank you @RussTreadon-NOAA for the helpful review! I will merge this in now.

Issue #7, opened 2022-03-21
Add in scripts to convert and save to R2D2 sat bias files


CoryMartin-NOAA:
Need to take GSI bc files, convert them to netCDF4 format, and store them in R2D2 database. I have quick and dirty scripts somewhere for this already, but need to clean them up and add them to ush/

CoryMartin-NOAA:
Closed through #36 and #37

PR #36, opened 2022-03-31
Add driver script to convert GSI bias coefficient files for UFO

CoryMartin-NOAA:
Addresses part of #7 

This adds a new ush script, `run_satbias_conv.py` that will loop through specified cycles and convert GSI bias files (eg. `gdas.t00z.abias` and `gdas.t00z.abias_pc`) to files readable by UFO.

Usage:
`./run_satbias_conv.py --config input.yaml`

This input YAML file should contain the following:
```
start time: 2021-12-20T18:00:00Z
end time: 2021-12-21T18:00:00Z
assim_freq: 6
dump: gdas
gsi_bc_root: /path/to/gsi/rotdirs
ufo_bc_root: /path/to/output/toplevel
work_root: /path/to/toplevel/work
satbias2ioda: /path/to/build/bin/satbias2ioda.x
```

All three directories should/will have the structure within them:
`{dump}.YYYYMMDD/HH/atmos`

Still to come: the associated unit test to convert a sample GSI file, and the R2D2 test to place the converted files in R2D2

### DIFF
--- a/ush/run_satbias_conv.py
+++ b/ush/run_satbias_conv.py
@@ -16,6 +16,71 @@ logging.basicConfig(format='%(asctime)s:%(levelname)s:%(message)s',
 def run_satbias_conv(config):
     # run ioda-converters satbias converter to take GSI satbias file
     # and create UFO compatible input files
+    # get times from config
+    startTime = config['start time']
+    endTime = config['end time']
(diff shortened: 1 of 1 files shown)


guillaumevernieres (approved):
Don't know about the science, but the code look :+1:

PR #37, opened 2022-04-01
Add tests to convert sat bias files and save them to R2D2; make calls to R2D2 more generic

CoryMartin-NOAA:
Addresses the remainder of #7 

Two new tests are added (but only if the full bundle is built)

- test_gdasapp_convert_gsi_satbias will take the script added in #36 and run it with new sample data added
- test_gdasapp_store_gsi_satbias will store the output of the above test into a test R2D2 database

Additionally, a generic call to R2D2 store has been added to `ush/ufsda/r2d2.py`, and changes have been made to the SOCA test to use this generic R2D2 store function.


test/atm/test_convert_gsi_satbias.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+set -ex
```
guillaumevernieres, line 2: still need -x?

test/atm/test_store_gsi_satbias.py
```diff
@@ -0,0 +1,72 @@
...
+if __name__ == "__main__":
+    # create a dummy R2D2 config YAML
+    r2d2_config = {
+        'databases':
+            {'test': {'cache_fetch': False,
+                      'class': 'LocalDB',
+                      'root': os.path.join(os.getcwd(), 'r2d2-test')}, },
+        'fetch_order': ['test'],
+        'store_order': ['test'],
+    }
```
guillaumevernieres, line 24: couldn't that be a call to ``genyaml`` and we keep an r2d2_config.yaml template somewhere?

test/soca/create_obsdb.py
```diff
@@ -2,41 +2,15 @@
...
-                type=type,
-                experiment=experiment,
-                database=database,
-                date=date,
-                obs_type=obs_type,
-                time_window=step,
-                source_file=f'{source_dir}/{obs_type}_{year}{month}{day}.nc4',
-                ignore_missing=True,
-            )
+    ufsda.r2d2.store(config)
```
guillaumevernieres, line 13: :+1:

guillaumevernieres (dismissed):
A few minor comments/questions. Looks good, thanks @CoryMartin-NOAA .

test/atm/test_convert_gsi_satbias.sh
CoryMartin-NOAA: I thought it might be useful for debugging purposes

test/atm/test_store_gsi_satbias.py
CoryMartin-NOAA: Sure, do you think that is what should be done in this and your existing test? Or stick with this approach for now? Does it make sense for the same r2d2 db to be used for both the atm and soca tests?

CoryMartin-NOAA:
Build fails on `Hera` because `da` disk quota has been exceeded.

test/atm/test_store_gsi_satbias.py
guillaumevernieres: I think so, we have a few r2d2_config dictionaries defined here and there, it might be a good idea to consolidate how we deal with the config. I would say create an issue somewhere so we can remember to deal with it later, or change it now? Up to you. I won't block this pr for this issue.
As for the r2d2 db, yes, I do think we should have the same one.

guillaumevernieres (approved):
Looks good. Thanks @CoryMartin-NOAA .
I added a few comments that could be addressed in future pr's.

PR #41, opened 2022-04-05
Example script to store GFS RESTART backgrounds in R2D2

CoryMartin-NOAA:
This PR adds an example script that copies GFS C96 backgrounds from @RussTreadon-NOAA and puts them in my R2D2 shared database. 

It also makes some changes to the `ufsda/r2d2` python `store` general function to handle backgrounds in addition to obs and bias correction coefficients.

### DIFF
--- a/ush/r2d2/example_store_bkg.py
+++ b/ush/r2d2/example_store_bkg.py
@@ -16,6 +16,7 @@ if __name__ == "__main__":
         'end': '2021-12-21T18:00:00Z',
         'step': 'PT6H',
         'forecast_steps': ['PT6H'],
+        'file_type_list': ['fv_core.res', 'fv_srf_wnd.res', 'fv_tracer.res', 'phy_data', 'sfc_data'],
         'source_dir': '/work/noaa/stmp/rtreadon/comrot/prufsda/',
         'source_file_fmt': '{source_dir}/{dump}.{year}{month}{day}/{hour}/atmos/\
 RESTART_GES/$(valid_date).$(file_type).tile$(tile).nc',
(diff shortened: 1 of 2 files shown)


ush/r2d2/example_store_bkg.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+if __name__ == "__main__":
+    # first create dict of config for UFS RESTART files
+    config = {
+        'start': '2021-12-21T00:00:00Z',
+        'end': '2021-12-21T18:00:00Z',
+        'step': 'PT6H',
+        'forecast_steps': ['PT6H'],
+        'source_dir': '/work/noaa/stmp/rtreadon/comrot/prufsda/',
```
RussTreadon-NOAA, line 20: This will break when I eventually remove this directory.

CoryMartin-NOAA: Correct. By changing this directory path to `$ROTDIR`, and changing `RESTART_GES` to `RESTART`, this should work for any global-workflow experiment though.

ush/ufsda/r2d2.py
```diff
@@ -2,47 +2,87 @@
...
     for time in times:
         year = Hour(time).format('%Y')
         month = Hour(time).format('%m')
         day = Hour(time).format('%d')
         hour = Hour(time).format('%H')
-        for obs_type in obs_types:
-            if type == 'bc':
+        if type in ['bc', 'ob']:
+            for obs_type in obs_types:
+                if type == 'bc':
```
RussTreadon-NOAA: With several lines being identical between the if and else blocks, I wonder if we collapse this into a single `r2d2.store`.  Of course, this simply moves the if/else to where we populate the arguments sent to `rt2d.store`

CoryMartin-NOAA: You made me think of a possible alternative solution, something in python called `**kwargs`. I bet we can use this to remove the if-statements entirely.

ush/r2d2/example_store_bkg.py
RussTreadon-NOAA: OK.  As the filename explains, this is an example.

ush/ufsda/r2d2.py
CoryMartin-NOAA: the latest commit greatly reduces the amount of code in if-else blocks and still achieves the same result. Thanks for pointing me in the right direction :-)

RussTreadon-NOAA (commented):
Use of kwargs does a nice job of compressing the if block.

RussTreadon-NOAA (approved):
Looks good

PR #42, opened 2022-04-07
Stand alone script to run a single atm 3DVar analysis

CoryMartin-NOAA:
This PR adds a new ush script, `run_single_atm_var_analysis.py`.

This script does the following:

- Reads an input user configurable YAML (see below)
- Creates a config dict matching what `ush/genYAML` expects
- Creates the 3DVar YAML
- Uses R2D2 to stage backgrounds, obs, bias coeff files
- Links fix files from a fix/ directory I compiled
- Links the executable to the working directory
- Sets up all additionally needed directories
- Generates a SLURM batch script and submits the job

Along with this script, the associated functions have been added/modified to support each of the above steps.

A sample input YAML:
```
working directory: /work2/noaa/stmp/cmartin/gdas_single_test
GDASApp home: /work2/noaa/da/cmartin/GDASApp/work/GDASApp
analysis options:
  berror_yaml: /work2/noaa/da/cmartin/GDASApp/work/GDASApp/parm/atm/berror/staticb_bump.yaml
  obs_yaml_dir: /work2/noaa/da/cmartin/GDASApp/work/GDASApp/parm/atm/obs/config
  var_yaml: /work2/noaa/da/cmartin/GDASApp/work/GDASApp/parm/atm/variational/3dvar_dripcg.yaml
  obs_list: /work2/noaa/da/cmartin/GDASApp/work/GDASApp/parm/atm/obs/lists/gdas_prototype.yaml
  gdas_fix_root: /work2/noaa/da/cmartin/GDASApp/fix
  atm: true
  layout_x: 1
  layout_y: 1
  atm_window_length: PT6H
  valid_time: 2021-12-21T06:00:00Z
  dump: gdas
  case: C96
  case_enkf: C48
  dohybvar: true
  levs: 128
job options:
  machine: orion
  account: da-cpu
  queue: debug
  partition: debug
  walltime: '10:00'
  ntasks: 6
  cpus-per-task: 1
  modulepath: /work2/noaa/da/cmartin/GDASApp/work/GDASApp/modulefiles
```

To run:
```
module use /path/to/clone/modulefiles
module load GDAS/orion
export PYTHONPATH:/path/to/clone/ush:$PYTHONPATH
python run_single_atm_var_analysis.py --config sample.yaml
```

Sample working directory can be found at: `/work2/noaa/stmp/cmartin/gdas_single_test`

### DIFF
--- a/modulefiles/GDAS/orion.lua
+++ b/modulefiles/GDAS/orion.lua
@@ -53,6 +53,8 @@ local mpinproc = '-n'
 setenv('MPIEXEC_EXEC', mpiexec)
 setenv('MPIEXEC_NPROC', mpinproc)
 
+setenv('R2D2_CONFIG', '/work2/noaa/da/cmartin/GDASApp/R2D2_SHARED/config_orion.yaml')
+
 
 whatis("Name: ".. pkgName)
 whatis("Version: " .. pkgVersion)
--- a/modulefiles/GDAS/orion.lua
+++ b/modulefiles/GDAS/orion.lua
@@ -55,6 +55,7 @@ setenv('MPIEXEC_NPROC', mpinproc)
 
 setenv('R2D2_CONFIG', '/work2/noaa/da/cmartin/GDASApp/R2D2_SHARED/config_orion.yaml')
 
+execute{cmd="ulimit -s unlimited",modeA={"load"}}
 
 whatis("Name: ".. pkgName)
 whatis("Version: " .. pkgVersion)
--- a/parm/atm/berror/staticb_identity.yaml
+++ b/parm/atm/berror/staticb_identity.yaml
@@ -1,0 +1,2 @@
+covariance model: FV3JEDI-ID
+date: $(window_begin)
--- a/parm/atm/obs/config/amsua_n19.yaml
+++ b/parm/atm/obs/config/amsua_n19.yaml
@@ -44,7 +44,7 @@ obs bias:
       inflation:
         ratio: 1.1
         ratio for small dataset: 2.0
-    output file: $(BIAS_DIR)/$(OBS_PREFIX).amsua_n19.satbias.$(OBS_DATE).nc4
+    output file: $(BIAS_DIR)/$(OBS_PREFIX).amsua_n19.satbias_cov.$(OBS_DATE).nc4
 obs filters:
 - filter: BlackList
   filter variables:
--- a/parm/atm/obs/config/amsua_n19.yaml
+++ b/parm/atm/obs/config/amsua_n19.yaml
@@ -14,7 +14,7 @@ obs operator:
   obs options:
     Sensor_ID: amsua_n19
     EndianType: little_endian
-    CoefficientPath: $(CRTM_COEFF_DIR)
+    CoefficientPath: $(CRTM_COEFF_DIR)/
 obs bias:
   input file: $(BIAS_DIR)/$(BIAS_PREFIX).amsua_n19.satbias.$(BIAS_DATE).nc4
   output file: $(BIAS_DIR)/$(OBS_PREFIX).amsua_n19.satbias.$(OBS_DATE).nc4
--- a/parm/atm/obs/config/sondes.yaml
+++ b/parm/atm/obs/config/sondes.yaml
@@ -3,7 +3,7 @@ obs space:
   obsdatain:
     obsfile: $(OBS_DIR)/$(OBS_PREFIX).sondes.$(OBS_DATE).nc4
     obsgrouping:
-      group variables: ["station_id", "LaunchTime"]
+      group variables: ["station_id"]
       sort variable: "air_pressure"
       sort order: "descending"
   obsdataout:
--- a/test/genYAML_prep.sh
+++ b/test/genYAML_prep.sh
@@ -6,7 +6,8 @@ cat > testoutput/genYAML_example.yaml << EOF
 template: ${srcdir}/parm/atm/variational/3dvar_dripcg.yaml
 output: ${bindir}/test/testoutput/genYAML_output_3dvar.yaml
 config:
-  paths: $<< ${srcdir}/parm/atm/common/paths.yaml
+  BERROR_YAML: \${PARMgfs}/atm/berror/staticb_bump.yaml
+  OBS_YAML_DIR: \${PARMgfs}/atm/obs/config
   atm: true
   layout_x: '1'
   layout_y: '2'
--- a/test/test_generate_yaml.py
+++ b/test/test_generate_yaml.py
@@ -10,7 +10,8 @@ def test_yaml_gen_yaml(parm_dir):
     output_file = os.path.join(os.getcwd(), 'testoutput', 'test_yaml_gen.yaml')
     template = os.path.join(parm_dir, 'atm/variational/3dvar_dripcg.yaml')
     config_dict = {
-        'paths': f'$<< {parm_dir}/atm/common/paths.yaml',
+        'BERROR_YAML': '${PARMgfs}/atm/berror/staticb_bump.yaml',
+        'OBS_YAML_DIR': '${PARMgfs}/atm/obs/config',
         'atm': True,
         'layout_x': '1',
         'layout_y': '2',
--- a/ush/run_single_atm_var_analysis.py
+++ b/ush/run_single_atm_var_analysis.py
@@ -1,0 +1,38 @@
+#!/usr/bin/env python3
+import argparse
+import logging
+import os
+import subprocess
+import ufsda
+import yaml
+
+
+def run_atm_var_analysis(yamlconfig):
+    logging.basicConfig(format='%(asctime)s:%(levelname)s:%(message)s', level=logging.INFO, datefmt='%Y-%m-%d %H:%M:%S')
+    # open YAML file to get config
+    try:
+        with open(yamlconfig, 'r') as yamlconfig_opened:
+            all_config_dict = yaml.safe_load(yamlconfig_opened)
+        logging.info(f'Loading configuration from {yamlconfig}')
+    except Exception as e:
+        logging.error(f'Error occurred when attempting to load: {yamlconfig}, error: {e}')
+    # create working directory
+    workdir = all_config_dict['working directory']
+    try:
+        os.makedirs(workdir, exist_ok = True)
+        logging.info(f'Created working directory {workdir}')
+    except OSError as error:
+        logging.error(f'Error creating {workdir}: {error}')
+    # use R2D2 to stage backgrounds, obs, bias correction files, etc.
+    # link additional fix files needed (CRTM, fieldsets, etc.)
+    # link executable
+    # generate YAML for executable based on input config
+    # generate job submission script
+    # submit job to queue
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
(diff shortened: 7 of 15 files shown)


CoryMartin-NOAA:
Note `dohybvar=true` isn't quite correct, as it is not hybrid-3DVar, but this controls making the analysis resolution 1/2 that of the background resolution.

@FanHan-NOAA you might want to try being a beta tester here to clone, build, and run this single script, as it will help with your interpolation work for the ISDA conference.

ush/ufsda/misc_utils.py
```diff
@@ -10,3 +15,38 @@ def isTrue(str_in):
...
+
+def create_batch_job(job_config, working_dir, exe_path, yaml_path):
+    """
+    create a batch job submission shell script
+    """
+    # open up a file for writing
+    batch_script = os.path.join(working_dir, 'submit_job.sh')
+    with open(batch_script, 'w') as f:
+        f.write('#!/bin/bash\n')
+        if job_config['machine'] in ['orion', 'hera']:
```
aerorahul: I would suggest using `scheduler` instead of `machine` where these kind of decisions need to be made.

CoryMartin-NOAA: Yeah I thought about that, but wanted to make it simpler for the user. Should I do the workflow approach where scheduler is the if statement but it is determined based on a dict (`if scheduler[job_config['machine']] == 'slurm'`)?

FanHan-NOAA:
> Note `dohybvar=true` isn't quite correct, as it is not hybrid-3DVar, but this controls making the analysis resolution 1/2 that of the background resolution.
> 
> @FanHan-NOAA you might want to try being a beta tester here to clone, build, and run this single script, as it will help with your interpolation work for the ISDA conference.

Great! Thanks for this addition. I'll start testing it soon.

RussTreadon-NOAA:
@CoryMartin-NOAA :  As we review this PR shall we ignore, for the the time being, pieces specific to Hera and focus solely on Orion?

CoryMartin-NOAA:
@RussTreadon-NOAA I'm open to see what you and others think. Currently this will *only* work on Orion, as I need to stage fix files and a R2D2 database on Hera. Does it make sense to support both systems now, or focus on Orion and port when things have stabilized slightly?

RussTreadon-NOAA:
@CoryMartin-NOAA, I agree with you.  To keep things moving focus on Orion and port when GDASApp has stabilized.

modulefiles/GDAS/orion.lua
```diff
@@ -53,6 +53,9 @@ local mpinproc = '-n'
 setenv('MPIEXEC_EXEC', mpiexec)
 setenv('MPIEXEC_NPROC', mpinproc)
 
+setenv('R2D2_CONFIG', '/work2/noaa/da/cmartin/GDASApp/R2D2_SHARED/config_orion.yaml')
```
RussTreadon-NOAA, line 56: Not a request for a change, but echoing previous comments expressing need for a central location (e.g., role account) in which to store shared files.

parm/atm/berror/staticb_identity.yaml
```diff
@@ -0,0 +1,2 @@
+covariance model: FV3JEDI-ID
```
RussTreadon-NOAA, line 1: Just curious.   Is `FV3JEDI-ID` a standard JEDI variable or name for the identity model?   I can't find `FV3JEDI-ID` defined in `feature/single_anl`

CoryMartin-NOAA: This is the FV3-JEDI YAML entry for the identity **B**, see for example: https://github.com/JCSDA-internal/fv3-jedi/blob/7e9e61222e32e494482dd442d6e86d9d31098c81/test/testinput/hyb-3dvar_geos.yaml#L29

parm/atm/obs/config/amsua_n19.yaml
```diff
@@ -44,7 +44,7 @@ obs bias:
       inflation:
         ratio: 1.1
         ratio for small dataset: 2.0
-    output file: $(BIAS_DIR)/$(OBS_PREFIX).amsua_n19.satbias.$(OBS_DATE).nc4
+    output file: $(BIAS_DIR)/$(OBS_PREFIX).amsua_n19.satbias_cov.$(OBS_DATE).nc4
```
RussTreadon-NOAA, line 47: This comment is not for the output satbias_cov file but rather for the input file.   I think the input covariance file should be `$(BIAS_DIR)/$(BIAS_PREFIX).amsua_n19.satbias_cov.$(BIAS_DATE).nc4`

CoryMartin-NOAA: Yes, I opened an issue here https://github.com/JCSDA-internal/ufo/issues/1964 I am not sure about if they are supposed to be separate or not, as the converter suggests they are all in one file...

parm/atm/obs/config/sondes.yaml
```diff
@@ -3,7 +3,7 @@ obs space:
   obsdatain:
     obsfile: $(OBS_DIR)/$(OBS_PREFIX).sondes.$(OBS_DATE).nc4
     obsgrouping:
-      group variables: ["station_id", "LaunchTime"]
+      group variables: ["station_id"]
```
RussTreadon-NOAA, line 6: Removing `LaunchTime` is only necessary because this variable is not in the R2D2 obs.  If we regenerate the R2D2 obs with this variable, we should add `LaunchTime` back.  Is this correct?   If "yes", what's the long term plan for sonde data?  Will it include `LanuchTime`?   If this is where we are heading, should we redo the R2D2 database to include this varaible now?

test/test_generate_yaml.py
```diff
@@ -10,7 +10,8 @@ def test_yaml_gen_yaml(parm_dir):
     output_file = os.path.join(os.getcwd(), 'testoutput', 'test_yaml_gen.yaml')
     template = os.path.join(parm_dir, 'atm/variational/3dvar_dripcg.yaml')
     config_dict = {
-        'paths': f'$<< {parm_dir}/atm/common/paths.yaml',
+        'BERROR_YAML': '${PARMgfs}/atm/berror/staticb_bump.yaml',
```
RussTreadon-NOAA, line 13: If users want to use `staticb_identity.yaml`, is this the line they need to change?

parm/atm/obs/config/sondes.yaml
CoryMartin-NOAA: Yes, this is because currently obs come from GSI ncdiags. The operational ncdiags do not contain `LaunchTime`. To get it from ncdiags, we would have to rerun the GSI observer with a modified GSI executable. Long term, this should be coming directly from BUFR files, which should have (hopefully) `LaunchTime`.

test/genYAML_prep.sh
```diff
@@ -6,7 +6,8 @@ cat > testoutput/genYAML_example.yaml << EOF
 template: ${srcdir}/parm/atm/variational/3dvar_dripcg.yaml
 output: ${bindir}/test/testoutput/genYAML_output_3dvar.yaml
 config:
-  paths: $<< ${srcdir}/parm/atm/common/paths.yaml
+  BERROR_YAML: \${PARMgfs}/atm/berror/staticb_bump.yaml
```
RussTreadon-NOAA, line 9: If users want to run with `staticb_identity.yaml`, is this the line they need to change?

modulefiles/GDAS/orion.lua
CoryMartin-NOAA: Yes, tagging @dtkleist for awareness of a use case

RussTreadon-NOAA (commented):
Successfully installed and ran `run_single_atm_var_analysis.py` on Orion.   The description for PR #42 provides a `sample.yaml`.   Should we commit a templated `sample.yaml` to  `feature/single_anl`?

test/test_generate_yaml.py
CoryMartin-NOAA: Correct.

CoryMartin-NOAA:
> Should we commit a templated `sample.yaml` to `feature/single_anl`?

I was actually thinking the next step is to add a "regression test" that would do a nccmp on the increment file and compare it to the existing increment file and determine if the analysis still is produced. That way the regression test would hold the sample YAML file, but I can alternatively create a `parm/examples` directory or something like that?

test/test_generate_yaml.py
CoryMartin-NOAA: This is for the unit test. Most users would change it in `sample.yaml`.

parm/atm/berror/staticb_identity.yaml
RussTreadon-NOAA: Thanks, @CoryMartin-NOAA .   There's still much for me to learn.

parm/atm/obs/config/amsua_n19.yaml
RussTreadon-NOAA: OK.   Good to open [ufo #1964](https://github.com/JCSDA-internal/ufo/issues/1964) to get clarification.

parm/atm/obs/config/sondes.yaml
RussTreadon-NOAA: Got it.  It would be nice to directly read from BUFR files sooner rather than later.   Populating R2D2 with GSI ncdiags is extra work and consumes additional disk space.

RussTreadon-NOAA:
> > Should we commit a templated `sample.yaml` to `feature/single_anl`?
> 
> I was actually thinking the next step is to add a "regression test" that would do a nccmp on the increment file and compare it to the existing increment file and determine if the analysis still is produced. That way the regression test would hold the sample YAML file, but I can alternatively create a `parm/examples` directory or something like that?

What you outline is feasible.   This will provide users with an example of what `sample.yaml` looks like.

RussTreadon-NOAA (dismissed):
Thank you, @CoryMartin-NOAA , for answering my questions.   The changes look good to me. 
I see that you are working with @aerorahul regarding `scheduler` in `ush/ufsda/misc_utils.py`.   I'm more an observer than a reviewer when it comes to these details.  It's a good learning experience for me.  Thanks!

ush/ufsda/misc_utils.py
CoryMartin-NOAA: @aerorahul how is this?

CoryMartin-NOAA:
Thanks @RussTreadon-NOAA , I have pushed a new commit that dismissed your approval, if @aerorahul is happy with my new changes, I will merge this in once I have one 'approval'

ush/run_single_atm_var_analysis.py
```diff
@@ -0,0 +1,103 @@
...
+    workdir = all_config_dict['working directory']
+    try:
+        os.makedirs(workdir, exist_ok=True)
+        logging.info(f'Created working directory {workdir}')
+    except OSError as error:
+        logging.error(f'Error creating {workdir}: {error}')
+    # add your ufsda python package to path
+    ufsda_path = os.path.join(all_config_dict['GDASApp home'], 'ush')
+    sys.path.append(ufsda_path)
+    import ufsda
```
aerorahul, line 31: What's going on here? And why is this import in a function?

aerorahul (dismissed):
Changes look good.

RussTreadon-NOAA (approved):
Looks good to me.

PR #49, opened 2022-04-26
Updates to cycle GDASApp in global-workflow

RussTreadon-NOAA:
This PR is opened to request merger of the following changes from branch `feature/cycling` in fork [Russ-Treadon-NOAA/GDASApp](https://github.com/RussTreadon-NOAA/GDASApp) into the authoritative `feature/cycling`.
 * `parm/atm/obs/config/amsua_n19.yaml` - set radiance bias correction error covariance file to satbias_cov
 * `scripts/exgdas_global_atmos_analysis_prep.py` - add function to stage bias correction files
 * `scripts/exgdas_global_atmos_analysis_run.sh` - replace DOHYBVAR with CASE_ANL, trap INCPY return code, fix output cp
 * `ush/get_obs_list.py` - add satbias_cov to list of files to process
 * `ush/jediinc2fv3.py` - add lowercase t (air_temperature) as a variable to check
 * `ush/run_jedi_exe.py` - replace BIAS_DIR with BIAS_IN_DIR and BIAS_OUT_DIR, add CASE_ANL, create bc directory
 * `ush/ufsda/archive.py` - store satbias_cov
 * `ush/ufsda/misc_utils.py` - add CASE_ANL
 * `ush/ufsda/stage.py` - replace DOHYBVAR logic with direct use of CASE_ANL, add function bias_obs
 * `ush/ufsda/yamltools.py` - add case_anl, set GEOM_ANL based on case_anl

These changes have been tested via a C96L127 parallel running on Orion using g-w branch [feature/ufsda-gdasapp](https://github.com/NOAA-EMC/global-workflow/tree/feature/ufsda-gdasapp).  Expected behavior was observed.   `run_jedi_exe.py` was tested using a sample yaml file.  Expected behavior was observed.

### DIFF
--- a/ush/ufsda/archive.py
+++ b/ush/ufsda/archive.py
@@ -35,16 +35,26 @@ def atm_diags(config):
         r2d2_config['source_file_fmt'] = input_file.replace('.nc4', '_0000.nc4')
         r2d2_config['obs_types'] = [ob['obs space']['name']]
         ufsda.r2d2.store(r2d2_config)
-        # get bias files if needed
+        # store bias files
         if 'obs bias' in ob.keys():
             r2d2_config['type'] = 'bc'
             r2d2_config['provider'] = 'gsi'
             r2d2_config['start'] = config['valid_time']
             r2d2_config['end'] = config['valid_time']
+
+            # store satbias
             r2d2_config['file_type'] = 'satbias'
             target_file = ob['obs bias']['output file']
             r2d2_config['source_file_fmt'] = target_file
             ufsda.r2d2.store(r2d2_config)
+
+            # store satbias_cov
+            r2d2_config['file_type'] = 'satbias_cov'
+            target_file = target_file.replace('satbias', 'satbias_cov')
+            r2d2_config['source_file_fmt'] = target_file
+            ufsda.r2d2.store(r2d2_config)
+
+            # store tlapse
             r2d2_config['file_type'] = 'tlapse'
             target_file = target_file.replace('satbias', 'tlapse')
             target_file = target_file.replace('nc4', 'txt')
--- a/ush/ufsda/stage.py
+++ b/ush/ufsda/stage.py
@@ -133,37 +133,23 @@ def bias_obs(config):
             r2d2_config['file_type'] = 'satbias'
             target_file = ob['obs bias']['input file']
             r2d2_config['target_file_fmt'] = target_file
-            print(f"config is {config}")
-            try:
-                r2d2_config['experiment'] = config['experiment']
-                ufsda.r2d2.fetch(r2d2_config)
-            except:
-                r2d2_config['experiment'] = 'oper_gdas'
-                ufsda.r2d2.fetch(r2d2_config)
+            r2d2_config['experiment'] = config.get('experiment', 'oper_gdas')
+            ufsda.r2d2.fetch(r2d2_config)
 
             # fetch satbias_cov
             r2d2_config['file_type'] = 'satbias_cov'
             target_file = target_file.replace('satbias', 'satbias_cov')
             r2d2_config['target_file_fmt'] = target_file
-            try:
-                r2d2_config['experiment'] = config['experiment']
-                ufsda.r2d2.fetch(r2d2_config)
-            except:
-                r2d2_config['experiment'] = 'oper_gdas'
-                ufsda.r2d2.fetch(r2d2_config)
+            r2d2_config['experiment'] = config.get('experiment', 'oper_gdas')
+            ufsda.r2d2.fetch(r2d2_config)
 
             # fetch tlapse
             r2d2_config['file_type'] = 'tlapse'
             target_file = target_file.replace('satbias_cov', 'tlapse')
             target_file = target_file.replace('nc4', 'txt')
             r2d2_config['target_file_fmt'] = target_file
-            try:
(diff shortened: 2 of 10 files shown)


ush/ufsda/archive.py
```diff
@@ -35,16 +35,26 @@ def atm_diags(config):
...
             r2d2_config['end'] = config['valid_time']
+
+            # store satbias
             r2d2_config['file_type'] = 'satbias'
             target_file = ob['obs bias']['output file']
             r2d2_config['source_file_fmt'] = target_file
             ufsda.r2d2.store(r2d2_config)
+
+            # store satbias_cov
+            r2d2_config['file_type'] = 'satbias_cov'
```
CoryMartin-NOAA, line 52: Not a comment on this PR but a mental note for me/us, this is 'non-standard' R2D2. I had to manually modify the schema to support this. Need to discuss with @aerorahul if we are going to maintain our own NCEP schema (if we use R2D2).

ush/ufsda/stage.py
```diff
@@ -106,26 +103,74 @@ def atm_obs(config):
...
             r2d2_config['start'] = config['prev_valid_time']
             r2d2_config['end'] = r2d2_config['start']
+
+            # fetch satbias
             r2d2_config['file_type'] = 'satbias'
             target_file = ob['obs bias']['input file']
             r2d2_config['target_file_fmt'] = target_file
-            ufsda.r2d2.fetch(r2d2_config)
+            print(f"config is {config}")
+            try:
```
CoryMartin-NOAA: Instead of try, except, we can do:

```
r2d2_config['experiment'] = config.get('experiment', 'oper_gdas')
```

The .get method will return the key value if it exists, otherwise the default 'oper_gdas' is selected.

ush/ufsda/yamltools.py
```diff
@@ -69,13 +69,18 @@ def parse_config(input_config_dict, template=None, clean=True):
...
     case_enkf = int(config.get('CASE_ENKF', os.environ.get('CASE_ENKF', 'C384'))[1:])
     levs = int(config.get('LEVS', os.environ.get('LEVS', '128')))
     if 'DOHYBVAR' in config:
         dohybvar = config['DOHYBVAR']
         del config['DOHYBVAR']
     else:
         dohybvar = isTrue(os.environ.get('DOHYBVAR', 'NO'))
+    # if dohybar is true, we currently need to ensure case_enkf = case_anl
+    if dohybvar and not case_enkf == case_anl:
+        raise KeyError(f"dohybvar is '{dohybvar}' but case_enkf= '{case_enkf}' does not equal case_anl= '{case_anl}'")
```
CoryMartin-NOAA: Nit picking, but I don't think this is a 'key error' (that is reserved for missing/wrong keys in a dictionary). I think this would be a ValueError perhaps? (case_enkf == case_anl, so the values are inconsistent?)

ush/ufsda/stage.py
CoryMartin-NOAA: Wasn't sure where to put this comment, but how does the workflow know when to send 'experiment' and when to not? Like the first analysis it should use `oper_gdas` I would think, for example.

CoryMartin-NOAA (commented):
A few minor comments, but thanks @RussTreadon-NOAA for making these additions/changes!

ush/ufsda/archive.py
RussTreadon-NOAA: It would be good to get clarity on whether or not we will use R2D2 in operations sooner rather than later.

ush/ufsda/stage.py
RussTreadon-NOAA: I manually prime parallels by copying into ROTDIR the files needed to start the experiment.  It looks like g-w `setup_expt.py` is designed to do this along with other setup stuff.   We could extend this script to support parallels using GDASApp.

RussTreadon-NOAA: Replacing `try except` with 

`r2d2_config['experiment'] = config.get('experiment', 'oper_gdas')`
is a good suggestion.  

`experiment` is always defined in a parallel.   Thus, we always search for the requested file using the variable `experiment` in the path.   This is what we want for cycled files like `satbias` and `satbias_cov`.  We don't want this for (usually) fixed files like `tlapse`.   I can hardwire the `tlapse` `r2d2.fetch` to only look in `oper_gdas`.  

If the yaml used to run the stand-alone `run*py` scripts do not specify `experiment`, the default `oper_gdas` will be used.  I think this is the desired behavior.

ush/ufsda/yamltools.py
RussTreadon-NOAA: Thanks.  I'm not familiar with python lingo.   What you say makes sense.  This is a value error.  I'll change `KeyError` to `ValueError`.

I don't like having two variables, `case_anl` and `case_enkf` with the same value when `dohybvar` is true.  One variable would suffice.   However, `case_enkf` has uses in g-w beyond variational DA.  `case_enkf` is an appropriate name for these other uses.

CoryMartin-NOAA (approved):

guillaumevernieres (approved):
Looks good :+1:

FanHan-NOAA (approved):
Looks good to me! Thanks @RussTreadon-NOAA! And I'm more than happy to review your future PRs.

PR #53, opened 2022-05-19
Add support for 4D H(x) to single cycle script

### LINKED ISSUE #52: Make app handle 4D H(x)
Need YAML changes, additional backgrounds, and the use of the 4DH(x) executable.

CoryMartin-NOAA:
Closes #52 

See `/work2/noaa/da/cmartin/GDASApp/work/GDASApp/ush/sample_4dhofx.yaml` for a sample YAML file.

This is very similar to the 3D H(x), but uses a different executable, and adds in additional options for the list of forecast times and the step between them.

### DIFF
--- a/ush/run_jedi_exe.py
+++ b/ush/run_jedi_exe.py
@@ -74,7 +74,8 @@ def run_jedi_exe(yamlconfig):
         'CASE_ENKF': executable_subconfig.get('case_enkf', executable_subconfig['case']),
         'DOHYBVAR': executable_subconfig.get('dohybvar', False),
         'LEVS': str(executable_subconfig['levs']),
-        'forecast_step': executable_subconfig.get('forecast_step', 'PT6H'),
+        'forecast_steps': executable_subconfig.get('forecast_steps', 'PT6H'),
+        'BKG_TSTEP': executable_subconfig.get('forecast_step', 'PT6H'),
     }
     template = executable_subconfig['yaml_template']
     output_file = os.path.join(workdir, f"gdas_{app_mode}.yaml")
--- a/ush/run_jedi_exe.py
+++ b/ush/run_jedi_exe.py
@@ -74,6 +74,7 @@ def run_jedi_exe(yamlconfig):
         'CASE_ENKF': executable_subconfig.get('case_enkf', executable_subconfig['case']),
         'DOHYBVAR': executable_subconfig.get('dohybvar', False),
         'LEVS': str(executable_subconfig['levs']),
+        'forecast_step': executable_subconfig.get('forecast_step', 'PT6H'),
     }
     template = executable_subconfig['yaml_template']
     output_file = os.path.join(workdir, f"gdas_{app_mode}.yaml")
--- a/ush/run_jedi_exe.py
+++ b/ush/run_jedi_exe.py
@@ -7,6 +7,7 @@ import re
 import subprocess
 import sys
 import yaml
+from ufsda.misc_utils import calc_fcst_steps
 
 
 def run_jedi_exe(yamlconfig):
@@ -74,7 +75,8 @@ def run_jedi_exe(yamlconfig):
         'CASE_ENKF': executable_subconfig.get('case_enkf', executable_subconfig['case']),
         'DOHYBVAR': executable_subconfig.get('dohybvar', False),
         'LEVS': str(executable_subconfig['levs']),
-        'forecast_steps': executable_subconfig.get('forecast_steps', 'PT6H'),
+        'forecast_steps': calc_fcst_steps(executable_subconfig.get('forecast_step', 'PT6H'),
+                                          executable_subconfig['atm_window_length']),
         'BKG_TSTEP': executable_subconfig.get('forecast_step', 'PT6H'),
     }
     template = executable_subconfig['yaml_template']
--- a/ush/ufsda/misc_utils.py
+++ b/ush/ufsda/misc_utils.py
@@ -19,9 +19,10 @@ def calc_fcst_steps(fcst_step, win_length):
     """
     # need to get +- half of the window length
     # assumes only hours for now, probably bad...
+    # also assumes the window is symmetric, probably also bad
     h = int(re.findall('PT(\\d+)H', win_length)[0])
-    start = f'PT{h//2}H'
-    end = f'PT{h*2}H'
+    start = f'PT{h-h//2}H'
+    end = f'PT{h+h//2}H'
     # solo has a nice utility for this
     fcst_steps = solo.date.step_sequence(start, end, fcst_step)
     return fcst_steps
--- a/ush/ufsda/misc_utils.py
+++ b/ush/ufsda/misc_utils.py
@@ -19,7 +19,7 @@ def calc_fcst_steps(fcst_step, win_length):
     """
     # need to get +- half of the window length
     # assumes only hours for now, probably bad...
-    h = re.findall('PT(\\d+)H', win_length)[0]
+    h = int(re.findall('PT(\\d+)H', win_length)[0])
     start = f'PT{h/2}H'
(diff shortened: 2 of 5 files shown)


guillaumevernieres (dismissed):
:+1:

ush/run_jedi_exe.py
```diff
@@ -74,6 +74,8 @@ def run_jedi_exe(yamlconfig):
         'CASE_ENKF': executable_subconfig.get('case_enkf', executable_subconfig['case']),
         'DOHYBVAR': executable_subconfig.get('dohybvar', False),
         'LEVS': str(executable_subconfig['levs']),
+        'forecast_steps': executable_subconfig.get('forecast_steps', 'PT6H'),
+        'BKG_TSTEP': executable_subconfig.get('forecast_step', 'PT6H'),
```
FanHan-NOAA, line 80: It is not intuitively clear to me the purpose of adding the configurable `forecast_step` here. I assumed that `forecast_steps` is a list of times with forecasts available and `forecast_step` just indicates the temporal frequency of the list, yes? But isn't the frequency already indicated by `forecast_steps`? Correct me if I'm wrong ;)

CoryMartin-NOAA: No you're not wrong, you're just noticing my laziness :-) Do you think it should compute BKG_TSTEP from the forecast_steps now? Or is it fine to let the user have to define this? Also, then, which should be defined and which derived? PT6H -> PT6H but PT3H -> PT3H PT6H PT9H and PT1H -> PT3H PT4H...

FanHan-NOAA: I think it could be written so one can supply either one of the parameters. If only `forecast_step` is specified, then calculate the list of `forecast_steps`, if `forecast_steps` is specified, then use it directly. I can see that in some cases when the forecast times are not equally spaced, one can only specify with `forecast_steps`, but in normal situation `forecast_step` is more straightforward. So I would suggest to let the program check `forecast_steps` first, if not, then check `forecast_step` and calculate the list of times from it assuming equal spacing.

CoryMartin-NOAA: They actually do have to be equally spaced. The 'forecast steps' are used for R2D2, and that only matters based on what is stored in the database. The 'step' is used in the 4D H(x) to know when to look for the next background, so they really do have to be consistent.

FanHan-NOAA: OK, if the two have to be consistent, then only one of them needs/should to be specified in yaml. The code can be responsible for deriving one from another. If there are two configurable options, what if the inputs are inconsistent? Then you'll need to write additional logic to handle that. As a user of GDASApp, I have a slight preference for having `forecast step` in yaml as it is easier to specify.

CoryMartin-NOAA: I agree, making `forecast step` support either `PT1H`, `PT3H`, or `PT6H` is the most logical for the user. It's what GSI takes right now I think (nhrs_assim or something like that = 1,3,6). I'll make this change today and push a commit.

ush/ufsda/misc_utils.py
```diff
@@ -19,7 +19,7 @@ def calc_fcst_steps(fcst_step, win_length):
     """
     # need to get +- half of the window length
     # assumes only hours for now, probably bad...
-    h = re.findall('PT(\\d+)H', win_length)[0]
+    h = int(re.findall('PT(\\d+)H', win_length)[0])
     start = f'PT{h/2}H'
     end = f'PT{h*2}H'
```
FanHan-NOAA: I believe `h` is already int type here, but `h/2` turns float. To keep it as int type, I use `h//2` and so the window would go from `h-h//2` to `h+h//2`. But I see it could be dangerous if h is not divisible by 2.

CoryMartin-NOAA: it is string before I explicitly make it `int` but good point about checking to make sure it is still an int after the division.

FanHan-NOAA: Ah, you are right. I just tested, it is string and needs to be converted to int first.

ush/ufsda/misc_utils.py
```diff
@@ -9,6 +11,22 @@
...
+    """
+    function to return a list of forecast steps
+    for a given fcst_step (forecast step)
+    and win_length (window length)
+    """
+    # need to get +- half of the window length
+    # assumes only hours for now, probably bad...
+    h = int(re.findall('PT(\\d+)H', win_length)[0])
+    start = f'PT{h//2}H'
+    end = f'PT{h*2}H'
```
FanHan-NOAA: I think start and end should be `h-h//2` and `h+h//2`

FanHan-NOAA (dismissed):
Looks good to me except one minor thing

ush/ufsda/misc_utils.py
CoryMartin-NOAA: ha yes, I will fix.

FanHan-NOAA (approved):

ush/ufsda/misc_utils.py
```diff
@@ -9,6 +11,23 @@
...
+    function to return a list of forecast steps
+    for a given fcst_step (forecast step)
+    and win_length (window length)
+    """
+    # need to get +- half of the window length
+    # assumes only hours for now, probably bad...
+    # also assumes the window is symmetric, probably also bad
+    h = int(re.findall('PT(\\d+)H', win_length)[0])
+    start = f'PT{h-h//2}H'
+    end = f'PT{h+h//2}H'
```
FanHan-NOAA, line 25: Sorry about the late reply on this. I tested the newest version of GDASApp this morning and I found additional issue stemming from here. If both `fcst_step` and `win_length` are `'PT6H'`, the return value of this function is `['PT3H', 'PT9H']` but what we really want is `['PT6H']`. So I suggest you add a condition before l24 (something like below):
```
h2 = int(re.findall('PT(\\d+)H', fcst_step)[0])
if h2 > h//2: return [fcst_step]
```
@CoryMartin-NOAA

RussTreadon-NOAA: @FanHan-NOAA , thank you for identifying the problem and suggesting a fix.  I tried your suggestion.  It works.   PT6H files are correctly identified and copied from @CoryMartin-NOAA R2D2.   

Now `fv3jedi_hofx_nomodel.x` and `fv3jedi_var.x` seg fault when reading the background.  The executables abort with the message `unable to find mandatory variable stc in restart file 20211221.120000.fv_core.res.tile*.nc` for tiles 1, 2, 3, ..., 6.   `stc` is in `20211221.120000.sfc_data.tile*.nc`, not the fv_cor.res file.   Something must have changed elsewhere in JEDI.

FanHan-NOAA: @RussTreadon-NOAA I'm debugging a slightly different segfault with GDASApp hofx and 3dvar applications as well. I suspect something wrong with R2D2 because I tried other background files Cory provided months ago with the newest version of fv3-bundle and had no problem running `fv3jedi_hofx_nomodel.x`. If you are interested, here is my directory on Orion so you can experiment with my yaml: `/work/noaa/da/fhan/JEDI/test_humidity_assign_error/`. I used the most up to date `fv3jedi_hofx_nomodel.x` I built this morning. 

I will continue to debug what is causing problems in GDASApp.

RussTreadon-NOAA: Thanks @FanHan-NOAA .  Seems adding the line

`
field metadata override: /work/noaa/da/fhan/JEDI/src/fv3-bundle/build/fv3-jedi/test/Data/fieldmetadata/gfs-restart.yaml
`

to my yaml gets me past the stc error.   The 3dvar job later fails when it attempts to read the satbias_cov file.  This file does not exist in the R2D2 database.   I'll update PR #58 since PR #58 deals with sample yaml files.

PR #55, opened 2022-05-26
Update variational components for atmospheric cycling

RussTreadon-NOAA:
Update variational components of prototype JEDI-GDAS cycling


ush/ufsda/archive.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def atm_diags(config):
+    # fetch atm analysis obs
+    r2d2_config = {
+        'start': config['prev_valid_time'],
+        'end': config['prev_valid_time'],
+        'step': config['atm_window_length'],
+        'dump': 'gdas',
+        'experiment': config['experiment'],  # change this here and other places to be oper_{dump}
```
CoryMartin-NOAA: Comment here I don't think is valid, we would always want the diags to be archived as `experiment`. The `dump` part above should probably be in the config though (not necessary for now)

ush/ufsda/archive.py
```diff
@@ -0,0 +1,62 @@
...
+            r2d2_config['start'] = config['valid_time']
+            r2d2_config['end'] = config['valid_time']
+
+            # store satbias
+            r2d2_config['file_type'] = 'satbias'
+            target_file = ob['obs bias']['output file']
+            r2d2_config['source_file_fmt'] = target_file
+            ufsda.r2d2.store(r2d2_config)
+
+            # store satbias_cov
```
CoryMartin-NOAA: I think I remember this is 'non-standard' R2D2, meaning I added it to our schema (just on orion). We should make a note of this somewhere perhaps?

ush/ufsda/stage.py
```diff
@@ -55,6 +55,111 @@ def gdas_fix(input_fix_dir, working_dir, config):
...
+            r2d2_config['end'] = r2d2_config['start']
+
+            # fetch satbias
+            r2d2_config['file_type'] = 'satbias'
+            target_file = ob['obs bias']['input file']
+            r2d2_config['target_file_fmt'] = target_file
+            r2d2_config['experiment'] = config.get('experiment', 'oper_gdas')
+            ufsda.r2d2.fetch(r2d2_config)
+
+            # fetch satbias_cov
```
CoryMartin-NOAA: Same comment as above, I think `satbias_cov` is non-standard R2D2, which is fine, but we need to be aware of that.

CoryMartin-NOAA (dismissed):
Mostly looks good, although I suspect we will have to reconcile some things between this and @guillaumevernieres 's work. Just a few comments about needing/removing a few comments, none of which impact the code here.

ush/ufsda/archive.py
RussTreadon-NOAA: Comment on `experiment` line removed.   Comment added to `dump` line.   Done at [564a4cd](https://github.com/NOAA-EMC/GDASApp/pull/55/commits/564a4cdd53e312925018e274b27e70d1035efece)

ush/ufsda/archive.py
RussTreadon-NOAA: Comment added for `satbias_cov`.  Done at [564a4cd](https://github.com/NOAA-EMC/GDASApp/pull/55/commits/564a4cdd53e312925018e274b27e70d1035efece)

ush/ufsda/stage.py
RussTreadon-NOAA: Comment added for `satbias_cov`.  Done at [564a4cd](https://github.com/NOAA-EMC/GDASApp/pull/55/commits/564a4cdd53e312925018e274b27e70d1035efece)

CoryMartin-NOAA (approved):
Thanks @RussTreadon-NOAA !

guillaumevernieres (approved):
Looks good. Thanks @RussTreadon-NOAA .
Next step is to add unit testing for all of this!

guillaumevernieres:
> Mostly looks good, although I suspect we will have to reconcile some things between this and @guillaumevernieres 's work. Just a few comments about needing/removing a few comments, none of which impact the code here.

@CoryMartin-NOAA and @RussTreadon-NOAA don't worry about the marine stuff for now. I'll have a look once this is merged, but it didn't break my tests ... so there's that.

PR #57, opened 2022-05-26
Add VADER and Update Modulefiles

CoryMartin-NOAA:
Closes #56 

This PR adds VADER to the CMakeLists.txt for the pseudo bundle, adds vader/ to the .gitignore, and updates eckit and fckit module versions for Hera and Orion.

PR #58, opened 2022-06-03
Add sample run_jedi_exe YAMLs to repo

CoryMartin-NOAA:
As the title suggests, this adds the sample YAML files that users can modify to run through run_jedi_exe.py.

### DIFF
--- a/ush/examples/run_jedi_exe/3dvar.yaml
+++ b/ush/examples/run_jedi_exe/3dvar.yaml
@@ -1,4 +1,4 @@
-working directory: /work2/noaa/stmp/cmartin/gdas_single_test
+working directory: /work2/noaa/stmp/cmartin/gdas_single_test_3dvar
 GDASApp home: /work2/noaa/da/cmartin/GDASApp/work/GDASApp
 GDASApp mode: variational
 executable options:
@@ -15,8 +15,8 @@ executable options:
   valid_time: 2021-12-21T06:00:00Z
   dump: gdas
   case: C96
-  case_enkf: C48
-  dohybvar: true
+  case_anl: C48
+  dohybvar: false
   levs: 128
 job options:
   machine: orion
--- a/ush/examples/run_jedi_exe/3dvar.yaml
+++ b/ush/examples/run_jedi_exe/3dvar.yaml
@@ -1,0 +1,29 @@
+working directory: /work2/noaa/stmp/cmartin/gdas_single_test
+GDASApp home: /work2/noaa/da/cmartin/GDASApp/work/GDASApp
+GDASApp mode: variational
+executable options:
+  berror_yaml: /work2/noaa/da/cmartin/GDASApp/work/GDASApp/parm/atm/berror/staticb_bump.yaml
+  obs_yaml_dir: /work2/noaa/da/cmartin/GDASApp/work/GDASApp/parm/atm/obs/config
+  yaml_template: /work2/noaa/da/cmartin/GDASApp/work/GDASApp/parm/atm/variational/3dvar_dripcg.yaml
+  executable: /work2/noaa/da/cmartin/GDASApp/work/GDASApp/build/bin/fv3jedi_var.x
+  obs_list: /work2/noaa/da/cmartin/GDASApp/work/GDASApp/parm/atm/obs/lists/gdas_prototype.yaml
+  gdas_fix_root: /work2/noaa/da/cmartin/GDASApp/fix
+  atm: true
+  layout_x: 1
+  layout_y: 1
+  atm_window_length: PT6H
+  valid_time: 2021-12-21T06:00:00Z
+  dump: gdas
+  case: C96
+  case_enkf: C48
+  dohybvar: true
+  levs: 128
+job options:
+  machine: orion
+  account: da-cpu
+  queue: debug
+  partition: debug
+  walltime: '10:00'
+  ntasks: 6
+  cpus-per-task: 1
+  modulepath: /work2/noaa/da/cmartin/GDASApp/work/GDASApp/modulefiles
--- a/ush/examples/run_jedi_exe/3dhofx.yaml
+++ b/ush/examples/run_jedi_exe/3dhofx.yaml
@@ -1,0 +1,26 @@
+working directory: /work2/noaa/stmp/cmartin/gdas_single_test_hofx
+GDASApp home: /work2/noaa/da/cmartin/GDASApp/work/GDASApp
+GDASApp mode: hofx
+executable options:
+  obs_yaml_dir: /work2/noaa/da/cmartin/GDASApp/work/GDASApp/parm/atm/obs/config
+  yaml_template: /work2/noaa/da/cmartin/GDASApp/work/GDASApp/parm/atm/hofx/hofx_nomodel.yaml
+  executable: /work2/noaa/da/cmartin/GDASApp/work/GDASApp/build/bin/fv3jedi_hofx_nomodel.x
+  obs_list: /work2/noaa/da/cmartin/GDASApp/work/GDASApp/parm/atm/obs/lists/gdas_prototype.yaml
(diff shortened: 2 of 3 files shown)


ADCollard (dismissed):

RussTreadon-NOAA (commented):
Have the three new sample yamls been tested to ensure they work?

CoryMartin-NOAA:
@RussTreadon-NOAA I have used them previously but not recently confirmed they work. They will probably require recompiling on my part to use the most recent modulefiles, for example. I can confirm by testing all three later today after finishing up my upcoming unrelated PR.

CoryMartin-NOAA:
Testing suggested by @RussTreadon-NOAA unearthed problems. 1) the 4DHofX YAML will need #53 merged in first and 2) the other two tests fail (likely due to some YAML changes in recent JEDI builds). Because of this, I am marking it a draft PR until I am able to fix it (probably not until after ISDA).

CoryMartin-NOAA:
actually, this might just be user error, so I will first have the automated testing try to build and run tests and see if that looks good. I think I built/mixed old and new in my testing env which is annoying at 5pm Friday...

RussTreadon-NOAA:
Attempts to run C768 `3dhofx.yaml` or `3dvar.yaml` failed with exceptions during execution of fv3jedi_hofx_nomodel.x or fv3jedi_var.x, respectively.   gdas_prototype.yaml is missing `observers:`.    Add `observers:` to gdas_prototype.yaml.   Also add `['observers']` to stage.py.   These changes alone still don't get fv3jedi_var.x to run to completion.   Issues with missing `coupler.res` files.  Need to examine `stage.py` more closely.

ush/examples/run_jedi_exe/3dvar.yaml
```diff
@@ -0,0 +1,29 @@
+working directory: /work2/noaa/stmp/cmartin/gdas_single_test_3dvar
+GDASApp home: /work2/noaa/da/cmartin/GDASApp/work/GDASApp
+GDASApp mode: variational
+executable options:
+  berror_yaml: /work2/noaa/da/cmartin/GDASApp/work/GDASApp/parm/atm/berror/staticb_bump.yaml
+  obs_yaml_dir: /work2/noaa/da/cmartin/GDASApp/work/GDASApp/parm/atm/obs/config
+  yaml_template: /work2/noaa/da/cmartin/GDASApp/work/GDASApp/parm/atm/variational/3dvar_dripcg.yaml
+  executable: /work2/noaa/da/cmartin/GDASApp/work/GDASApp/build/bin/fv3jedi_var.x
+  obs_list: /work2/noaa/da/cmartin/GDASApp/work/GDASApp/parm/atm/obs/lists/gdas_prototype.yaml
```
RussTreadon-NOAA, line 9: Need to add `observers:` keyword to `gdas_prototype.yaml`

CoryMartin-NOAA:
>  Issues with missing `coupler.res` files. 

@RussTreadon-NOAA is it an issue with the file in the R2D2 database or is the stage script looking for/copying the wrong file?

RussTreadon-NOAA:
> > Issues with missing `coupler.res` files.
> 
> @RussTreadon-NOAA is it an issue with the file in the R2D2 database or is the stage script looking for/copying the wrong file?

@CoryMartin-NOAA , the issue may have been more to do with what @FanHan-NOAA noted in PR #53 ([comment here](https://github.com/NOAA-EMC/GDASApp/pull/53#discussion_r891273569)).   `3dvar.yaml works up to when `fv3jedi_var.x` reads in the background.  Each task fails with 

`FATAL from PE     4: fms_io(restore_state_all): unable to find mandatory variable stc in restart file 20211221.060000.fv_core.res.tile*.nc`

Adding

`field metadata override: /work/noaa/da/fhan/JEDI/src/fv3-bundle/build/fv3-jedi/test/Data/fieldmetadata/gfs-restart.yaml`


to the `geometry` section gets me past this failure.   `fv3jedi_var.x` then fails when it attempts to read the non-existent `satbias_cov` file.   We don't have and amsua_n19 `satbias_cov` file in the R2D2 database.

CoryMartin-NOAA:
@RussTreadon-NOAA got it, thanks. I remember this being an issue from before. The GSI converted bias files are combined, the UFO generated ones are separated. And I think we can't just read the same file in twice because it tries to open an already opened file... Do you have any suggestions on this? Should we just duplicate files in R2D2 to keep the two different file i/o paradigm? Other ideas?

RussTreadon-NOAA:
@CoryMartin-NOAA , we have two files for the GSI, `abias` and `abias_pc`.   I'm pretty sure these map to the JEDI `satbias` and `satbias_cov`.   JEDI adds another file, `tlapse`.   The GSI carries `tlapse` information in `abias`.   

Seems the easiest thing to do is to maintain the current paradigm of maintaining two files, `satbias` and `satbias_cov`.   This means duplicating files in R2D2 which consumes more disk space.  

The other option is to modify JEDI code to look for satbias and satbias_cov information in a single input file.  If we go down this route, I would add tlapse to the single file.  Then a single file covers radiance bias correction for both UFO and variational BC.

FanHan-NOAA:
@CoryMartin-NOAA @RussTreadon-NOAA I finally got to the bottom of why my GDASApp hofx and 3dvar application failed this morning (ran with `run_jedi_exe.py` and sample scripts from here). The problem is with #57, specifically, the module load versions of `eckit` and `fckit`. The applications ran successfully after I changed the versions back to `eckit/ecmwf-1.16.0` and `fckit/ecmwf-0.9.2` (on Orion).

Here lists two other issues with yaml changes recently merged in JEDI repos that @RussTreadon-NOAA and I discussed here and in the other thread: 

- the refactoring of `geometry:` block with the addition of `field metadata override:` and removal of `fieldsets:` (see https://github.com/JCSDA-internal/fv3-jedi/pull/480). 
- the refactoring of `observations:` block with the addition of `observers:` under `observations:` (see https://github.com/JCSDA-internal/oops/pull/1698).

CoryMartin-NOAA:
@RussTreadon-NOAA the other complication (if I am remembering correctly) is that R2D2 does not handle `satbias_cov` by default, I had to "hack" my install on Orion to support that filetype. That's not a big problem as it is a relatively easy hack, and our users will use R2D2 as a library, but something to keep in mind. 

I think the JEDI code *can* read both from a single file, the issue is that the writer writes them out separately. So a cycled experiment needs 2 input files, but one that uses GSI bias correction coefficients takes one (due to the way the converter writes out right now). However, I can't totally wrap my head around it right now to be sure what I said is correct.

@FanHan-NOAA is your issue that you are using a build compiled with the old libraries? Yes, those other two YAML changes are recent to oops and fv3-jedi and need to be made (probably in a new PR) to the GDASApp YAML templates.

FanHan-NOAA:
@CoryMartin-NOAA You are probably correct about this. Thanks for the heads up. I'll try re-building with the new libraries and see if it works

CoryMartin-NOAA:
This PR *was* to add those sample YAMLs, but Fan had already done so, now this PR solely moves the `fix` directory back to under my path, rather than hers. @CatherineThomas-NOAA has been able to run 3dhofx with this change before Orion maintenance.

RussTreadon-NOAA (commented):
Most recent change replaces `fhan` with `cmartin` in the `gdas_fix_root` path.  Assuming `fhan` files have been copied to `cmartin`, the sample yamls should still work.   It would be good to test `em to ensure this is indeed the case.   Can we add ctests at some point which use these yamls with run_jedi_exe.

CoryMartin-NOAA:
@RussTreadon-NOAA we should *definitely* soon add a 'regression' ctest capabililty to test hofx and 3dvar (at low res)

RussTreadon-NOAA (approved):
Given Cathy's success on Orion, approve.

CatherineThomas-NOAA (approved):
The 3dhofx test ran successfully on Orion with this change.  Approve.

PR #59, opened 2022-06-03
Converter script to change/add to EWOK YAMLs for use in GDASApp

CoryMartin-NOAA:
This script will:

- Change paths where appropriate
- Remove CO2 from the obs operator where appropriate
- Add a standard covariance section

so that EWOK YAMLs for each ob type (e.g. in ufo/ewok/jedi-gdas) can be translated for use here in GDASApp with minimal user intervention.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -42,6 +42,7 @@ list(APPEND test_input
   ${PROJECT_SOURCE_DIR}/test/testinput/check_yaml_keys_ref.yaml
   ${PROJECT_SOURCE_DIR}/test/testinput/check_yaml_keys_test.yaml
   ${PROJECT_SOURCE_DIR}/test/testinput/bufr_adpsfc.yaml
+  ${PROJECT_SOURCE_DIR}/test/testinput/amsua_n19_ewok.yaml
 )
 
 # create testinput dir
@@ -59,6 +60,24 @@ install(FILES ${test_input}
 
 # create testout dir
 file(MAKE_DIRECTORY ${PROJECT_BINARY_DIR}/test/testoutput)
+
+# create testreference dir
+file(MAKE_DIRECTORY ${PROJECT_BINARY_DIR}/test/testreference)
+
+# list of test reference files to install
+list(APPEND test_reference
+  ${PROJECT_SOURCE_DIR}/test/testreference/amsua_n19_gdas.yaml
+)
+
+# symlink
+foreach(FILENAME ${test_reference})
+  file(RELATIVE_PATH FILENAME_REL ${PROJECT_SOURCE_DIR}/test/testreference/ ${FILENAME})
+  file(CREATE_LINK ${FILENAME} ${PROJECT_BINARY_DIR}/test/testreference/${FILENAME_REL} SYMBOLIC)
+endforeach(FILENAME)
+
+# install
+install(FILES ${test_reference}
+        DESTINATION "test/testreference/")
 
 ##### unit tests
 # test for python coding norms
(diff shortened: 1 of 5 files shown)


ADCollard (dismissed):

emilyhcliu (dismissed):

guillaumevernieres (commented):
How about adding a simple test @CoryMartin-NOAA ?

CoryMartin-NOAA:
@guillaumevernieres can do, soon, you and your good testing practices!

RussTreadon-NOAA (commented):
Changes look good.   Tested `ush/covert_yaml_ewok2gdas.py` with several `ufo/ewok/jedi-gdas` yaml files.   Script seems to work as advertised.    

`yaml.dump` appears to write the output yaml in alphabetical order.   This is OK but it complicates comparing the `jedi-gdas` yaml with the converted yaml.   I added `sort_keys=False` to the `yaml.dump` arguments.    With this change the order remains the same between the two.   The converted yaml replaces horizontal `[  ]` lists into vertical `- ` lists. 

Ideally I'd like to see an output yaml which is identical to the input yaml except for lines which were removed, added, or modified.   Is this possible?   This isn't a requirement for this PR.  Just curious.

CoryMartin-NOAA:
@RussTreadon-NOAA I don't think it is possible unless we parse the text file manually and find/replace, which I was hoping to avoid as that can get messy. However, I have found that there is `default_flow_style=None` (which I think is default) which will determine whether to make the lists on one or multiple lines depending on the size of the list. Would that be preferred?

FanHan-NOAA:
> Changes look good. Tested `ush/covert_yaml_ewok2gdas.py` with several `ufo/ewok/jedi-gdas` yaml files. Script seems to work as advertised.
> 
> `yaml.dump` appears to write the output yaml in alphabetical order. This is OK but it complicates comparing the `jedi-gdas` yaml with the converted yaml. I added `sort_keys=False` to the `yaml.dump` arguments. With this change the order remains the same between the two. The converted yaml replaces horizontal `[ ]` lists into vertical `- ` lists.
> 
> Ideally I'd like to see an output yaml which is identical to the input yaml except for lines which were removed, added, or modified. Is this possible? This isn't a requirement for this PR. Just curious.

I know that yaml parsing tools (and yaml itself in design) generally do not care the order of mappings (except list order is preserved), so @CoryMartin-NOAA do you know of any utility tool (e.g., yamldiff a.yaml b.yaml) that compares two yamls and print out inconsistencies? I recall you had such tool and it would be handy in this case.

CoryMartin-NOAA:
@FanHan-NOAA there is https://github.com/NOAA-EMC/GDASApp/blob/develop/ush/check_yaml_keys.py, combined with its driver test script, theoretically this could do the trick but I would need to test it to see how robust it is.

guillaumevernieres (approved):
Thanks for addressing this reviewer's tantrum @CoryMartin-NOAA !

RussTreadon-NOAA:
> @RussTreadon-NOAA I don't think it is possible unless we parse the text file manually and find/replace, which I was hoping to avoid as that can get messy. However, I have found that there is `default_flow_style=None` (which I think is default) which will determine whether to make the lists on one or multiple lines depending on the size of the list. Would that be preferred?

Thank you, @CoryMartin-NOAA , for trying different options.  

I ran `sondes.yaml`, `omi_arua.yaml`, and `amsua_n19.yaml through `convert_yaml_ewok2gdas.py` with various options for `yaml.dump`.  

With `default_flow_style=None` the original section of `ufo/ewok/jedi-gdas/sondes.yaml` 
```
  - filter: Bounds Check
    filter variables:
    - name: eastward_wind
    - name: northward_wind
    minvalue: -130
    maxvalue: 130
    action:
      name: reject
```
is converted to  
```
- action: {name: reject}
  filter: Bounds Check
  filter variables:
  - {name: eastward_wind}
  - {name: northward_wind}
  maxvalue: 130
  minvalue: -130
```
I can easily see the addition of `{ }` pairs tripping me up if I were to take a converted yaml and start editing it. 

`convert_yaml_ewok2gdas.py` without any changes (retain `default_flow_style=False`) yields
```
- action:
    name: reject
  filter: Bounds Check
  filter variables:
  - name: eastward_wind
  - name: northward_wind
  maxvalue: 130
  minvalue: -130
```
This is OK but it reordered the section alphabetically:  `action`, `filter`, `maxvalue`, `minvalue`.

Retaining `default_flow_style=False` and adding `sort_keys=False` yields
```
- filter: Bounds Check
  filter variables:
  - name: eastward_wind
  - name: northward_wind
  minvalue: -130
  maxvalue: 130
  action:
    name: reject
```
This retains the order of the original input `ufo/ewok/jedi-gdas/sondes.yaml`.

Yaml files can be complicated.  There are likely cases for which `sort_keys=False` doesn't yield _pretty_ output.   Whatever options we use on the `yaml.dump` line, the converted should be readable by JEDI apps.   Don't let my comments slow the addition of `convert_yaml_ewok2gdas.py` to GDASApp.

RussTreadon-NOAA (approved):
Thank you, @CoryMartin-NOAA , for adding a ctest.

PR #64, opened 2022-06-10
Make interpolation method configurable

FanHan-NOAA:
This PR adds changes to make interpolation method configurable. Interpolation method refers to the method that transforms model fields to GeoVaLs (after model side variable change). 

Currently JEDI supports two methods for interpolating model fields to GeoVaLs: `barycentric` (default) and `inverse distance`. Interpolation method is configurable via `get values` within the `observers` block and each `obs space` has its own associated `get values`. This should not be confused with the `get values` within the `observations` block (see [hofx_nomodel.yaml](https://github.com/JCSDA-internal/fv3-jedi/blob/develop/test/testinput/hofx_nomodel.yaml#L26)). I'm not sure if the JEDI code will be refactored to consolidate the two `get values` in the future, but currently interpolation to GeoVaLs are controlled specifically by `get values` within `observers`.

This change should not impact any existing functionalities of GDASApp. But if one wants to investigate interpolation-induced uncertainty and how such uncertainty propagates from GeoVaLs to H(x) to analysis, experiments with different interpolation methods can be run by specifying an additional `interp_method` parameter in the sample yamls (in `ush/examples/run_jedi_exe/`). This is related to an ongoing work on interpolation-induced uncertainty and you can read more [here](https://docs.google.com/presentation/d/11UZ63r3ZzXB_a7qHIynuibZ-R4b3MZFsyBPPfw2HyQo/edit#slide=id.g88c2cd9c02_0_0) if interested.

### DIFF
--- a/parm/atm/obs/config/amsua_n19.yaml
+++ b/parm/atm/obs/config/amsua_n19.yaml
@@ -8,6 +8,8 @@ obs space:
     max pool size: 1
   simulated variables: [brightness_temperature]
   channels: &amsua_n19_channels 1-15
+get values:
+  interpolation method: $(INTERP_METHOD)
 obs operator:
   name: CRTM
   Absorbers: [H2O,O3]
--- a/parm/atm/obs/config/sondes.yaml
+++ b/parm/atm/obs/config/sondes.yaml
@@ -11,6 +11,8 @@ obs space:
   io pool:
     max pool size: 1
   simulated variables: [eastward_wind, northward_wind, air_temperature, specific_humidity, surface_pressure]
+get values:
+  interpolation method: $(INTERP_METHOD)
 obs operator:
   name: Composite
   components:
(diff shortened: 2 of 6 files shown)


FanHan-NOAA:
To test whether the configurable `interpolation method` is correctly placed in yaml, try setting it to something not supported by JEDI (e.g., bilinear). The program should abort with an assertion failed message like below
```
Assertion failed: interp_method_ == "barycentric" || interp_method_ == "inverse distance" in UnstructuredInterpolator
```
Or, we can run two identical experiments with differences in interpolation method only and compare the logs. Below is a snapshot of log comparison from two 3dhofx experiments I ran with `barycentric` interpolation (left) and `inverse distance` interpolation` (right) and it shows they have slightly different QC stats. 

Screen Shot 2022-06-10 at 7 32 57 AM

guillaumevernieres (approved):
Looks good @FanHan-NOAA , although I have  not tested it.

CoryMartin-NOAA (approved):
Looks good, thanks for adding this!

FanHan-NOAA:
Bilinear interpolation can be tested via `interp_method: bilinear` in the sample yamls (in `ush/examples/run_jedi_exe/`) together with the feature branch `feature/bilinear_interp` of oops (see https://github.com/JCSDA-internal/oops/pull/1790).

PR #68, opened 2022-06-28
Add YAML file for ATMS NOAA-20 

CatherineThomas-NOAA:
Added a new GDASApp hofx YAML file for ATMS NOAA-20 based on the AMSU-A NOAA-19 file.  

It is located in the parm/atm/obs/config directory.  This file was tested on Orion using the head of GDASApp develop, the 3dhofx example YAML, and run_jedi_exe.py.  It ran to completion and produced sensible diag files (plotted below using eva):

atms_ch1_ch8_hofx

### DIFF
--- a/parm/atm/obs/config/atms_n20.yaml
+++ b/parm/atm/obs/config/atms_n20.yaml
@@ -1,0 +1,371 @@
+obs space:
+  name: atms_n20
+  obsdatain:
+    obsfile: $(OBS_DIR)/$(OBS_PREFIX)atms_n20.$(OBS_DATE).nc4
+  obsdataout:
+    obsfile: $(DIAG_DIR)/diag_atms_n20_$(OBS_DATE).nc4
+  io pool:
+    max pool size: 1
+  simulated variables: [brightness_temperature]
+  channels: &atms_n20_channels 1-22
+get values:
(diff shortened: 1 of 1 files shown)


joshk24 (dismissed):
Looks good (as far as I can tell!)

parm/atm/obs/config/atms_n20.yaml
```diff
@@ -0,0 +1,371 @@
...
+    obsfile: $(DIAG_DIR)/diag_atms_n20_$(OBS_DATE).nc4
+  io pool:
+    max pool size: 1
+  simulated variables: [brightness_temperature]
+  channels: &atms_n20_channels 1-22
+get values:
+  interpolation method: $(INTERP_METHOD)
+obs operator:
+  name: CRTM
+  Absorbers: [H2O,O3,CO2]
```
CoryMartin-NOAA: For H(x) this is fine, but running with Variational DA, we will need a `linear obs operator` section with `CO2` removed, like the amsua file has. Up to you @CatherineThomas-NOAA if you want to do that now or we can do it later.

CatherineThomas-NOAA: Updated. Thanks for catching that.

CoryMartin-NOAA (approved):

PR #89, opened 2022-07-11
Force fewer tasks per node for hyperspectral IR instruments

CoryMartin-NOAA:
GDASApp crashed on Orion with CrIS, this is to reduce the number of tasks per node so that we don't use too much memory per node.

### DIFF
--- a/ush/ufsda/misc_utils.py
+++ b/ush/ufsda/misc_utils.py
@@ -60,7 +60,7 @@ def create_batch_job(job_config, working_dir, executable, yaml_path):
                 taskspernode = job_config.get("ntasks-per-node", 18)
             elif job_config['machine'] == 'orion':
                 taskspernode = job_config.get("ntasks-per-node", 24)
-            else
+            else:
                 taskspernode = job_config.get("ntasks-per-node", 12)
             sbatch = f"""#SBATCH -J GDASApp
 #SBATCH -o GDASApp.o%J
--- a/ush/ufsda/misc_utils.py
+++ b/ush/ufsda/misc_utils.py
@@ -69,7 +69,7 @@ def create_batch_job(job_config, working_dir, executable, yaml_path):
 #SBATCH -p {job_config['partition']}
 #SBATCH --ntasks={job_config['ntasks']}
 #SBATCH --ntasks-per-node={taskspernode}
-#SBATCH --cpus-per-task={job_config['cpus-per-task']}
+#SBATCH --cpus-per-task=1
 #SBATCH --exclusive
 #SBATCH -t {job_config['walltime']}"""
             f.write(sbatch)
--- a/ush/ufsda/misc_utils.py
+++ b/ush/ufsda/misc_utils.py
@@ -56,20 +56,23 @@ def create_batch_job(job_config, working_dir, executable, yaml_path):
     with open(batch_script, 'w') as f:
         f.write('#!/bin/bash\n')
         if scheduler[job_config['machine']] == 'slurm':
+            if job_config['machine'] == 'hera':
+                taskspernode = job_config.get("ntasks-per-node", 18)
+            elif job_config['machine'] == 'orion':
+                taskspernode = job_config.get("ntasks-per-node", 24)
+            else
+                taskspernode = job_config.get("ntasks-per-node", 12)
             sbatch = f"""#SBATCH -J GDASApp
 #SBATCH -o GDASApp.o%J
 #SBATCH -A {job_config['account']}
 #SBATCH -q {job_config['queue']}
 #SBATCH -p {job_config['partition']}
 #SBATCH --ntasks={job_config['ntasks']}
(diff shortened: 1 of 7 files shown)


ush/ufsda/misc_utils.py
```diff
@@ -56,20 +56,23 @@ def create_batch_job(job_config, working_dir, executable, yaml_path):
     with open(batch_script, 'w') as f:
         f.write('#!/bin/bash\n')
         if scheduler[job_config['machine']] == 'slurm':
+            if job_config['machine'] == 'hera':
+                taskspernode = job_config.get("ntasks-per-node", 18)
```
RussTreadon-NOAA, line 60: Does this PR need to add `ntasks-per-node` to the yaml files in `ush/examples/run_jedi_exe/`?

ush/ufsda/misc_utils.py
```diff
@@ -56,20 +56,23 @@ def create_batch_job(job_config, working_dir, executable, yaml_path):
...
+                taskspernode = job_config.get("ntasks-per-node", 24)
+            else:
+                taskspernode = job_config.get("ntasks-per-node", 12)
             sbatch = f"""#SBATCH -J GDASApp
 #SBATCH -o GDASApp.o%J
 #SBATCH -A {job_config['account']}
 #SBATCH -q {job_config['queue']}
 #SBATCH -p {job_config['partition']}
 #SBATCH --ntasks={job_config['ntasks']}
+#SBATCH --ntasks-per-node={taskspernode}
```
RussTreadon-NOAA, line 71: We should add logic to ensure the combination of `taskspernode` and `cpu-per-task` does not oversubscribe the node.   This requires us to know the maximum number of processes per node.

ush/ufsda/misc_utils.py
CoryMartin-NOAA: No, the second argument is the 'default' option if it is missing in the YAML.

ush/ufsda/misc_utils.py
CoryMartin-NOAA: As far as I know, for now, nothing in JEDI is written for OMP. So perhaps we should force it to be `cpu-per-task=1` for now? I agree, we don't want it to be oversubscribed, but I guess my question is it worth doing vs ensuring the user does not oversubscribe?

CatherineThomas-NOAA (dismissed):
Ran 3dhofx on Orion.  The output is the same before and after the change.  Approved.

ush/ufsda/misc_utils.py
CoryMartin-NOAA: But I suppose for an example it might be a good idea to include optional entries?

RussTreadon-NOAA: I guess the default will suffice for most cases.  My concern is that some users will not know they can set `ntasks-per-node` because they only interact with the configuration yamls in `ush/examples/run_jedi_exe/`.   I agree that we don't want to clutter the configuration yamls with too much stuff.   Is there another way to inform users that they can adjust the number of tasks per node?

CoryMartin-NOAA: I have updated the wiki page: https://github.com/NOAA-EMC/GDASApp/wiki/Using-GDASApp-in-Standalone-Mode
"tasks-per-node: - (optional) number of tasks per node (defaults: hera=18, orion=24, other machines=12)"

ush/ufsda/misc_utils.py
RussTreadon-NOAA: Yes, I run JEDI apps with OMP_NUM_THREADS=1.   This is a serious weakness of JEDI in its current form.   I hope (expect) this limitation to be removed at some point.   Your reply raises a good point.  When users see `cpus-per-tasks` in the configuration yamls, they may think we can run JEDI with multiple threads.   This is not advisable at present.  Hard wiring `cpus-per-task` to 1 is the safest thing to do now.   We'll just need to remember to undo this change once threading is safely added to JEDI.

ush/ufsda/misc_utils.py
RussTreadon-NOAA: Works for me!

ush/ufsda/misc_utils.py
CoryMartin-NOAA: ok let me update this PR with removing it from the YAMLs and hard wiring it for now

RussTreadon-NOAA: Thanks.  Sorry to make extra work for you ... especially since we will need to undo this change as some point in the future.

CoryMartin-NOAA: only took a minute, and I think it's the right call, don't want to make things confusing for the users

RussTreadon-NOAA (approved):
I haven't tried running the standalone tests from `bugfix/cris` with various ppn but the changes look OK.

PR #107, opened 2022-08-08
Add script to run UFO test executable and produce plots

### LINKED ISSUE #106: Make ush script to streamline UFO testing process
To validate QC and forward operators, we have GeoVaLs and obs that are produced by GSI and converted for use in JEDI. While the obs can work in the GDASApp as is, we want to be able to utilize the GeoVaLs in the test executables in UFO for a more apples to apples validation.

This script will facilitate the following:

- Load modules for GDASApp and EVA
- Link/copy files to a working directory
- Run the exe
- Use EVA to produce figures

The user should only need to make minor modifications (working directory, cycle, input instrument YAML) and everything else should be handled by the script.

CoryMartin-NOAA:
Closes #106

This PR does a few things but mainly in support of adding a shell script that will facilitate the process of checking QC and H(X) when using GSI-produced GeoVaLs.

Added:
- a new YAML template for the UFO test executables
- a new directory for obs YAMLs for debugging/testing
- the aforementioned shell script to ush/ufoeval

Modified:
- YAML generation tool supporting functions (mainly to remove old things that are no longer needed)
- the EVA YAML generation tool to handle input YAML from both UFO tests and FV3-JEDI executables
- Added an hofx difference vs ob pressure plot to the EVA template

### DIFF
--- a/parm/atm/hofx/hofx_ufotest.yaml
+++ b/parm/atm/hofx/hofx_ufotest.yaml
@@ -1,3 +1,3 @@
 window begin: '$(ATM_WINDOW_BEGIN)'
-window length: $(ATM_WINDOW_LENGTH)
+window end: $(ATM_WINDOW_END)
 observations: $<< $(OBS_LIST)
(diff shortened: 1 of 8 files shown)


RussTreadon-NOAA (approved):
Changes look straight forward.   Automated checks passed.

PR #110, opened 2022-08-16
Additional support for surface pressure observations

CoryMartin-NOAA:
This PR does 3 main things:

- Adds official support for surface pressure observations in hofx/var applications (note: results may not yet be right using obs in R2D2 at this time due to a bug in combining t/ps obs, see previously merged https://github.com/JCSDA-internal/ioda-converters/pull/953)
- Adds a test file for geovals for other surface ob types
- Adds additional plot types to the example EVA generation template
- Update the paths to the input files for running UFO with GSI geovals

### DIFF
--- a/parm/atm/obs/config/sfc.yaml
+++ b/parm/atm/obs/config/sfc.yaml
@@ -1,14 +1,14 @@
 obs space:
-  name: sfc_ps
+  name: sfc
   obsdatain:
-    obsfile: sfc_ps_obs_$(OBS_DATE).nc4
+    obsfile: $(OBS_DIR)/$(OBS_PREFIX)sfc.$(OBS_DATE).nc4
   obsdataout:
-    obsfile: sfc_ps_diag_$(OBS_DATE).nc4
+    obsfile: $(DIAG_DIR)/diag_sfc_$(OBS_DATE).nc4
+  io pool:
+    max pool size: 1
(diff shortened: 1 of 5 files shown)


CoryMartin-NOAA:
"(note: results may not yet be right using obs in R2D2 at this time ...)" this has been resolved on Orion. The R2D2 database there now should have corrected combined obs. Will wait to copy to Hera once fully validated.

nicholasesposito (approved):
All makes sense to me. Thanks Cory

parm/atm/obs/config/sfc.yaml
```diff
@@ -1,14 +1,14 @@
 obs space:
```
nicholasesposito, line 1: Is there a difference in purposes between the sfc.yaml in the config directory and the sfc.yaml in the testing directory?

CoryMartin-NOAA: the intention was for config/ to be the 'production' ready files, as in they have been tested and work, whereas testing/ might not yet work properly but are there as examples/a work in progress. But I'm open to changing/removing that.

CoryMartin-NOAA: The input formats of the YAMLs in the UFO tests are different than the 'real' executables, so that was why I split them into separate directories.

ADCollard (approved):

Issue #116, opened 2022-08-17
Add B Error Templated YAML for Aerosol DA


CoryMartin-NOAA:
Using https://github.com/NOAA-EMC/GDASApp/blob/develop/parm/atm/berror/staticb_bump.yaml as an example, add in a templated YAML for static B from BUMP for global aerosol DA.

CoryMartin-NOAA:
@andytangborn for a first go, I propose only creating a YAML for 'identity' **B** so that we don't have to worry about staging BUMP files, just to get the plumbing going.

andytangborn:
Ok.  I've started on the BUMP yaml, but I can do ID as well, and use that
first.

CoryMartin-NOAA:
Closed through #122

Issue #115, opened 2022-08-17
Add Obs List YAML for Aerosol DA


CoryMartin-NOAA:
Following the examples in `parm/atm/obs/lists`, add in templated YAML for AOD observations to be assimilated.

andytangborn:
Thanks. These are all helpful.

CoryMartin-NOAA:
Merged in #122

Issue #114, opened 2022-08-17
Add Templated Obs YAMLs for Global Aerosol DA


CoryMartin-NOAA:
Following the examples in `parm/atm/obs/config`, add in configuration templated YAMLs for VIIRS AOD from NOAA-20.

Issue #113, opened 2022-08-17
Add 3DVar YAML for Global Aerosol DA


CoryMartin-NOAA:
Following the 3dvar_dripcg.yaml example for atmospheric DA, add a templated YAML for global aerosol 3DVar.

CoryMartin-NOAA:
Closed through #122

PR #122, opened 2022-08-23
Feature/aero parm to add directory parm/aero yaml files for AOD DA .

andytangborn:
parm directory for aerosol DA. Includes subdirectories: berror, hofx,  lists,  obs,  variational with template yaml files.


CoryMartin-NOAA:
This PR addresses issues #113 #114 #115 and #116

parm/aero/obs/config/viirs_n20_aod.yaml
```diff
@@ -0,0 +1,24 @@
...
+    obsfile: $(DIAG_DIR)/diag_viirs_n20_$(OBS_DATE).nc4
+  io pool:
+    max pool size: 1
+  simulated variables: [aerosol_optical_depth]
+  channels: 4 
+get values:
+  interpolation method: $(INTERP_METHOD)
+obs operator:
+  name: AodCRTM
+  Absorbers: [H2O,O3]
```
CoryMartin-NOAA, line 15: Absorbers, clouds and cloud_fraction are not needed/used for AOD, right?

parm/aero/obs/config/viirs_npp_aod.yaml
```diff
@@ -0,0 +1,24 @@
...
+    obsfile: $(DIAG_DIR)/diag_viirs_npp_$(OBS_DATE).nc4
+  io pool:
+    max pool size: 1
+  simulated variables: [aerosol_optical_depth]
+  channels: 4 
+get values:
+  interpolation method: $(INTERP_METHOD)
+obs operator:
+  name: AodCRTM
+  Absorbers: [H2O,O3]
```
CoryMartin-NOAA, line 15: same comment as for N20

parm/aero/variational/3dvar_gfs_aero_bump.yaml
```diff
@@ -0,0 +1,42 @@
...
+variational:
+  minimizer:
+    algorithm: DRIPCG
+  iterations:
+  - ninner: 35 
+    gradient norm reduction: 1e-10
+    test: on
+    geometry: $(GEOM_ANL)
+    diagnostics:
+      departures: bkgmob
```
CoryMartin-NOAA: Note to self/reader: there is an OOPS PR to "flip" this so that it will actually be `ombg`

parm/aero/variational/3dvar_gfs_aero_bump.yaml
```diff
@@ -0,0 +1,42 @@
...
+    gradient norm reduction: 1e-10
+    test: on
+    geometry: $(GEOM_ANL)
+    diagnostics:
+      departures: bkgmob1
+final:
+  diagnostics:
+    departures: anlmob
+  increment:
+    output:
```
CoryMartin-NOAA: FV3 cannot currently read aerosol increments. We should for our prototype testing instead write out the analysis in FMS restart format.

CoryMartin-NOAA (changes requested):
Mostly looks good but a few things still need changed I think.

CoryMartin-NOAA:
@andytangborn can you edit this PR title and description to give it a little more 'body'? The title should be descriptive enough so we can revisit it later and the description should summarize what is added/changed.

parm/aero/obs/config/viirs_n20_aod.yaml
andytangborn: I see them in all the yaml files that we've been using, for example:  
      obs operator:
        name: AodCRTM
        Absorbers: [H2O,O3]
        obs options:
          Sensor_ID: v.viirs-m_npp
          EndianType: little_endian
          CoefficientPath: Data/crtm/
          AerosolOption: aerosols_gocart_default

parm/aero/obs/config/viirs_npp_aod.yaml
andytangborn: same comment here .

parm/aero/variational/3dvar_gfs_aero_bump.yaml
andytangborn: Should I change anything?

CoryMartin-NOAA: for now, no

parm/aero/obs/config/viirs_n20_aod.yaml
CoryMartin-NOAA: ok absorbers yes, but cloud/cloud fraction are not supposed to be included, right?

parm/aero/berror/staticb_bump_aero.yaml
```diff
@@ -0,0 +1,40 @@
...
+      filename_cplr: cor_rh.coupler.res
+- saber block name: StdDev
+  input variables: *control_vars
+  output variables: *control_vars
+  file:
+    filetype: fms restart
+    datetime: 2016-06-30T00:00:00Z
+    set datetime on read: true
+    psinfile: true
+    datapath: *staticb_dir
```
CoryMartin-NOAA: The CI tests are failing on this line it should be `*staticb_aero_dir`

parm/aero/variational/3dvar_gfs_aero_bump.yaml
CoryMartin-NOAA: I don't think this is quite right, it needs to match your 3dvar YAML you've been running with

CoryMartin-NOAA:
```
final:
  diagnostics:
    departures: anlmob
output:
  filetype: fms restart
  datapath: $(ANL_DIR)
  prefix: aeroanl
```
something like this?

andytangborn: yes. I was trying to figure out what $(GEOM_ANL) includes (like nx, ny, fieldsets etc) since these are not in the template. Shouldn't GEOM_ANL be included here as well?

CoryMartin-NOAA: you only need to specify a GEOM_ANL if the geometry for the analysis differs from the background geometry. For example, if we have C384 model but C192 DA. But we want the 'analysis' written out at the full resolution, so that is the default. If we specified the analysis geometry here, it would write out C192 RESTART files, which the model wouldn't be able to use.

parm/aero/variational/3dvar_gfs_aero_bump.yaml
```diff
@@ -0,0 +1,41 @@
+cost function:
```
CoryMartin-NOAA: This file should be renamed to something like `3dvar_dripcg_aero.yaml` or at the very least, remove `bump` from the name, since it can run with the ID **B**

andytangborn: I removed bump.

parm/aero/obs/config/viirs_n20_aod.yaml
CoryMartin-NOAA: This file doesn't have clouds or cloud fraction: https://github.com/JCSDA-internal/fv3-jedi/blob/develop/test/testinput/3dvar_gfs_aero.yaml it might not matter. Do your test YAMLs include clouds/cloud fraction?

CoryMartin-NOAA:
Just the lingering question about CRTM Clouds/cloud_fraction, otherwise I think we are about ready to merge.

parm/aero/obs/config/viirs_n20_aod.yaml
andytangborn: right. pushing now.

CoryMartin-NOAA: Still need Clouds removed too

CoryMartin-NOAA (approved):
Thanks @andytangborn looks good for a first round of testing!

CoryMartin-NOAA:
Unit test failed due to a EMC/NCO FTP issue, will merge in without that check passing.

PR #123, opened 2022-08-25
Read MOM6 diag files instead of restarts

### LINKED ISSUE #103: SOCA: read diag files instead of restarts
## Description
Read the ocn_da diag collection instead of the MOM6 restarts.

## Todo/notes
- the layer thickness field needs to be modified (swap missing and fill values with 0). Ugly, but there isn't a way to specify the missing/fill value in the fms diag manager, so something like this will have to be done:
```bash
# Change missing and _Fill to 0.0
ncatted -a _FillValue,h,o,d,0.0 ocn_da.nc
ncatted -a missing_value,h,o,d,0.0 ocn_da.nc

# Delete missing and Fill att
#ncatted -a _FillValue,h,d,d,1.0 ocn_da.nc
#ncatted -a missing_value,h,d,d,1.0 ocn_da.nc
```
- the `ssh` field in the diag file might not be the same one that we were reading from the restarts
- currents `(u,v)` appear to have been rotated already ... Need to be checked

## Dependencies
- [UFO PR #2190](https://github.com/JCSDA-internal/ufo/pull/2190)
- [SOCA PR #796](https://github.com/JCSDA-internal/soca/pull/796)

guillaumevernieres:
# Description
What the title says, mostly.

What was done:
- post-processing step for the mom6 diag files so they can be read by `soca` 
- removal of the layer thickness variable from the increment 
- removed the background r2d2 test since we won't use r2d2 for anything but obs.

fixes #103


RussTreadon-NOAA (approved):
Changes are consistent with the description.   Automated checks pass.   Looks good.

test/soca/static.sh
```diff
@@ -6,7 +6,6 @@ project_source_dir=$2
 
 soca_static=${project_binary_dir}/soca_static
 mkdir -p ${soca_static}
-mkdir -p ${soca_static}/inputnml
```
liuxiao37k, line 9: Reason to remove this line? wouldn't we need to pre-create the directory to put the input*.nml files during `cmake`? `run_jedi_exe.py soca5deg.yaml` was complaining `input.nml doesn't exist`. See error message:
```
vi /work/noaa/ng-godas/xliu/prs/pr8/test-soca-gdasapp/test-5deg/GDASApp.o6570283
```

@guillaumevernieres

CoryMartin-NOAA (approved):

liuxiao37k (commented):
It only took me a week to figure out I had to click on “finish your review” on the right top corner to actually submit it.

PR #124, opened 2022-09-02
Move aero obs list to proper place

CoryMartin-NOAA:
As the title says, I missed in #122 that the aero/ obs list is not in a consistent location with the atm/ obs list. This fixes that.


guillaumevernieres:
@CoryMartin-NOAA , Looks like an issue downloading gdasapp-fix-232c89.tgz?

andytangborn (approved):
This path change seems fine to me.

guillaumevernieres (approved):

PR #128, opened 2022-09-09
Fix variables in 3dvar_gfs_aero.yaml

andytangborn:
Correct variable names in 3dvar_gfs_aero.yaml and change output to analysis increment in model (tiles) space.

### DIFF
--- a/parm/aero/variational/3dvar_gfs_aero.yaml
+++ b/parm/aero/variational/3dvar_gfs_aero.yaml
@@ -2,7 +2,9 @@ cost function:
   cost type: 3D-Var
   window begin: '$(AERO_WINDOW_BEGIN)'
   window length: $(AERO_WINDOW_LENGTH)
-  analysis variables: &3dvars [ua,va,t,ps,sphum,ice_wat,liq_wat,o3mr]
+  analysis variables: &3dvars [sulf,bc1,bc2,oc1,oc2,
+                      dust1,dust2,dust3,dust4,dust5,
+                      seas1,seas2,seas3,seas4]
   geometry: $(GEOM_BKG)
   background:
     datapath: $(BKG_DIR)


parm/aero/variational/3dvar_gfs_aero.yaml
```diff
@@ -2,7 +2,9 @@ cost function:
   cost type: 3D-Var
   window begin: '$(AERO_WINDOW_BEGIN)'
   window length: $(AERO_WINDOW_LENGTH)
-  analysis variables: &3dvars [ua,va,t,ps,sphum,ice_wat,liq_wat,o3mr]
+  analysis variables: &3dvars [sulf,bc1,bc2,oc1,oc2,
```
CoryMartin-NOAA, line 5: We need to talk about this offline, I fear we have work ahead of us in UFO/FV3-JEDI to support the UFS coupled model with aerosols. So let's hold off on merging this until I know for sure.

andytangborn: I pushed a new commit that includes the analysis increment that we tested in the workflow, so this PR both corrects the variable list and changes the analysis output to increment.

CoryMartin-NOAA: Does `sulf` work? Or does it fail on that?

CoryMartin-NOAA: Also it looks like this PR does *not* add in increment write...

CoryMartin-NOAA: @andytangborn I just realized that you did not push a new commit to this PR... it is from 10 days ago.

andytangborn:
Sulf will fail. When I ran it I used dms.

andytangborn:
Changing to a different branch.

andytangborn:
OK. Let me see what happened.

PR #132, opened 2022-09-15
Bug fix/gsiqc eva yaml

ADCollard:
A small correction to apply the correct QC to GSI OmB for radiances in EVA


ADCollard:
This was only supposed to be one change.   I will create a clean pull request.

PR #134, opened 2022-09-20
Fixed variable list and added analysis increment.

andytangborn:
Changed 3var_gfs_aerol.yaml to fix the variable list, and write out the analysis increment.

### DIFF
--- a/parm/aero/variational/3dvar_gfs_aero.yaml
+++ b/parm/aero/variational/3dvar_gfs_aero.yaml
@@ -2,7 +2,7 @@ cost function:
   cost type: 3D-Var
   window begin: '$(AERO_WINDOW_BEGIN)'
   window length: $(AERO_WINDOW_LENGTH)
-  analysis variables: &3dvars [sulf,bc1,bc2,oc1,oc2,
+  analysis variables: &3dvars [so4,bc1,bc2,oc1,oc2,
                       dust1,dust2,dust3,dust4,dust5,
                       seas1,seas2,seas3,seas4]
   geometry: $(GEOM_BKG)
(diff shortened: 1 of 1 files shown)


parm/aero/variational/3dvar_gfs_aero.yaml
```diff
@@ -2,7 +2,9 @@ cost function:
   cost type: 3D-Var
   window begin: '$(AERO_WINDOW_BEGIN)'
   window length: $(AERO_WINDOW_LENGTH)
-  analysis variables: &3dvars [ua,va,t,ps,sphum,ice_wat,liq_wat,o3mr]
+  analysis variables: &3dvars [sulf,bc1,bc2,oc1,oc2,
```
CoryMartin-NOAA: Still the question from before, does `sulf` work with the UFS model?

CoryMartin-NOAA: Sounds like it should be `so4` according to Mariusz.

andytangborn: Oh, I thought you were referring to the workflow. I'm running the ctests in this updated version now. If not, I'll be implementing Mariusz's changes today. So it might be better to hold off on the PR. I'll let you know.

andytangborn: Also, I noticed that Mariusz while changed the bumpparameters yamls, but not the staticb yaml's (which I will update), but they all passed the tests. But I'll sort this out in a while.

andytangborn:
Yes, plus 3 more species.

I hadn't realized that fv3-bundle was now using UFS-aerosols. Last week I
was able to run all the tests (in gdasapp) still using sulf.

CoryMartin-NOAA:
@andytangborn this is for the workflow, which will be using ufs-aerosols... Mariusz is working now forwards putting those backgrounds in fv3-jedi. Does this PR work in the workflow?

andytangborn:
This PR is exactly what I ran in the workflow last week. (with DMS
substituted for sulf).    But it looks like what needs to go into fv3-jedi
needs to fit with what Mariusz is doing.

CoryMartin-NOAA (approved):
Going to approve this @andytangborn, any changes we need to make to get cycling going can go in a subsequent PR.

Issue #136, opened 2022-09-22
Add land DA component to GDASApp (1)


jiaruidong2017:
- [x] Add a templated LETKF-OI YAML for global land DA
       Following the 3dvar_gfs_aero.yaml in parm/aero/variational/
- [x] Add Obs List YAML for snow observations to be assimilated
      Following the examples in parm/aero/obs/lists/
- [x] Modify the yamltool to generate the yaml file for the snow DA test. 
      Code: ush/ufsda/yamltools.py

jiaruidong2017:
@CoryMartin-NOAA I tried to build the GDASApp, but it failed with the following error message:

```
-- Feature TESTS enabled
CMake Error at build/jedicmake/jedicmake-config.cmake:19 (message):
  File or directory
  /scratch1/NCEPDEV/global/Jiarui.Dong/JEDI/GDASApp/jedicmake/CMakeModules/Modules
  referenced by variable jedicmake_MODULES does not exist !
Call Stack (most recent call first):
  build/jedicmake/jedicmake-config.cmake:36 (set_and_check)
  crtm/CMakeLists.txt:20 (find_package)

-- Configuring incomplete, errors occurred!
See also "/scratch1/NCEPDEV/global/Jiarui.Dong/JEDI/GDASApp/build/CMakeFiles/CMakeOutput.log".
```

The whole build message is as below:
[error_in_build_gdasapp.docx](https://github.com/NOAA-EMC/GDASApp/files/9678023/error_in_build_gdasapp.docx)

Thanks.

RussTreadon-NOAA:
@jiaruidong2017 , this is a known problem.   It's been reported by Cory [(jedicmake #22](https://github.com/JCSDA-internal/jedi-cmake/issues/22)) and me ([GDASApp #140](https://github.com/NOAA-EMC/GDASApp/issues/140)).   Workarounds are being explored while we await final resolution.

jiaruidong2017:
@RussTreadon-NOAA Thank you very much for your information.

jiaruidong2017:
@CoryMartin-NOAA After I built the GDASApp package on hera, I tried to load the necessary modules as below: 

```
module use /scratch1/NCEPDEV/global/Jiarui.Dong/JEDI/GDASWork/GDASApp/modulefiles
module load GDAS/hera
```

However, I received the following error messages as below:
`
conda: Command not found.
`

I checked the modules loaded  as below:

>module list
```
Currently Loaded Modules:
[...]

PR #145, opened 2022-10-04
Add YAML files for land DA

jiaruidong2017:
https://github.com/NOAA-EMC/GDASApp/issues/136
Add YAML files (LETKF-OI and OBS LIST) to the GDASApp for land DA.


parm/land/hofx/hofx_nomodel.yaml
```diff
@@ -0,0 +1,11 @@
+window begin: '$(LAND_WINDOW_BEGIN)'
+window length: $(LAND_WINDOW_LENGTH)
+geometry: $(GEOM_BKG)
+state:
+  datapath: $(BKG_DIR)
+  filetype: fms restart
+  datetime: $(BKG_ISOTIME)
```
CoryMartin-NOAA: We probably also need this and the following to be `$(LAND_BKG_ISOTIME)` and `$(LAND_BKG_YYYYmmddHHMMSS)`? Just for the added flexibility?

parm/land/letkfoi/letkfoi.yaml
```diff
@@ -0,0 +1,40 @@
+geometry: $(GEOM_BKG)
+
+window begin: '$(LAND_WINDOW_BEGIN)'
+window length: $(LAND_WINDOW_LENGTH)
+
+background:
+   datetime: $(BKG_ISOTIME)
```
CoryMartin-NOAA: Same comment on the time variables

parm/land/letkfoi/letkfoi.yaml
```diff
@@ -0,0 +1,40 @@
...
+   members:
+     - datetime: $(BKG_ISOTIME)
+       filetype: fms restart
+       state variables: [snwdph,vtype,slmsk]
+       datapath: mem_pos/
```
CoryMartin-NOAA: Didn't @ClaraDraper-NOAA say to uses the templating here?

CoryMartin-NOAA (commented):
Few small comments

parm/land/letkfoi/letkfoi.yaml
```diff
@@ -0,0 +1,40 @@
+geometry: $(GEOM_BKG)
```
ClaraDraper-NOAA, line 1: I think we need a specific "field metadata override" within the geometry block for the land DA.

ClaraDraper-NOAA (commented):
Thanks Jiarui - before finalizing this, can you please create a yaml file for the test case in the land-offline_workflow, and check that is passes? You can override my yaml constructor by specifying the yaml name.

ClaraDraper-NOAA:
@jiaruidong2017 Note also that I'm currently doing hofx using the letkf executable, but here it looks like you're using the nomodel executable. I switched to using the letkf executable so that I didn't have to maintain two yaml structures. I suppose it's OK to switch back if we're going to use the GDASApp to generate the yaml.

parm/land/hofx/hofx_nomodel.yaml
jiaruidong2017: Updated

parm/land/letkfoi/letkfoi.yaml
jiaruidong2017: Updated

parm/land/letkfoi/letkfoi.yaml
jiaruidong2017: Made changes to use `template` format.

parm/land/letkfoi/letkfoi.yaml
jiaruidong2017: @ClaraDraper-NOAA This `field metadata override` is included in the geom dictionary `GEOM_BKG`. We can add the entry `gfs-land.yaml` for our land DA in the yamltool later.

jiaruidong2017:
> @jiaruidong2017 Note also that I'm currently doing hofx using the letkf executable, but here it looks like you're using the nomodel executable. I switched to using the letkf executable so that I didn't have to maintain two yaml structures. I suppose it's OK to switch back if we're going to use the GDASApp to generate the yaml.

@ClaraDraper-NOAA This hofx using nomodel executable is a place holder, because all other components have this doing hofx for nomodel. We can add the hofx using letkf executable later. 
@CoryMartin-NOAA Please correct me if it is not the case.

jiaruidong2017:
> Thanks Jiarui - before finalizing this, can you please create a yaml file for the test case in the land-offline_workflow, and check that is passes? You can override my yaml constructor by specifying the yaml name.

@ClaraDraper-NOAA Thanks for your comments. I can do the test in the land-offline_workflow when I updated the yamltool to generate a final yaml file from GDASApp.

ClaraDraper-NOAA:
@jiaruidong2017 It looks like everything here has been addressed (right?). Once you can confirm that the yaml generated by the GDASApp can be used in in my workflow, I'll approve the merge. 

@CoryMartin-NOAA At the moment I am maintaining builds of fv3-bundle and ioda-converters. However, my build may get out of synch with the one GDASApp is using, so I will switch to using the same versions as you. Are GDASApp users all pointing to the same build, or maintaining their own versions?

CoryMartin-NOAA:
@ClaraDraper-NOAA PR #146 will go in today/Tuesday. It switches to using tags/commits for JEDI repos instead of `develop`. I think that will hopefully make it easy to ensure the builds are the same.

ClaraDraper-NOAA:
> @ClaraDraper-NOAA PR #146 will go in today/Tuesday. It switches to using tags/commits for JEDI repos instead of `develop`. I think that will hopefully make it easy to ensure the builds are the same.

Fantastic, this will be much easier. I'll switch to using the same version. Will you have an authoritative installation on hera that I can point to? Or should I install my own version for the land DA workflow to point to?

CoryMartin-NOAA:
@ClaraDraper-NOAA no authoritative install planned at this point (but perhaps eventually?), so for now I would install your own version. That PR includes switching to `spack-stack` as well

ClaraDraper-NOAA:
Got it, thanks.

jiaruidong2017:
@ClaraDraper-NOAA Yes, I believe I have addressed all the questions. I am currently working on the yamltool to generate the final yaml file for the land DA. However, it will take some time.
In addition, all the yaml files used in the GDASApp came from the `land-offline_workflow`, which already passed both ims snow DA and gts snow DA tests in the `land-offline_workflow`.

ClaraDraper-NOAA:
@jiaruidong2017 To approve this, I'm waiting for you to use the GDASApp to create a land DA yaml, and then confirm that that yaml is correct (e.g., can be used in the offline system).

jiaruidong2017:
@CoryMartin-NOAA The current `run_jedi_exe` in the GDASApp just supported  three modes (`hofx`, `variational`, `gw_scripts`). For the land DA, we currently use the `letkf` mode. To get the final yaml file for the land DA, the `letfk` mode needs to be added to the supported modes in the GDASApp. Is it correct? I would appreciate your any comments and suggestions. Thank you very much.

CoryMartin-NOAA:
@jiaruidong2017 the `run_jedi_exe` script is to run a stand-alone analysis for testing and is not explicitly required at this time to include this for your work. The YAML should be able to be produced by the `ush/genYAML` script and should match what goes into FV3-JEDI (at least that is what I think @ClaraDraper-NOAA is asking you to do)

jiaruidong2017:
> @jiaruidong2017 To approve this, I'm waiting for you to use the GDASApp to create a land DA yaml, and then confirm that that yaml is correct (e.g., can be used in the offline system).

@CoryMartin-NOAA @ClaraDraper-NOAA I have generated the yaml file, and conducted the snow DA tests successfully. I confirm the derived yaml file is correct, and I pushed the changes into the GDASApp repo for your review. Please let me know if you have any questions. Thanks.

parm/land/obs/config/adpsfc_snow.yaml
```diff
@@ -0,0 +1,97 @@
+obs space:
+  name: adpsfc_snow
+  distribution:
+    name: Halo
+    halo size: 250e3
+  obsdatain:
+    engine:
+      type: H5File
+      obsfile: $(OBS_DIR)/adpsfc_snow_$(OBS_DATE).nc4
```
CoryMartin-NOAA: I think we probably want to follow a convention more akin to what is in ops now, suggest:
`$(OBS_DIR)/$(OBS_PREFIX)adpsfc_snow.$(OBS_DATE).nc4`

So that it can be `gdas` and `gfs` if we want.

parm/land/obs/lists/gdas_land_prototype.yaml
```diff
@@ -0,0 +1,2 @@
+observers:
+- $<< $(OBS_YAML_DIR)/adpsfc_snow.yaml
```
CoryMartin-NOAA, line 2: Just confirming, this is the only ob that will be used (at least for initial prototyping)?

ush/ufsda/yamltools.py
```diff
@@ -47,7 +47,12 @@ def parse_config(input_config_dict, template=None, clean=True):
     # compute common resolution variables
     if config_out.get('atm', True):
         config_out = atmanl_case(config_out)
+    elif config_out.get('land', True):
+        config_out = landanl_case(config_out)
+    else:
+        raise KeyError("Neither $(atm) nor ${land} defined")
```
CoryMartin-NOAA: one is () and the other is {}, I don't care which for this error print but make it consistent

ush/ufsda/yamltools.py
```diff
@@ -93,6 +98,33 @@ def atmanl_case(config):
     return config
 
 
+def landanl_case(config):
```
CoryMartin-NOAA: is this any different than the atm case? Is there a way we can reuse one?

ush/ufsda/yamltools.py
```diff
@@ -112,6 +144,36 @@ def fv3_geom_dict(case, levs, ntiles, layout, io_layout):
...
+        'ntiles': str(ntiles),
+        'field metadata override': '$(fv3jedi_fieldmetadata_dir)/gfs-land.yaml',
+
+        'time invariant fields': {
+            'state fields': {
+                'datetime': '$(LAND_WINDOW_BEGIN)',
+                'filetype': 'fms restart',
+                'skip coupler file': 'true',
+                'state variables': '[orog_filt]',
+                'datapath': '/scratch2/BMC/gsienkf/Clara.Draper/data_RnR/orog_files_Mike/',
```
CoryMartin-NOAA: We will need to stage these in a fix dir before merging it in, we will want to support both hera and orion

ush/ufsda/yamltools.py
```diff
@@ -112,6 +144,36 @@ def fv3_geom_dict(case, levs, ntiles, layout, io_layout):
...
+                'filename_orog': 'C96_oro_data.nc',
```
CoryMartin-NOAA: This needs to be resolution dependent, right?

CoryMartin-NOAA (commented):
A few more comments, but it's mostly coming together I think

parm/land/obs/config/adpsfc_snow.yaml
jiaruidong2017: Added $(OBS_PREFIX). Thanks.

parm/land/obs/lists/gdas_land_prototype.yaml
jiaruidong2017: Added the second one (ims snow).

ush/ufsda/yamltools.py
jiaruidong2017: Made change for keeping the consistency.

ush/ufsda/yamltools.py
jiaruidong2017: Yes, I made the changes.

ush/ufsda/yamltools.py
jiaruidong2017: The only difference is at the end section to call `fv3land_geom_dict` as below: 

```
config['GEOM_BKG'] = fv3land_geom_dict(case, levs, ntiles, layout, io_layout)
config['GEOM_ANL'] = fv3land_geom_dict(case_anl, levs, ntiles, layout, io_layout)
```
There are many differences in the `fv3land_geom_dict` compared with `fv3_geom_dict`. The reason I added the `landanl_case` is to show the difference in name for land analysis from the `atmanl_case` for atm analysis. If you want me to reuse the `atmanl_case` for the land, I will make the corresponding changes. Thanks.

ush/ufsda/yamltools.py
jiaruidong2017: Is it okay for the fix dir like `/scratch2/NCEPDEV/land/data/fix/` on hera?

parm/land/letkfoi/letkfoi.yaml
jiaruidong2017: `field metadata override: gfs-land.yaml` is added to the yamltools.py for the land DA.

ush/ufsda/yamltools.py
CoryMartin-NOAA: Yeah I think we rename it to 'fv3anl_case' or something like that and just add the differing `if` statement, that way we aren't duplicating 80% of the routine again. I think this change will then cover atm, land, and aero with only minor additions needed for their specific tasks.

ush/ufsda/yamltools.py
CoryMartin-NOAA: No, eventually we will want all the fix files with the other fix files, and these are maintained by the global-workflow team. This is fine for testing, but making a note here that this will have to be changed before we add land DA to the workflow itself.

ush/ufsda/yamltools.py
jiaruidong2017: As suggested to reduce the duplications, I renamed the `atmanl_case` to `fv3anl_case`, removed the `landanl_case`, and added the `landanl_case` into the `fv3anl_case`.

ush/ufsda/yamltools.py
jiaruidong2017: Add the comment line making a note to indicate future changes required before we add land DA to the workflow.

parm/land/letkfoi/letkfoi.yaml
```diff
@@ -0,0 +1,38 @@
...
+   members from template:
+     template:
+       datetime: $(LAND_BKG_ISOTIME)
+       filetype: fms restart
+       state variables: [snwdph,vtype,slmsk]
+       datapath: bkg/mem%mem%/RESTART
+       filename_sfcd: $(LAND_BKG_YYYYmmddHHMMSS).sfc_data.nc
+       filename_cplr: $(LAND_BKG_YYYYmmddHHMMSS).coupler.res
+     pattern: '%mem%'
+     nmembers: $(NMEM_LETKFOI)
```
ClaraDraper-NOAA: The number of members should be hard-coded to 2 for the LETKFOI.

parm/land/obs/config/adpsfc_snow.yaml
```diff
@@ -0,0 +1,97 @@
...
+  obsdataout:
+    engine:
+      type: H5File
+      obsfile: $(DIAG_DIR)/diag_adpsfc_snow_$(OBS_DATE).nc4
+  simulated variables: [totalSnowDepth]
+obs operator:
+  name: Composite
+  components:
+  - name: Identity
+  - name: BackgroundErrorIdentity
```
ClaraDraper-NOAA, line 19: Out of curiosity: I haven't see this before. Why are there the two components? What does specifying the BackgroundErrorIdentity here do?

ush/ufsda/yamltools.py
ClaraDraper-NOAA: This path was just intended as a temporary one for my tests while EMC finalized the fractional grids. I recommend changing this to the fix directory with the orography files that are being used by the rest of the work flow, since the land DA should be using the same files as the atmos DA.

ush/ufsda/yamltools.py
```diff
@@ -128,19 +168,32 @@ def calc_time_vars(config):
     bkg_time_obj = valid_time_obj
     config['BKG_YYYYmmddHHMMSS'] = bkg_time_obj.strftime('%Y%m%d.%H%M%S')
     config['BKG_ISOTIME'] = bkg_time_obj.strftime('%Y-%m-%dT%H:%M:%SZ')
```
ClaraDraper-NOAA, line 169: What is the ISOTIME?

ush/ufsda/yamltools.py
```diff
@@ -128,19 +168,32 @@ def calc_time_vars(config):
...
-    config['BEGIN_YYYYmmddHHMMSS'] = win_begin.strftime('%Y%m%d.%H%M%S')
+    if 'atm_window_length' in config.keys():
+        h = re.findall('PT(\\d+)H', config['ATM_WINDOW_LENGTH'])[0]
+        win_begin = valid_time_obj - datetime.timedelta(hours=int(h)/2)
+        win_end = valid_time_obj + datetime.timedelta(hours=int(h)/2)
+        config['ATM_WINDOW_BEGIN'] = win_begin.strftime('%Y-%m-%dT%H:%M:%SZ')
+        config['ATM_WINDOW_END'] = win_end.strftime('%Y-%m-%dT%H:%M:%SZ')
+        config['BEGIN_YYYYmmddHHMMSS'] = win_begin.strftime('%Y%m%d.%H%M%S')
+    # get land window begin
+    elif 'land_window_length' in config.keys():
```
ClaraDraper-NOAA, line 189: 3 questions: 
1. Should BEGIN_YYYYmmddHHMMSS have an ATM_ prefix, for consistency with the variables above? 
2. Is there any difference between the land and atm variables. If not, why do we need separate land variables. 
3. The land DA is applied to a restart - for DA with IAU, the restarts are only available at the start of the DA window (and not the middle). I'm not sure how the valid_time is defined - will this give us the increment at a time that we have a restart?

parm/land/obs/config/adpsfc_snow.yaml
ClaraDraper-NOAA: @CoryMartin-NOAA When would something be labeled GFS and when would it be GDAS? Are there any other options?

ush/ufsda/yamltools.py
ClaraDraper-NOAA: Also, do we need to change the orog file name to include the ocean details (mxXX) -  or again, just copy what's done on the atmos side.

ClaraDraper-NOAA (changes requested):
Thanks @jiaruidong2017 . I have a couple of small code change requests, and then some questions to help me understand how the GDAS app works.

ush/ufsda/yamltools.py
CoryMartin-NOAA: This distinguishes in the YAML `20221031.120000` (like for FV3) and the ISO8601 string `2022-10-31T12:00:00Z`

ush/ufsda/yamltools.py
CoryMartin-NOAA: 1) probably good to be safe
2) I guess the question is, would there ever be a land analysis at a different resolution from atmosphere? Probably not?
3) This is something we still need to sort out, for this initial go, it will be 3DVar everywhere. When I have a solution for IAU, it will be updated appropriately.

parm/land/obs/config/adpsfc_snow.yaml
CoryMartin-NOAA: GFS = early cycle 
GDAS = late cycle

That should be the only options for global, it helps distinguish which obs are available (latency, etc.)

ush/ufsda/yamltools.py
CoryMartin-NOAA: I think it's just the atm resolution for the orography but I don't know the details about the fractional grid files. I would just be consistent with what the coupled model is using on the atmos side if possible.

ClaraDraper-NOAA: The fractional grids have slightly different orography, and different files names. I think the main thing here is just to make sure that this is the same as for the atmos DA / and model.

ush/ufsda/yamltools.py
ClaraDraper-NOAA: I think(?) that for IAU the timing of the restarts is different, so we may eventually need some options in here to change the timing of the land DA update within the window. If there's a restart available at valid_time, then we're good for now.

parm/land/letkfoi/letkfoi.yaml
jiaruidong2017: The number of members is assigned in the `genYAML_prep.sh` together with all other assigned variables, such as time, resolution etc.

parm/land/obs/config/adpsfc_snow.yaml
jiaruidong2017: This is required by the Buddy Check filter.

ush/ufsda/yamltools.py
jiaruidong2017: Thanks Cory.

parm/land/letkfoi/letkfoi.yaml
jiaruidong2017: @ClaraDraper-NOAA As you requested during today's meeting, I made the change to hard-code to 2 for the LETKFOI.

ush/ufsda/yamltools.py
ClaraDraper-NOAA: @jiaruidong2017 Can you please change this to use the same fix files as the atmos DA?

ush/ufsda/yamltools.py
ClaraDraper-NOAA: @jiaruidong2017 - can you response to this?

ClaraDraper-NOAA:
@jiaruidong2017 You still need to change the fix directory, and address the questions about the timing of the window.

ush/ufsda/yamltools.py
jiaruidong2017: @CoryMartin-NOAA Do you know where I can find the fix files for the atmos DA on hera? I  can do a check for the orography. Thanks.

CoryMartin-NOAA: we don't use fix files (as of now) for orography for DA, but can we point to the model fix files that Kate Friedman et al. maintain? I assume that is what you want (what FV3 uses for orography). Or are the fields you need in the model RESTARTs?

ush/ufsda/yamltools.py
jiaruidong2017: @ClaraDraper-NOAA I think the IAU timing of the restarts for the GTS snow DA is consistent with the atm DA, which is in the middle of the DA window. For the GTS snow DA, we use the snow depth observations from the adpsfc dump files, which are same as the atm DA to use the conventional dump or prepbufr files. Please let me know if you have different idea.

ush/ufsda/yamltools.py
jiaruidong2017: Thanks @CoryMartin-NOAA 
@barlage @KateFriedman-NOAA Do you know where we can find the FV3 model fix files for the variable `orography` maintained by the global workflow on hera? Thanks.

KateFriedman-NOAA: @jiaruidong2017 The fix file set on Hera is here:

/scratch1/NCEPDEV/global/glopara/fix

The orography subset is:

/scratch1/NCEPDEV/global/glopara/fix/orog/20220805/

Note, the `20220805` subfolder is the current version for the orog files. When updates are made to those files a new timestamp (YYYYMMDD) subfolder will be created and users will be directed to point there for the orog files.

FYI, the global-workflow handles the fix file symlinks within GFS clones using a version file that a link script sources when setting up the clone for running. See these if curious:

https://github.com/NOAA-EMC/global-workflow/blob/develop/sorc/link_workflow.sh
https://github.com/NOAA-EMC/global-workflow/blob/develop/versions/fix.ver

jiaruidong2017: @KateFriedman-NOAA Thank you very much for your information.

jiaruidong2017:
> @jiaruidong2017 You still need to change the fix directory, and address the questions about the timing of the window.

@ClaraDraper-NOAA @CoryMartin-NOAA I have updated the fix directory to point out to the `glopara` fix directory for orography, and the snow DA test was successful with new generated yaml file. The questions regarding the timing of the window was addressed above. Please let me know if you have any questions. Thanks.

ush/ufsda/yamltools.py
ClaraDraper-NOAA: @jiaruidong2017 Have you added the ATM prefix to BEGIN_YYYYmmddHHMMSS? See point 1 above. 

For the timing, the (possible) issue is that the land DA update is applied to a restart. So we should make the analysis time be the same as the restart time - which means it may not be the same as the atmos analysis time. For example,  for IAU the restart is at the start of the window and I think (?) the (nominal) atmos analysis time is the middle of the window. In this case the land analysis time should be the start of the window (or, the end of the previous window).

jiaruidong2017: > Have you added the ATM prefix to BEGIN_YYYYmmddHHMMSS? See point 1 above.

@ClaraDraper-NOAA I couldn't make this change, because this is belong to the atm DA. If making this change, the yaml files in the `parm/atm/` will require the corresponding changes. I don't have authority to make any changes for the atm DA. @CoryMartin-NOAA Am I correct?

CoryMartin-NOAA: @jiaruidong2017 no feel free to make the change, as long as it is changed everywhere it shouldn't break anything. @ClaraDraper-NOAA is right about the logic here on why it is needed.

jiaruidong2017: @ClaraDraper-NOAA @CoryMartin-NOAA I added an `ATM_` prefix to the `BEGIN_YYYYmmddHHMMSS` in `yamltools.py` and `parm/atm/hofx/hofx4d.yaml`. Thanks.

jiaruidong2017: > For the timing, the (possible) issue is that the land DA update is applied to a restart. So we should make the analysis time be the same as the restart time - which means it may not be the same as the atmos analysis time. For example, for IAU the restart is at the start of the window and I think (?) the (nominal) atmos analysis time is the middle of the window. In this case the land analysis time should be the start of the window (or, the end of the previous window).

@ClaraDraper-NOAA This is a key topic for adding our land analysis. We can continue this discussion before our next PR on adding the `add_increment` for land analysis into the GDASApp.

ClaraDraper-NOAA (approved):

CoryMartin-NOAA (changes requested):
One thing remaining @jiaruidong2017 , since you changed it to `ATM_BEGIN_YYYYmmddHHMMSS`, this needs to be reflected everywhere in the parm/ directory (so all the atm files need updated or they will break with the change to the python script you made). Otherwise I think this is ready after that.

jiaruidong2017:
> One thing remaining @jiaruidong2017 , since you changed it to `ATM_BEGIN_YYYYmmddHHMMSS`, this needs to be reflected everywhere in the parm/ directory (so all the atm files need updated or they will break with the change to the python script you made). Otherwise I think this is ready after that.

@CoryMartin-NOAA I found the only yaml file including the `BEGIN_YYYYmmddHHMMSS` is `parm/atm/hofx/hofx4d.yaml`, and I made the change from `BEGIN_YYYYmmddHHMMSS` to `ATM_BEGIN_YYYYmmddHHMMSS` and already committed the changes to the GDADSApp repo.

CoryMartin-NOAA:
@jiaruidong2017 yes my mistake, I was conflating that with the BKG variable.

CoryMartin-NOAA (approved):
Thanks @jiaruidong2017 this is a good first step. This will be an iterative process and things will likely change but this is great so far!

PR #157, opened 2022-10-13
Create sfcship yaml files with QC filters

kevindougherty-noaa:
Added are yaml files for testing and config with QC filters applied and tested. Discussion of these filters can be found here [JEDI-T20#36](https://github.com/NOAA-EMC/JEDI-T2O/issues/36) about the filters being used, the data that is not accepted by GSI, but being accepted in JEDI, and some unanswered questions (mainly specific humidity).

The filters mainly come from the existing UFO yaml files [here](https://github.com/JCSDA-internal/ufo/blob/develop/ewok/skylab/ship.yaml), with the addition of blacklisting all observations from type 183. None of these observations were being assimilated in GSI.

### DIFF
--- a/parm/atm/obs/config/sfcship.yaml
+++ b/parm/atm/obs/config/sfcship.yaml
@@ -1,0 +1,77 @@
+obs space:
+  name: sfcship
+  obsdatain:
+    engine:
+      type: H5File
+      obsfile: $(OBS_DIR)/$(OBS_PREFIX)sfcship.$(OBS_DATE).nc4
+  obsdataout:
+    engine:
+      type: H5File
+      obsfile: $(DIAG_DIR)/diag_sfcship_$(OBS_DATE).nc4
+      overwrite: true
+  io pool:
+    max pool size: 1
+  simulated variables: [surface_pressure, air_temperature, specific_humidity]
+get values:
+  interpolation method: $(INTERP_METHOD)
+obs operator:
+  name: Composite
+  components:
+    - name: VertInterp
+      variables:
+      - name: air_temperature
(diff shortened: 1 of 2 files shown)


parm/atm/obs/config/sfcship.yaml
```diff
@@ -0,0 +1,77 @@
...
+      geovar_geomz: geopotential_height
+linear obs operator:
+  name: Identity
+
+obs filters:
+  # Observation range sanity check
+  - filter: Bounds Check
+    filter variables:
+    - name: surface_pressure
+    minvalue: 37499
```
CoryMartin-NOAA: Based on our Wednesday group discussion, I think we need to make clear these are 'floats' and not integers (throughout this YAML).

nicholasesposito (commented):
Besides Cory's comment, I think this is fine.

parm/atm/obs/config/sfcship.yaml
```diff
@@ -0,0 +1,77 @@
...
+    action:
+      name: reject
+    defer to post: true
+
+  # Reject all ObsType 183
+  - filter: BlackList
+    where:
+    - variable:
+        name: surface_pressure@ObsType
+      is_in: 183.0
```
nicholasesposito: From what I understand, "is_in" should remain as an integer.

kevindougherty-noaa: Got it, thank you!

nicholasesposito (approved):
Looks good.

CoryMartin-NOAA (approved):

PR #139, opened 2022-09-28
Feature/delp delz

RussTreadon-NOAA:
This PR is opened to request merger of changes in `feature/delp_delz` into `develop`.    

`feature/delp_delz` at [c1e2a44](https://github.com/NOAA-EMC/GDASApp/commit/c1e2a44efda6206fc9de3069c5696342a687157f) contains two changes with respect to develop at [8e8d725](https://github.com/NOAA-EMC/GDASApp/commit/8e8d7257d6905108dfae820caa3ae1bf993e9269):

1. extend `ush/jediinc2fv3.py` to compute delp and hydrostatic delz increments
2. update `test_gdasapp_jedi_increment_to_fv3` to accept three arguments:  (1) background (input), (2) JEDI generated analysis increment (input), and (3) FV3GFS compatible analysis increment

Details regarding the development and testing of the changes in `feature/delp_delz` are found in issue #135.

PR #168, opened 2022-10-28
Add bufr2ioda ctest for marine obs profiles

ShastriPaturi:
This PR adds bufr2ioda ctests for marine obs profile that includes both temp and salinity
Daily Tanks:
1. dbuoy: drifting buoy (b001/xx002: old tank)
2. mbuouy: moored buoy (b001/xx103: new tank)
Monthly Tanks:
1. tesac: Temperature and Salinity data (b031/xx002)
2. trkob: Temperature and Salinity data (b031/xx003)


ShastriPaturi:
some unit tests are failing. I am not sure why. I have only tested for the ctests I have added.

test/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 ##### get test data from EMC FTP server
 # set URL and hash
 set(URL "https://ftp.emc.ncep.noaa.gov/static_files/public/GDASApp")
-set(SHA "2f85f9469ee0788b3765a667fa6e9cd52b599dfb986b7dc773aaf4bcdaf05b4b")
+set(SHA "43c9e0990ea663e6da1c21121391c10731f53f31838dae112e0e63e5c38f6af7")
```
CoryMartin-NOAA: Is this the right hash? I think this might be why the unit test is failing.

ShastriPaturi: It is the hash you shared to get the subsampled files

CoryMartin-NOAA: Yes but that was over 6 weeks ago. Russ updated that file 28 days ago in https://github.com/NOAA-EMC/GDASApp/pull/139

guillaumevernieres (changes requested):
1. I don't think a license header is needed for your `yaml` files. @RussTreadon-NOAA ? @CoryMartin-NOAA ? And if it is, the year is wrong.
2. You're missing ctests for `mbuoyb`, `tesac` and`trkob`. Remove the unused yaml's or add ctests, whatever makes more sense.

CoryMartin-NOAA:
> 1. I don't think a license header is needed for your `yaml` files. @RussTreadon-NOAA ? @CoryMartin-NOAA ? And if it is, the year is wrong.
> 
>     2. You're missing ctests for `mbuoyb`, `tesac` and`trkob`. Remove the unused yaml's or add ctests, whatever makes more sense.

Yeah we really shouldn't be putting copyright statements in things. It's pervasive in the JEDI repos, but all NOAA stuff is taxpayer funded and needs to be publicly available anyways.

test/soca/CMakeLists.txt
```diff
@@ -29,6 +29,25 @@ endforeach(FILENAME)
 # tests below require the full build
 if(BUILD_GDASBUNDLE)
 
+  # link input file from iodaconv to test directory
+  #file(CREATE_LINK ${PROJECT_SOURCE_DIR}/iodaconv/test/testinput/dbuoy_20180415_bufr ${PROJECT_BINARY_DIR}/test/testdata/dbuoy_20180415_bufr SYMBOLIC)
```
CoryMartin-NOAA: remove commented out line

ShastriPaturi:
@CoryMartin-NOAA and @guillaumevernieres the requested changes are addressed.
@CoryMartin-NOAA can you remove dbuoyb.20180415_subsampled and bathy.201804.dcom_subsampled from the tarball? I will revisit them later as the subsampled files are failing. The original files work fine but too large (>5MB) to be included.

CoryMartin-NOAA:
@ShastriPaturi if you plan to revisit it later, I would prefer to keep the files there since that means I don't have to update with yet another hash and have another version of the tar ball archived on rzdm.

test/soca/CMakeLists.txt
CoryMartin-NOAA: This change still needs addressed @ShastriPaturi

ShastriPaturi:
> @ShastriPaturi if you plan to revisit it later, I would prefer to keep the files there since that means I don't have to update with yet another hash and have another version of the tar ball archived on rzdm.

Fair enough @CoryMartin-NOAA

test/soca/CMakeLists.txt
ShastriPaturi: Done

CoryMartin-NOAA (approved):
Thanks @ShastriPaturi I'm happy whenever @guillaumevernieres is happy we can merge

guillaumevernieres (approved):
Exiting stuff! Thanks @ShastriPaturi .

PR #176, opened 2022-11-02
Switch to reading a history file for the IC's

### LINKED ISSUE #172: Use a history file as initial condition for the pseudo model
We are currently pointing to a restart for the IC's and MOM6 history output for the subsequent backgrounds. 
A test I did a while back using a history file in lieu of a restart failed, I tried again yesterday and it seems fine! 
- [x] Swap MOM.res.nc with a history file in the ```background``` definition for the ocean. This is going to simplify the 1/4 deg case where MOM6 dumps at least 4 restarts per checkpoint
- [x] Update the increment function: Thickness index name isn't  the same

### LINKED ISSUE #173: Interpolate backgrounds to the same vertical geometry
... before I say something like ... Oops, I forgot, 3 years down the road! Taking care of this today.

guillaumevernieres:
- fixes #172 
- fixes #173 

In addition to the above, I added a check for the date of the background, to make sure that the date that we write in the yaml is the same as the one in the history file.


guillaumevernieres:
I was a little too trigger happy on that one. I need to change the dates in the test bkg files and update the tarball ... the assertion is doing it's job, and the test is failing! You'd think I'd know better than submitting a pr without running the ctests first, but nope.

@CoryMartin-NOAA , @RussTreadon-NOAA , don't bother triggering the `ci` on hera and orion, that PR should have been a draft.

RussTreadon-NOAA (approved):
Good to see that the Hera and Orion automated soca prep and run tests pass.   Approve.

CoryMartin-NOAA (approved):

PR #188, opened 2022-11-10
Fix to the cmake error msg when building 

guillaumevernieres:
...


CoryMartin-NOAA (approved):

RussTreadon-NOAA (approved):
Changes look good.  Approve.

PR #195, opened 2022-11-18
Make cloning JCSDA test data repositories optional

CoryMartin-NOAA:
In anticipation of the ability to make the building of JEDI ctests optional, this PR adds the cloning of the JCSDA test data repositories (ufo-data, fv3-jedi-data, etc.) optional.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -136,16 +136,6 @@ if(CLONE_JCSDADATA)
                       BRANCH ${GIT_BRANCH_FUNC} )
     endif()
 
-  # same procedure for saber-data
-    find_branch_name(REPO_DIR_NAME saber)
-    if( NOT DEFINED ENV{LOCAL_PATH_JEDI_TESTFILES} AND NOT DEFINED GIT_TAG_FUNC )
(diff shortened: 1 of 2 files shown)


aerorahul:
It is likely the tests will be enabled/disabled by the conventional CMake option `-DENABLE_TESTS=ON|OFF`.
Can one not cue from that? i.e. if `-DENABLE_TESTS=ON`, then download the test data?

Other than that, I have no opinion.  The changes seem to do the same thing, albeit introduces a `CLONE_JCSDADATA` option.

CoryMartin-NOAA:
@aerorahul I want to distinguish between their tests and ours though, because I suspect we may want ~50 ctests for GDAS that we write that are not the 1500 bundle tests. I feel like ENABLE_TESTS is too generic and wanted to avoid that

guillaumevernieres (approved):
Approving your PR preemptively @CoryMartin-NOAA to jinx the `ci`!

guillaumevernieres:
Great ... So I can jinx myself but not others ... A good addition to my list of useless powers.
Thanks for doing that work @CoryMartin-NOAA .

PR #197, opened 2022-11-22
Changes to build script and some preparation for CRTM 2.4

CoryMartin-NOAA:
This PR does 3 main things:

1. Make the build script have an option (`-f` off by default) to clean the build directory. This way the default behavior should now only rebuild what is necessary 
2. The build script now no longer runs `make` in the root build directory, but rather within fv3-jedi, soca, and iodaconv. This is to speed up build time and not build unnecessary tests and other things from the other repositories that we only need as libraries. To build everything, use option `-a`
3. Places pieces in modulefiles and CMakeLists.txt to prepare for using CRTM2.4 from the spack-stack rather than cloning/building CRTM

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,7 +60,7 @@ if(BUILD_GDASBUNDLE)
   ecbuild_bundle( PROJECT atlas GIT "https://github.com/ecmwf/atlas.git" TAG 0.29.0 )
 
 # External (required) observation operators
-  option("BUNDLE_SKIP_CRTM" "Don't build CRTM" "ON") # Don't build crtm unless user passes -DBUNDLE_SKIP_CRTM=OFF
+  option("BUNDLE_SKIP_CRTM" "Don't build CRTM" "OFF") # Don't build crtm unless user passes -DBUNDLE_SKIP_CRTM=OFF
   ecbuild_bundle( PROJECT crtm GIT "https://github.com/jcsda/crtm.git"   TAG v2.3-jedi.3 )
 
 # Build GSI-B
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,6 +60,7 @@ if(BUILD_GDASBUNDLE)
   ecbuild_bundle( PROJECT atlas GIT "https://github.com/ecmwf/atlas.git" TAG 0.29.0 )
 
 # External (required) observation operators
+  option("BUNDLE_SKIP_CRTM" "Don't build CRTM" "ON") # Don't build crtm unless user passes -DBUNDLE_SKIP_CRTM=OFF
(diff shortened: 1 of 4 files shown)


guillaumevernieres (approved):
:+1:

CMakeLists.txt
```diff
@@ -60,6 +60,7 @@ if(BUILD_GDASBUNDLE)
   ecbuild_bundle( PROJECT atlas GIT "https://github.com/ecmwf/atlas.git" TAG 0.29.0 )
 
 # External (required) observation operators
+  option("BUNDLE_SKIP_CRTM" "Don't build CRTM" "OFF") # Don't build crtm unless user passes -DBUNDLE_SKIP_CRTM=OFF
```
RussTreadon-NOAA, line 63: This works but it seems odd that setting an option `OFF` turns on   an action (ie, `OFF` = execute ecbuild_bundle for crtm).   Would something like `BUNDLE_BUILD_CRTM` or `BUILD_CRTM_BUNDLE` with `YES` as triggering ecbuild be clearer?

modulefiles/GDAS/hera.lua
```diff
@@ -63,6 +63,7 @@ load("ncview/2.1.8")
 load("netcdf-cxx4/4.3.1")
 load("json/3.10.5")
 load("py-pybind11/2.8.1")
+--load("crtm/v2.4_jedi")
```
RussTreadon-NOAA, line 66: Apart from the endianness, is `crtm/v2.4_jedi` consistent with WCOSS2 `crtm/2.4.0` which goes operational with GFS v16.3.3?

RussTreadon-NOAA (commented):
Changes look good.  Only one comment and one question.

CMakeLists.txt
CoryMartin-NOAA: 1000% agree... This is the ecbuild syntax though for some reason. It knows to skip X if you set BUNDLE_SKIP_X where X is the name of a PROJECT within the bundle

modulefiles/GDAS/hera.lua
CoryMartin-NOAA: great question, and I'm not sure. I would hope so but to be honest I wouldn't wager a large amount of my daily double on it.

RussTreadon-NOAA: OK.  I guess we'll cross this bridge when we get to it.

CMakeLists.txt
RussTreadon-NOAA: I was unaware of this behavior.    We need to work with what we have.

RussTreadon-NOAA (approved):
Question and comment addressed.  Thanks, @CoryMartin-NOAA 

Approve.

CMakeLists.txt
guillaumevernieres: I always wondered if I was the only one confused by this!

PR #203, opened 2022-11-22
EVA emcpy for marine DA

### LINKED ISSUE #202: Generate marine DA plots using EVA in the GDASApp
- [ ] Modify the existing python tool in the GDASApp [gen_eva_obs_yaml.py](https://github.com/NOAA-EMC/GDASApp/blob/develop/ush/eva/gen_eva_obs_yaml.py) to generate obs space plots using a valid ```var.yaml``` from marine DA experiments
- [ ] create eva templates for the marine obs
- [ ] test the generated yaml with eva

mjhossen:
# Description

This PR is for testing whether the existing python tools in the GDASApp works for plotting obs space on hera.

## Issue(s) addressed

- fixes #202 

## Acceptance Criteria (Definition of Done)
can produce obs space plots 

## Dependency
Copy  ```/GDASApp/ush/exgdas_global_marine_analysis_verify.sh``` in your workdir and run it as ```./exgdas_global_marine_analysis_verify.sh```


guillaumevernieres (commented):
Thanks for creating this draft PR @mjhossen .
`jedi_marine_mapplots.yaml` should be:
- split per obs spaces, which will allow you to pick the adequate color scale for the obs space. That should be ok for us since 1 obs space = 1 variable.
- renamed to something like `marine_gdas_sst.yaml` , `marine_gdas_adt.yaml` ... or something 

I'm not a big fan of the script `run_eva_marine.sh`, lets talk about is tomorrow, but instead I would rather have a script here `/GDASApp/scripts/exgdas_global_marine_analysis_verify.sh that we would eventually run with the global workflow and the stand alone app.

mjhossen:
@guillaumevernieres Completed revisions accordingly.

ShastriPaturi (changes requested):
Are the commented lines necessary to be kept?
If not, please remove them.

mjhossen:
> Are the commented lines necessary to be kept? If not, please remove them.

@ShastriPaturi Ok, I will do. Did you have a chance to run the script to plot some figures?

ShastriPaturi:
> > Are the commented lines necessary to be kept? If not, please remove them.
> 
> @ShastriPaturi Ok, I will do. Did you have a chance to run the script to plot some figures?

@mjhossen, will do

ush/eva/jedi_marine_mapplots.yaml
```diff
@@ -0,0 +1,247 @@
+# template YAML file to create EVA YAML files
```
guillaumevernieres: that file isn't needed anymore, remove.

ush/eva/marine_gdas_adt.yaml
```diff
@@ -0,0 +1,247 @@
+# template YAML file to create EVA YAML files
```
guillaumevernieres: It seems that the only difference is the variable name and the bounds between the adt and sst yamls. 
Write a utility that will take a generic yaml (based on what you have here) and will swap the bounds and var name.

ush/eva/marine_gdas_sst.yaml
```diff
@@ -0,0 +1,247 @@
+# template YAML file to create EVA YAML files
```
guillaumevernieres: see previous comment above, consolidate with adt yaml.

guillaumevernieres (commented):
How about writing a simple utility script like this:
**input:**
- variable name
- bounds

**output:**
- eva yaml for one obs space/1 variable (sst, adt, ...) with var and bounds properly populated

ush/eva/marine_gdas_adt.yaml
mjhossen: @guillaumevernieres modify the generic yaml file to swap the bounds and var name.

guillaumevernieres:
@mjhossen , did you check if the Atm examples are still working fine?

mjhossen:
> @mjhossen , did you check if the Atm examples are still working fine?



> @mjhossen , did you check if the Atm examples are still working fine?



> @mjhossen , did you check if the Atm examples are still working fine?



> @mjhossen , did you check if the Atm examples are still working fine?

@guillaumevernieres Working on it. It seems cost function is a problem for ATM.

mjhossen:
@guillaumevernieres The generic script ```gen_eva_obs_yaml.py``` is working for both marine DA and ATM DA.

guillaumevernieres:
> @guillaumevernieres The generic script `gen_eva_obs_yaml.py` is working for both marine DA and ATM DA.



> @guillaumevernieres The generic script `gen_eva_obs_yaml.py` is working for both marine DA and ATM DA.

Maybe @mjhossen , but you've definitely changed the functionality of that script. What did you do to test if your modifications still work for the ATM? 
Could you share figures maybe?

mjhossen:
@guillaumevernieres To see whether the generic script works for atm, I use an yaml file (amsua_n19_2021080100.yaml) from @CoryMartin-NOAA work dir as an input yaml. I also write a simple templateyaml for atm. It works and is able to produces figures (here is one of them). 
Screen Shot 2022-12-09 at 10 44 26 AM

guillaumevernieres:
@mjhossen , I somehow suspect that this is not what you would get with the develop version. You're also imposing new arguments, can't these be made optional? That's what we talked about a while back. While we are fine with one yaml/obsspace/group/variable, I'm not sure this is the case for the atmosphere. Did you try to make your changes backward compatible? If this isn't possible (I haven't looked at the code, so I don't know), we might want to talk to @CoryMartin-NOAA or the current users.

CoryMartin-NOAA:
In fact, one of the advantages of EVA is looping through multiple variables for making batch figures. With this approach, it seems like that is removed if you are explicitly prescribing only one variable/group.

mjhossen:
I agree with @CoryMartin-NOAA. The original script was able to create yamls for individual variables and put them in a directory. Then, EVA plots figures using each yaml files. 
@guillaumevernieres To use group and bounds, we need to put them in the templateyaml. Thus, I created a separate templateyaml for marine. I do not know is it right or wrong. It seems bound option is useful and it depends on both groups and variables.

guillaumevernieres:
> In fact, one of the advantages of EVA is looping through multiple variables for making batch figures. With this approach, it seems like that is removed if you are explicitly prescribing only one variable/group.

We're not really interested in that eva feature for now @CoryMartin-NOAA . What we want is dump eva yamls for a var/group and impose color scale bounds. I asked @mjhossen to try to use the existing eva yaml generator to do this, but maybe we should create a separate one?

CoryMartin-NOAA:
Separate is okay, but making this one have the option is probably the best route, that way code duplication is minimized.

guillaumevernieres:
> I agree with @CoryMartin-NOAA. The original script was able to create yamls for individual variables and put them in a directory. Then, EVA plots figures using each yaml files. @guillaumevernieres To use group and bounds, we need to put them in the templateyaml. Thus, I created a separate templateyaml for marine. I do not know is it right or wrong. It seems bound option is useful and it depends on both groups and variables.

Wasn't the issue properly setting the bounds @mjhossen ? Definitely don't listen to me if you know of a better way to solve a problem! But so far we haven't been able to produce the figures that we need ... Well, probably the last few commits in this branch do the trick?

guillaumevernieres:
> Separate is okay, but making this one have the option is probably the best route, that way code duplication is minimized.

Ok, thanks for the feedback @CoryMartin-NOAA . I'll leave it up to you to find the better technical solution @mjhossen , however, simply fixing what you have here seem reasonable to me.

mjhossen:
Thanks, @guillaumevernieres and @CoryMartin-NOAA. We can use the same generic scripts for both ATM and marine. I put a default option "all" for variable; it means that it can generate yamls for all the variables available in the input yaml if no variable is provided. The other two options, such as groups or bounds does not affect unless we provide these through this generic script and in the templateyaml.  @CoryMartin-NOAA Please provide more feedback or any relevant data and input yamls so that I can test the script.

ush/eva/gen_eva_obs_yaml.py
```diff
@@ -7,7 +7,7 @@
 import yaml
 
 
-def gen_eva_obs_yaml(inputyaml, templateyaml, outputdir):
+def gen_eva_obs_yaml(inputyaml, templateyaml, outputdir, group, variable, bound):
```
guillaumevernieres: `group, variable, bound(s)` should be optional and set with a default value that would return what's currently in develop.

mjhossen:
@guillaumevernieres I make them optional in the bash script that only be used by users. Some one can call as 
```$GDASApp/ush/eva/gen_eva_obs_yaml.py -i ${VarYaml}/var.yaml -t $GDASApp/ush/eva//marine_gdas_plots.yaml -o $workdir```
In the generic script, I put ```required = False``` for group, variable and bound with a default value as 
```
parser.add_argument('-g', '--group', type=str, help='ioda groups [ObsValue, ombg, ...] ', required=False, default='ObsValue')
parser.add_argument('-v', '--variable', type=str, help='Variable name ...', required=False, default='all')
parser.add_argument('-b', '--bound', type=str, nargs='+', help='min, max', required=False, default=[0, 20])
args = parser.parse_args()
gen_eva_obs_yaml(args.inputjediyaml, args.templateyaml, args.outputdir, args.group, args.variable, args.bound)
 ```

ush/eva/gen_eva_obs_yaml.py
```diff
@@ -94,5 +111,8 @@ def gen_eva_obs_yaml(inputyaml, templateyaml, outputdir):
     parser.add_argument('-i', '--inputjediyaml', type=str, help='Input JEDI YAML Configuration', required=True)
     parser.add_argument('-t', '--templateyaml', type=str, help='Template EVA YAML', required=True)
     parser.add_argument('-o', '--outputdir', type=str, help='Output directory for EVA YAMLs', required=True)
+    parser.add_argument('-g', '--group', type=str, help='ioda groups [ObsValue, ombg, ...] ', required=False, default='ObsValue')
+    parser.add_argument('-v', '--variable', type=str, help='Variable name ...', required=False, default='all')
+    parser.add_argument('-b', '--bound', type=str, nargs='+', help='min, max', required=False, default=[0, 20])
```
guillaumevernieres: @mjhossen ,  why are [0, 20] default bounds? That is not going to reproduce the figures that the other users were generating.

mjhossen: @guillaumevernieres I put it based on obsvalue of sea-surface temperature. Can you please suggest that could be the appropriate for most? If someone needs, they can provide the bound. FYI, group and bound do not work unless there is an option in the templateyaml, like the one I create for marine. I guess templateyaml would be different for atm. But generic script works for both atm and marine.

CoryMartin-NOAA: We really need the default to either be a) min,max of the data or b) have a series of defaults keyed off of variable name and group (i.e. OmB vs ObsValue)

mjhossen: @CoryMartin-NOAA Thanks.

mjhossen: It seems min/max would be more appropriate for automation. I might be wrong.

CoryMartin-NOAA: the problem with min/max is then that 2 figures cannot be easily compared. I think that though is fine for default for this PR's purpose, we will likely need to add better defaults to EVA to handle realistic defaults to ensure consistency.

guillaumevernieres: @mjhossen , I have not looked at the code, but I suspect EVA was making use of default values set by the plotting routine ... Why not just keep that as an option?

mjhossen: @guillaumevernieres There was an issue related to dynamic_vmin/vmax. Finally, figured out the problem, fixed it and push it to the branch.

guillaumevernieres: Do you have example figures for that 1 cycle that I pointed you to @mjhossen ?

scripts/exgdas_global_marine_analysis_verify.sh
```diff
@@ -0,0 +1,39 @@
...
+# run_eva_plot_marine.sh
+# This script is intended to plot a cycle of marine da analysis
+# This script will:
+# - Create a working directory
+# - Stage fix files, obs, geovals, etc.
+# - Produce an EVA YAML
+# - Run EVA
+#-------------------------------------------------------------
+#--------------- User modified options below -----------------
+cycle=2021032318
```
CoryMartin-NOAA: is the intent for this script to be part of a workflow? Or totally hand edited/run independently? If the former, then all of this needs to not be hard coded, if the latter, then it probably should go in `ush/` somewhere

scripts/exgdas_global_marine_analysis_verify.sh
```diff
@@ -0,0 +1,39 @@
...
+GDASApp=/scratch2/NCEPDEV/marineda/Jakir.Hossen/sandbox/GDASApp # Change this to your own branch
+workdir=$PWD/$cycle
+EXP_DIR=/scratch2/NCEPDEV/ocean/Guillaume.Vernieres/runs/test4jakir
+VarYaml=/scratch2/NCEPDEV/ocean/Guillaume.Vernieres/runs/test4jakir/analysis
+machine=hera
+
+# Load Modules for GDASApp
+module use $GDASApp/modulefiles
+module load GDAS/$machine
```
CoryMartin-NOAA: Why load these modules? You only need the EVA modules right?

scripts/exgdas_global_marine_analysis_verify.sh
```diff
@@ -0,0 +1,39 @@
...
+export PYTHONPATH=$GDASApp/ush:$PYTHONPATH
+
+# Create and set up the working directory
+mkdir -p $workdir
+# Load EVA modules
+module load EVA/$machine
+
+# Generate EVA YAML
+#$GDASApp/ush/eva/gen_eva_obs_yaml.py -i ${VarYaml}/var.yaml -t $GDASApp/ush/eva/marine_gdas_plots.yaml -o $workdir -g ombg -v absolute_dynamic_topography
```
CoryMartin-NOAA: why have this line commented out?

ush/eva/gen_eva_obs_yaml.py
mjhossen: @guillaumevernieres I do have. Here is two of them. 
Screen Shot 2022-12-12 at 4 41 58 PM
Screen Shot 2022-12-12 at 4 42 38 PM

scripts/exgdas_global_marine_analysis_verify.sh
```diff
@@ -0,0 +1,39 @@
...
+#$GDASApp/ush/eva/gen_eva_obs_yaml.py -i ${VarYaml}/var.yaml -t $GDASApp/ush/eva/marine_gdas_plots.yaml -o $workdir -g ombg -v absolute_dynamic_topography 
+
+$GDASApp/ush/eva/gen_eva_obs_yaml.py -i ${VarYaml}/var.yaml -t $GDASApp/ush/eva/marine_gdas_plots.yaml -o $workdir -v sea_surface_temperature -g bkg 
+# Run EVA
+cd $workdir
+for yaml in $(ls eva_*.yaml); do
+  eva $yaml
+done
+
+rm *.yaml
```
CoryMartin-NOAA: is there a reason why you would want to remove the YAML files? Might be useful to keep them to see what was passed to EVA to make those figures.

ush/eva/gen_eva_obs_yaml.py
```diff
@@ -17,22 +17,34 @@ def gen_eva_obs_yaml(inputyaml, templateyaml, outputdir):
...
-    if 'observers' in jedi_yaml_dict['observations']:
-        jediobs = jedi_yaml_dict['observations']['observers']
+    if 'cost function' in jedi_yaml_dict:
+        # cost function is in marine DA var.yaml
+        if 'observations' in jedi_yaml_dict['cost function']:
+            jediobs = jedi_yaml_dict['cost function']['observations']['observers']
     else:
-        # the unit tests have a different YAML setup
-        jediobs = jedi_yaml_dict['observations']
+        # this is for atm DA var.yaml
```
CoryMartin-NOAA: No I think the comments are wrong, it is VAR vs HofX not ATM vs SOCA. Have you tried a YAML from SOCA HofX app?

ush/eva/gen_eva_obs_yaml.py
guillaumevernieres: why [-30, 30] for `gmi` @mjhossen ?

ush/eva/gen_eva_obs_yaml.py
```diff
@@ -17,22 +17,34 @@ def gen_eva_obs_yaml(inputyaml, templateyaml, outputdir):
...
+        if 'observers' in jedi_yaml_dict['observations']:
+            jediobs = jedi_yaml_dict['observations']['observers']
+        else:
+            # the unit tests have a different YAML setup
+            jediobs = jedi_yaml_dict['observations']
     # construct a simplified list of obsspaces for EVA
     evaobs = []
     for obsspace in jediobs:
         tmp_os = obsspace['obs space']
+        sim_var = tmp_os['simulated variables'][0]
```
CoryMartin-NOAA: So just the first simulated variable? Why?

ush/eva/gen_eva_obs_yaml.py
```diff
@@ -17,22 +17,34 @@ def gen_eva_obs_yaml(inputyaml, templateyaml, outputdir):
...
         tmp_dict = {
             'name': tmp_os['name'],
             'diagfile': tmp_os['obsdataout']['engine']['obsfile'].replace('.nc4', '_0000.nc4'),
             'vars': tmp_os['simulated variables'],
             'channels': tmp_os.get('channels', None),
         }
-        evaobs.append(tmp_dict)
+
+        if sim_var == variable:
```
CoryMartin-NOAA: `if variable == 'all' or variable==sim_var`?

ush/eva/gen_eva_obs_yaml.py
```diff
@@ -94,5 +110,7 @@ def gen_eva_obs_yaml(inputyaml, templateyaml, outputdir):
     parser.add_argument('-i', '--inputjediyaml', type=str, help='Input JEDI YAML Configuration', required=True)
     parser.add_argument('-t', '--templateyaml', type=str, help='Template EVA YAML', required=True)
     parser.add_argument('-o', '--outputdir', type=str, help='Output directory for EVA YAMLs', required=True)
+    parser.add_argument('-g', '--group', type=str, help='ioda groups [ObsValue, ombg, ...] ', required=False, default='ObsValue')
```
CoryMartin-NOAA: I'm not convinced this will reproduce current behavior even with the default/required=False... Can you check from one of the test YAMLs?

ush/eva/marine_gdas_plots.yaml
```diff
@@ -0,0 +1,99 @@
...
+    # ---------- Map Plots ----------
+    # Map plot of ObsValue 
+    # --------
+
+    - batch figure:
+        variables: *variables
+        @CHANNELSKEY@
+      dynamic options:
+        - type: vminvmaxcmap
+          data variable: experiment::@GRPNAME@::${variable}
```
CoryMartin-NOAA, line 51: I guess I don't see why you would be doing this versus utilizing the same infrastructure that `${variable}` is using to loop from within EVA, can you give an example/use case?

ush/eva/gen_eva_obs_yaml.py
guillaumevernieres: @CoryMartin-NOAA , we don't have a soca h(x) app yet ... But yes, the comments are wrong @mjhossen , please fix.

ush/eva/gen_eva_obs_yaml.py
mjhossen: @guillaumevernieres This is because dynamic_vmin is set up as negative vmax. We can put 0 for dynami_vmin for the time being. As @CoryMartin-NOAA mentioned, we can do something better.

scripts/exgdas_global_marine_analysis_verify.sh
mjhossen: Fixed

scripts/exgdas_global_marine_analysis_verify.sh
mjhossen: fixed

ush/eva/gen_eva_obs_yaml.py
mjhossen: There is only one variable in the list.  Use [0] only to get the variable name, not the list. Here is the output:
```
['absolute_dynamic_topography']
['sea_surface_temperature']
```

mjhossen: There is only one variable in the list.  Use [0] only to get the variable name, not the list. Here is the output:
```
['absolute_dynamic_topography']
['sea_surface_temperature']
```

ush/eva/gen_eva_obs_yaml.py
mjhossen: ``` variable == all ``` generate yamls for all the variables in the input yaml; On the otherhand, ```sim_var == variable``` will generate the yaml for the variable passed (only if the variable is in the input yaml)

ush/eva/marine_gdas_plots.yaml
mjhossen: @GRPNAME@ is for the group of the data we intend to use to produce the figure. For example, if group is ObsValue, then only the data under ObsValue will be plotted; if the group is bkg, then the data under bkg will be plotted and so on. We assume all the groups contains the same variable but with different data.

scripts/exgdas_global_marine_analysis_verify.sh
mjhossen: I am not sure! Maybe for workflow.

CoryMartin-NOAA: @guillaumevernieres what were you thinking for this?

ush/eva/gen_eva_obs_yaml.py
```diff
@@ -17,22 +17,34 @@ def gen_eva_obs_yaml(inputyaml, templateyaml, outputdir):
...
+        sim_var = tmp_os['simulated variables'][0]  # to get only the variable, not the list
         tmp_dict = {
             'name': tmp_os['name'],
             'diagfile': tmp_os['obsdataout']['engine']['obsfile'].replace('.nc4', '_0000.nc4'),
             'vars': tmp_os['simulated variables'],
             'channels': tmp_os.get('channels', None),
         }
-        evaobs.append(tmp_dict)
+
+        if variable == sim_var:
```
CoryMartin-NOAA: I'm still confused about this, so if the variable you pass is the *first* variable in the simulated variable list, or it is `all`, then you include this dictionary? What if it is the second variable in the list? Seems like the if statement should set the dictionary? or am I not understanding what you are trying to do

ush/eva/marine_gdas_plots.yaml
CoryMartin-NOAA: Ok, I think a lot of this is a hack that would be better to modify EVA for a real fix, but this is probably fine for now.

scripts/exgdas_global_marine_analysis_verify.sh
guillaumevernieres: @CoryMartin-NOAA , we want a verification step in the workflow, but this could be tackled in a different PR. Could you remove that script @mjhossen ?

ush/eva/gen_eva_obs_yaml.py
CoryMartin-NOAA: The default should probably be `''`, so that it replaces it with nothing

scripts/exgdas_global_marine_analysis_verify.sh
CoryMartin-NOAA: it's fine to put this in ush/ just not in scripts/

mjhossen: moved.

ush/eva/gen_eva_obs_yaml.py
mjhossen: fixed

ush/eva/gen_eva_obs_yaml.py
guillaumevernieres: Good catch @CoryMartin-NOAA .
@mjhossen , it works in your case because the marine ufo only have 1 simulated obs (at least the one you are testing with). That won't work in general.

mjhossen: Got your point. Thanks. It seems better to keep it as original to generate all obs yamls. Modify the script such as `exgdas..._verify.sh` to plot only what we want instead of plotting all.

ush/eva/gen_eva_obs_yaml.py
mjhossen: fixed

scripts/exgdas_global_marine_analysis_verify.sh
mjhossen: fixed

guillaumevernieres:
I'll take the blame for giving poor directions on that small project. I'm closing this PR. We will regroup/rethink how we want to implement this.

PR #212, opened 2022-11-30
Have the CI cron job load modules from PR

CoryMartin-NOAA:
As the title says, this PR changes the CI scripts so that before compiling/running tests, the modules from the PR are loaded, and not static ones from the CI version of the repo (which is usually lagging behind).

This PR also does a few other minor things:

1. compiles with 8 jobs not the default of 6
2. removes the logfiles before the CI runs in case it is an updated commit to the same PR (in this case previously the log file would get concatenated and thus be confusing to read on GitHub
3. Add the list of failed tests to the log that gets pushed to GitHub


guillaumevernieres (approved):
Progress! Thanks @CoryMartin-NOAA .

PR #214, opened 2022-11-30
Bugfix for when testrun dir does not yet exist

CoryMartin-NOAA:
I think the test added by @jiaruidong2017 may fail depending on in what order the unit tests run in. This was not caught in the testing because the build directory apparently was not completely removed... Let's see if this fixes it.


jiaruidong2017 (approved):
@CoryMartin-NOAA You are absolutely right, because we didn't consider to remove the upper level directories. Thanks for the bug fixes.

PR #215, opened 2022-11-30
Groundwork for workflow based testing

CoryMartin-NOAA:
This PR does the following:

1. Creates a new GitHub Action that will clone first the NCEP global-workflow, and then clone (manually) the GDASApp in `global-workflow/sorc/gdas.cd`. This action then will run ctests that are enabled with the workflow option (see below). The action runs `sudo mkdir -p /work/noaa` to trick global-workflow into thinking it is on Orion.
2. Adds a new option to the build script, and the top level CMakeLists.txt file (`WORKFLOW_TESTS`), to control if tests that need the global workflow should be built or not
3. Adds a new ctest that uses the g-w setup_expt.py script to create an experiment directory and populate it with config.* (e.g. config.base) files

Check here to see some output to confirm that config.base is being created like it would be on Orion: https://github.com/NOAA-EMC/GDASApp/actions/runs/3586750549/jobs/6036321385 (look in the run ctest section)

### DIFF
--- a/.github/workflows/unittests_g-w.yaml
+++ b/.github/workflows/unittests_g-w.yaml
@@ -54,7 +54,7 @@ jobs:
       run: |
         export ecbuild_DIR=$GITHUB_WORKSPACE/ecbuild/lib/cmake/ecbuild
         mkdir $GITHUB_WORKSPACE/global-workflow/sorc/gdas.cd/build && cd $GITHUB_WORKSPACE/global-workflow/sorc/gdas.cd/build
-        cmake -DBUILD_GDASBUNDLE=OFF $GITHUB_WORKSPACE/global-workflow/sorc/gdas.cd/
+        cmake -DBUILD_GDASBUNDLE=OFF -DWORKFLOW_TESTS=ON $GITHUB_WORKSPACE/global-workflow/sorc/gdas.cd/
 
     - name: Build GDASApp
       run: |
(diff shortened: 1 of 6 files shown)


CoryMartin-NOAA:
Eagle eyed observers may note that there are now labels like `Hera-GW-RT`... That will work the same way as the `Hera-RT` label does now, but is an upcoming cron that will combine what is in this PR with a full build of GDASApp within the global-workflow.

.github/workflows/unittests_g-w.yaml
```diff
@@ -0,0 +1,68 @@
...
+      with:
+        path: ecbuild
+        key: ${{ runner.os }}-ecbuild
+
+    - name: Install ecBuild
+      if: steps.cache-ecbuild.outputs.cache-hit != 'true'
+      run: |
+        git clone https://github.com/ecmwf/ecbuild.git ecbuild
+        cd ecbuild
+        git checkout 3.6.1
```
aerorahul, line 52: perhaps actually install it:
https://github.com/ecmwf/ecbuild/blob/develop/INSTALL.rst

aerorahul (dismissed):
how much can I pay you to do this in global-workflow as well?

.github/workflows/unittests_g-w.yaml
CoryMartin-NOAA: Yeah install is probably the way to go, this was the lazy path of least resistance.

guillaumevernieres: > Yeah install is probably the way to go, this was the lazy path of least resistance.

and what's wrong with that?

CoryMartin-NOAA: @aerorahul it now `sudo make install`'s

aerorahul (approved):

guillaumevernieres (approved):
Most excellent @CoryMartin-NOAA !

PR #217, opened 2022-12-01
Do not cache ecbuild

CoryMartin-NOAA:
Like the title says. Caching ecbuild in GitHub Actions is doing more harm than good, and a clone/install only takes a few seconds anyways


guillaumevernieres (approved):
:+1:

PR #219, opened 2022-12-02
Add cron scripts to run tests within NCEP global-workflow

CoryMartin-NOAA:
These scripts add a capability to (as part of a cron) do the following:

1. Get a list of GDASApp PRs with the label `$(machine)-GW-RT`
2. Clone NCEP's global-workflow develop branch
3. Checkout all of the global-workflow and the PR commit hash for GDASApp (in `sorc/gdas.cd`)
4. Build the workflow components and link files
5. Run GDASApp ctests (some of which now can depend on workflow configuration/scripts)

Note: this cron will not be enabled on Hera or Orion until this PR is merged to develop


guillaumevernieres (approved):
:+1:

PR #222, opened 2022-12-02
Add seaice

### LINKED ISSUE #213: Add seaice DA
To do list:
- [x] Add the CICE backgrounds to the pseudo model, ~~ideally from the cice history instead of the CICE restarts~~
- [x] BUMP init of the sea ice control variables
- [x] add the sea ice control variables to the var yaml
- ~~Apply the Soca2cice var change to insert the analysis in the restart ~~
- [x] make sure all the above is part of the CI

guillaumevernieres:
# Description (not really ...)
What the title says, the PR adds sea ice DA to the GDAS.
- fixes #213 

# Remaining issues:
- This PR does not implement the seaice analysis insertion, no iau in CICE so required at some point.
- As @AndrewEichmann-NOAA pointed out, the order of the templated yaml is not preserved, which makes the order of the linear variable changes and the obs filters wrong.

### DIFF
--- a/parm/soca/berror/parametric_stddev_b.yaml
+++ b/parm/soca/berror/parametric_stddev_b.yaml
@@ -9,8 +9,8 @@ output geometry:
   fields metadata: fields_metadata.yaml
 
 linear variable change:
-  input variables: [tocn, socn, ssh]
-  output variables: [tocn, socn, ssh]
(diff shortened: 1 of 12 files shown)


CoryMartin-NOAA (approved):

ShastriPaturi (approved):
Thanks

PR #234, opened 2022-12-07
Testing of the Marine DA jjobs

### LINKED ISSUE #232: 5deg Marine 3DVar ctest using j-job
Same as #229 but for ocean and sea-ice.

**Remaining issues, notes to self:**
- [x] yaml file passed to exp setup py script is still hard-coded ... do a sed or something
~- [ ] Fixing this: _line 41: setpdy.sh: command not found_ would make the test shell script a bit cleaner~
- [x] Add a config.base hard-coded to the testing needs somewhere
- [x] Analysis output link not created for some reason ...
~- [ ] fms io, not sure I care at this point~
- [x] copy backgrounds from comrot to rundir 

Not really needed yet:
- [x] minor updates to the g-w to be pushed as a pr <-- Not needed unless you run in containers

guillaumevernieres:
Still a bit of a work todo but good enough for a draft PR.

#### Addition of 3 ctests related to running with the g-w
  - Copy/paste of @CoryMartin-NOAA 's ``setup_expt.py`` test adjusted to `soca`
  - jjob for the Marine DA prep step (create yaml's, stage obs and bkg's, ...)
  - jjob for the Marine DA run step (gridgen, static B init., run var, create iau increment, ...)   

#### Stuff left to do
 - Only tested in a singularity container
 ~- As I type this I'm wondering if the changes I made to the g-w are needed (probably not)~
 - The prep jjob fails when run twice in a row, simple fix but still in the todo list
 - ...

#### Issues addressed
- fixes #232


scripts/exgdas_global_marine_analysis_prep.py
```diff
@@ -41,8 +41,7 @@
 my_home = os.path.dirname(os.path.dirname(my_dir))
 gdas_home = os.path.join(os.getenv('HOMEgfs'), 'sorc', 'gdas.cd')
 sys.path.append(os.path.join(os.getenv('HOMEgfs', my_home), 'ush'))
-print(f"sys.path={sys.path}")
-
+sys.path.append(os.path.join(gdas_home, 'ush'))
```
CoryMartin-NOAA, line 47: FYI, Rahul will want this set at the jjob_prep.sh step to be consistent with `jobs/rocoto/prep.sh` in g-w

scripts/exgdas_global_marine_analysis_prep.py
```diff
@@ -149,8 +148,7 @@ def gen_bkg_list(window_begin=' ', bkg_path='.', file_type='gdas.t*.ocnf00[3-9]'
 # create output directory for soca DA
 anl_out = os.path.join(comout, 'ocnanal_'+os.getenv('CDATE'), 'Data')
 ufsda.mkdir(anl_out)
-ufsda.symlink(os.path.join(anl_dir, 'Data'), anl_out, remove=False)
-
+ufsda.symlink(anl_out, os.path.join(anl_dir, 'Data'), remove=False)
```
CoryMartin-NOAA, line 166: I'm hoping to phase out `ufsda` in favor of `pygw`. I will not expect that of soca/atm just yet though but plan to try to do so for aero

test/soca/gw/jjob_prep.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+bindir=$1
+srcdir=$2
+
+# export env. var.
+source ${srcdir}/test/soca/gw/runtime_vars.sh $bindir $srcdir
+
+# prepare background from previous cycle
```
CoryMartin-NOAA, line 8: why is this being done here and not part of the prep job?

CoryMartin-NOAA (commented):
a few comments/questions

test/soca/gw/jjob_prep.sh
guillaumevernieres: Are you talking about the background for the previous cycle or the sourcing of the env. vars. @CoryMartin-NOAA ?

scripts/exgdas_global_marine_analysis_prep.py
guillaumevernieres: OK, thanks for the info.

test/soca/gw/jjob_prep.sh
CoryMartin-NOAA: the background, shouldn't it be 'available' already

guillaumevernieres: They would be after 1 cycle. But yes, good point ... It's not part of the prep jjob currently. I don't think I understand quite yet how the first cycle is handled in the g-w. I've been copying background files of the first cycle to a specific location to so far.  
Going to wait and see what @RussTreadon-NOAA is doing :)

CoryMartin-NOAA: I think you are talking about a 'half cycle' vs the 'full cycle', we shouldn't be testing the 'half cycle' because there is no DA yet. Assume there has already been a forecast, and the files you need are in `$COMROT` when you get to your prep step.

guillaumevernieres: Ha yes ... I forgot the jargon of the first cycle issue. 
So that's what that step above is doing, populating `$COMROT`, do you mean that step should be done in the previous test? Or hidden in the `CMakeLists.txt`?

CoryMartin-NOAA: `$COMROT` should be `$GDASAPP_TESTDATA/lowres/` I think? Perhaps it needs slightly modified. Or I would suggest as you said a new test that just stages the ROTDIR.

guillaumevernieres: it's definitely **not a test** and shouldn't be, it doesn't cover any code, just copies files somewhere. But it probably would make sense to bury the creation of the `$COMROT` in the `CMakeList.txt`. Just following the path of least resistance, it's simpler to do `cp` in `bash` than in `cmake` :)

guillaumevernieres:
Looks like I hard-coded `CONTAINER` in `config.base`, so the new tests are failing and ... the file name is 136 character long, so it's truncated when going through fms io. I'm not sure why that bug didn't show up the last time we ran the ci.

guillaumevernieres:
@RussTreadon-NOAA , @CoryMartin-NOAA , I pushed a few bug fixes related to the marine DA in the global-workflow, see [PR 1172](https://github.com/NOAA-EMC/global-workflow/pull/1172). We'll have to wait until this is merged before running the CI again ...

guillaumevernieres:
@RussTreadon-NOAA , @CoryMartin-NOAA : We'll need to wait for the [g-w pr#1172](https://github.com/NOAA-EMC/global-workflow/pull/1172) to be merged before triggering the CI, but it tested OK in the container and Orion.

guillaumevernieres:
That's funny, I did all the testing on Orion and none on Hera ...
@CoryMartin-NOAA  Any chance you could remove the orion pr 234 in your ci directory? The CI is failing because it's using a g-w develop that is a few commits behind and doesn't include the few bug fixes that were merge yesterday evening.

scripts/exgdas_global_marine_analysis_prep.py
CoryMartin-NOAA: This comment still applies :-), set this at the j-job with `$PYTHONPATH`

scripts/exgdas_global_marine_analysis_prep.py
```diff
@@ -113,18 +112,27 @@ def gen_bkg_list(window_begin=' ', bkg_path='.', file_type='gdas.t*.ocnf00[3-9]'
...
+
+        # aggregate seaice variables and dump aggregated ice bkg in out_path
         ice_filename = ocn_filename.replace("ocn", "ice")
         agg_ice_filename = ocn_filename.replace("ocn", "agg_ice")
         agg_seaice(os.path.join(bkg_path, ice_filename),
-                   os.path.join(bkg_path, agg_ice_filename))  # aggregate seaice variables
+                   os.path.join(out_path, agg_ice_filename))
+
+        # copy ocean bkg to out_path
+        ufsda.disk_utils.copyfile(os.path.join(bkg_path, ocn_filename),
```
CoryMartin-NOAA, line 127: good for now, but we will want to transition to `pygw` tools ASAP

test/soca/gw/bkg2comrot.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+bindir=$1
+srcdir=$2
+
+# export env. var.
+source ${srcdir}/test/soca/gw/runtime_vars.sh $bindir $srcdir
+
+# prepare background from previous cycle
+mkdir -p $bindir/test/soca/gw/testrun/ROTDIRS/gdas_test/gdas.20180415/06/ocean/
```
CoryMartin-NOAA: Use standard variables in the path here? like ROTDIR, PDY, cyc, etc

test/soca/gw/jjob_prep.sh
```diff
@@ -0,0 +1,13 @@
...
+mkdir -p $bindir/test/soca/gw/testrun/ROTDIRS/gdas_test/gdas.20180415/06/ocean/
```
CoryMartin-NOAA: Same comment, don't use hard coded path/dates here

scripts/exgdas_global_marine_analysis_prep.py
guillaumevernieres: created this issue: [GDASApp/242](https://github.com/NOAA-EMC/GDASApp/issues/242)

test/soca/gw/jjob_prep.sh
guillaumevernieres: I see I wrote it's definitely not a ctest ... But I was convinced otherwise last week. It was added as a separate ctest. Marking as resolved @CoryMartin-NOAA .

scripts/exgdas_global_marine_analysis_prep.py
guillaumevernieres: Resolving that one since you addressed that problem in one of the last few PR's @CoryMartin-NOAA .

test/soca/gw/bkg2comrot.sh
guillaumevernieres: Cleaned up the run time variables and now making use of pdy/cyc/...

test/soca/gw/jjob_prep.sh
guillaumevernieres: done as well. I did also run shellcheck since my bash style was all over the place.

test/soca/gw/setup_workflow_exp.sh
```diff
@@ -3,65 +3,65 @@
...
 srcdir=$2
 
+# export env. var.
+source "${srcdir}/test/soca/gw/runtime_vars.sh" "${bindir}" "${srcdir}"
+
 # test experiment variables
-idate=2021032318
-edate=2021032418
+idate=${PDY}${cyc}
+edate=${PDY}${cyc}
```
CoryMartin-NOAA, line 11: this is fine, but this is not equivalent to what was here before, just a comment...

guillaumevernieres: the dates were wrong.

CoryMartin-NOAA (dismissed):
Looks good to me

test/soca/gw/setup_workflow_exp.sh
guillaumevernieres: It doesn't look like that `setup_expt.py` does anything with the dates. `PDY` and `cyc` from the environment that we set is what's used.

guillaumevernieres:
We shall see, but I think all our test should be re-runable (it's a word ... ) now.

guillaumevernieres:
Re-requesting your review @CoryMartin-NOAA , I had to add a few changes to make some of the ctest "re-runable".

test/soca/gw/setup_workflow_exp.sh
CoryMartin-NOAA: the dates are just used to set up the rocoto XML (I believe)

CoryMartin-NOAA (approved):

PR #235, opened 2022-12-08
Change to using global-workflow YAML templates

CoryMartin-NOAA:
This PR moves from using the yamltools inside of GDASApp to those inside of the global-workflow (pygw).

The pygw YAML parser follows this convention:
`!INC` to include a YAML file
`!ENV` to denote a environment variable is somewhere in that entry
`${VAR}` is used for environment variables 
`$(VAR)` is used for variables defined in a python dictionary
`{{VAR}}` is used for variables defined in a python dictionary that are cycle dependent (note this may need to be in strings to work properly i.e. `'{{VAR}}'`


CoryMartin-NOAA:
@RussTreadon-NOAA FYI this PR will likely break the tests you've been working on this week. I think the YAML generation part of your ex-script will need to be modified to use the pygw tools. Let me know if I can help with this.

guillaumevernieres (approved):
Thanks for sorting out the marine stuff as well @CoryMartin-NOAA .

guillaumevernieres:
@RussTreadon-NOAA , can we merge this or do you want to look a it first?

RussTreadon-NOAA (approved):
Let's merge and go from there.

Issue #238, opened 2022-12-12
debug in ioda converter


jiaruidong2017:
When the ioda converter is used to convert IMS snow depth data in netcdf format into the ioda format under GDASApp settings, the commands are as below: 

```
IMS_IODA=${EXECDIR}/imsfv3_scf2ioda.py
echo 'do_landDA: calling ioda converter'
python ${IMS_IODA} -i IMSscf.${YYYY}${MM}${DD}.${TSTUB}.nc -o ${JEDIWORKDIR}ioda.IMSscf.${YYYY}${MM}${DD}.${TSTUB}.nc
```
It failed with the following error messages: 

```
Traceback (most recent call last):
  File "/scratch1/NCEPDEV/global/Jiarui.Dong/JEDI/GDASWork/GDASApp/build/bin/imsfv3_scf2ioda.py", line 23, in <module>
    import ioda_conv_engines as iconv
  File "/scratch1/NCEPDEV/global/Jiarui.Dong/JEDI/GDASWork/GDASApp/build/lib/pyiodaconv/ioda_conv_engines.py", line 2, in <module>
    import ioda_obs_space as ioda_os
ModuleNotFoundError: No module named 'ioda_obs_space'
```
Any idea? Thanks.

CoryMartin-NOAA:
@jiaruidong2017 you need to add the `ioda_obs_space` to your `$PYTHONPATH` environment variable. This is probably something like `export PYTHONPATH=$PYTHONPATH:/scratch1/NCEPDEV/global/Jiarui.Dong/JEDI/GDASWork/GDASApp/build/lib/python3.7/pyioda/
`

However, looking here, I don't see it at all, perhaps it requires running `make` inside of `ioda`? I assume you didn't run the build script with the option to build everything?

jiaruidong2017:
[...]

PR #240, opened 2022-12-12
Some updates to land_create_ensemble and land_imsproc. 

jiaruidong2017:
1. Add the code to convert the IMS OBS data into IODA format. (issue #238)
2. Modify the code to consider the different snow variable names in GFSv17 in the `test_gdasapp_land_create_ens` ctest.


test/land/test_imsproc.sh
```diff
@@ -53,6 +53,12 @@ done
 ulimit -Ss unlimited
 ${EXECDIR}/calcfIMS.exe
 
+export PYTHONPATH=$PYTHONPATH:${project_source_dir}/build/lib/python3.7/pyioda/
```
CoryMartin-NOAA: the `python3.7` part here is dangerous. @guillaumevernieres did you ever figure out how to do this the correct way?

CoryMartin-NOAA (commented):
One small comment that we need to sort out but otherwise looks good.

test/land/test_imsproc.sh
guillaumevernieres: Sorry, just seeing this. I don't think this is the correct way, but probably safer:
[stage.py#L19](https://github.com/NOAA-EMC/GDASApp/blob/9143acb3a1f5043e09b3ef7f67e72d536d50eb23/ush/ufsda/stage.py#L19)

jiaruidong2017: @guillaumevernieres Thanks for your reply. Do you suggest me to use something like `'python3.'+str(sys.version_info[1])` to replace `python3.7`?

jiaruidong2017: @guillaumevernieres Or to modify `imsfv3_scf2ioda.py` following the approach in the `stage.py`.

CoryMartin-NOAA: There must be some CMAKE variable here to use? How does CMAKE know to create/put the IODA python in this spot in the first place?

CoryMartin-NOAA: `python${Python3_VERSION_MAJOR}.${Python3_VERSION_MINOR}`
See: https://github.com/JCSDA-internal/ioda/blob/d9250472620f039dd8c590acee0350417206fc94/src/engines/ioda/python/CMakeLists.txt#L9

jiaruidong2017: @CoryMartin-NOAA Because I have little experience in the cmake usage, I still don't have an idea how to proceed and need further help with this. 

From my limited investigation, it seems both `Python3_VERSION_MAJOR` and `Python3_VERSION_MINOR` should be generated during the IODA build process.

CoryMartin-NOAA: @jiaruidong2017 does it work to do:
```
export PYTHONPATH=$PYTHONPATH:${project_source_dir}/build/lib/python${Python3_VERSION_MAJOR}.${Python3_VERSION_MINOR}/pyioda/
```
?

jiaruidong2017: No, because both `Python3_VERSION_MAJOR` and `Python3_VERSION_MINOR` are empty.

jiaruidong2017: See the `PYTHONPATH` below: 
/scratch2/NCEPDEV/stmp1/Jiarui.Dong/gdasapp/gdasapp.4/build/lib/`python.`/pyioda/

CoryMartin-NOAA: ah, okay, so we need to figure out how to get them set... ok let me think about it, stay tuned.

jiaruidong2017: @CoryMartin-NOAA Thanks.

CoryMartin-NOAA: Oh, this is a shell script, did you try setting them it in the CMake first?

jiaruidong2017: No. Do you mean I need to add some settings in the `test/land/CMakeLists.txt`?

CoryMartin-NOAA: yes

jiaruidong2017: @CoryMartin-NOAA I don't know how to set both `Python3_VERSION_MAJOR` and `Python3_VERSION_MINOR` in the CMakeLists.txt correctly. Do you have any example for me to learn how to set them? Thank you very much.

CoryMartin-NOAA: https://github.com/NOAA-EMC/GDASApp/blob/98a8294aa8f89da2e40699d833389079fccd370e/test/land/CMakeLists.txt#L15
Change to something like:
```
"GDASAPP_TESTDATA=$ENV{GDASAPP_TESTDATA};Python3_MAJOR_VERSION=${Python3_MAJOR_VERSION};Python3_MINOR_VERSION=${Python3_MINOR_VERSION}"
```
I think that should do the trick

jiaruidong2017: Thanks @CoryMartin-NOAA for your great help. Is it `Python3_MAJOR_VERSION` or `Python3_VERSION_MAJOR`?

CoryMartin-NOAA: Should be what I wrote before, ${Python3_VERSION_MAJOR}, not what I wrote just now from memory :-)

jiaruidong2017: @CoryMartin-NOAA It worked perfectly. I will push the changes to the repo. Thanks again.

CoryMartin-NOAA (approved):
Looks good, thanks @jiaruidong2017 !

PR #249, opened 2022-12-16
Make genYAML not depend on ufsda module; allows for UFO acceptance testing to work again

CoryMartin-NOAA:
Somewhere along the way, ufsda now has a dependency on IODA, so even if you were YAML substituting, you would need to do all the `export PYTHONPATH=...` stuff.

This PR makes genYAML not depend on ufsda, but rather just `pygw` which should be installed in the GDASApp env on Hera/Orion (and in the CI test). This should mean that now the stand alone ush/ufoeval script should work properly once again, assuming one is using the new updated YAMLs in develop that follow the pygw convention.


CoryMartin-NOAA:
@BrettHoover-NOAA this PR should fix the issue you were having to get the ush/ufoeval script to work with updated YAMLs

BrettHoover-NOAA (dismissed):
I appreciate your putting this together, thank you!

BrettHoover-NOAA (approved):

Issue #237, opened 2022-12-09
Use `${PDY}${cyc}` instead of `${CDATE}`?


guillaumevernieres:
can we please use `${PDY}${cyc}` instead of `${CDATE}`?
We will eventually deprecate the use of `${CDATE}` in favor of the above.

_Originally posted by @aerorahul in https://github.com/NOAA-EMC/GDASApp/pull/236#discussion_r1044605036_

See this issue for more details: [global-workflow](https://github.com/NOAA-EMC/global-workflow/issues/198)

AndrewEichmann-NOAA:
Addressed by https://github.com/NOAA-EMC/GDASApp/pull/252 and https://github.com/NOAA-EMC/global-workflow/pull/1191

PR #252, opened 2022-12-16
Feature/soca remove cdate

AndrewEichmann-NOAA:
Removes references to CDATE as non-local variable; in a couple of instances it is assigned locally to `${PDY}${cyc}`. Addresses issue https://github.com/NOAA-EMC/GDASApp/issues/237

Tested with ctests+ in build/test (ie, not just for soca)

Depends on the merging of global-workflow PR https://github.com/NOAA-EMC/global-workflow/pull/1191 to work for g-w, but  should work for soca alone

Files changed:
parm/soca/obs/config/adt_j3.yaml
parm/soca/obs/config/icec_emc.yaml
parm/soca/obs/config/salt_profile_fnmoc.yaml
parm/soca/obs/config/sss_smap.yaml
parm/soca/obs/config/sst_noaa19_l3u.yaml
parm/soca/obs/config/sst_noaa19_l3u_so025.yaml
scripts/exgdas_global_marine_analysis_prep.py
test/soca/gw/config.base
test/soca/gw/runtime_vars.sh
test/soca/runtime_vars.sh
ush/ufsda/misc_utils.py


scripts/exgdas_global_marine_analysis_prep.py
```diff
@@ -161,7 +161,8 @@ def gen_bkg_list(bkg_path, out_path, window_begin=' ', file_type='gdas.t*.ocnf00
 ufsda.mkdir(bkg_dir)
 
 # create output directory for soca DA
-anl_out = os.path.join(comout, 'ocnanal_'+os.getenv('CDATE'), 'Data')
+cdate = os.getenv('PDY')+os.getenv('cyc')
+anl_out = os.path.join(comout, 'ocnanal_' + cdate, 'Data')
```
CoryMartin-NOAA: I know this is @guillaumevernieres 's problem, not yours, but `$COMOUT` should not have `ocnanal_`... in it. Just making a note here.

test/soca/gw/config.base
```diff
@@ -95,6 +95,7 @@ export BASE_ENV="$HOMEgfs/env"
 export BASE_JOB="$HOMEgfs/jobs/rocoto"
 
 # EXPERIMENT specific environment parameters
+CDATE = ${PDY}${cyc}
```
CoryMartin-NOAA, line 98: if you don't export it, it only works in this file

test/soca/runtime_vars.sh
```diff
@@ -5,18 +5,16 @@ project_binary_dir=$1
...
 export CDUMP=gdas
 export GDUMP=gdas
 export OPREFIX=${CDUMP}.t${cyc}z.
-export PDY=20180415
-export cyc=12
-export OPREFIX=${CDUMP}.t${cyc}z.
 export assim_freq=6            # DA window
 export COMOUT=${project_binary_dir}/test/soca/3dvar
-export DATA=${project_binary_dir}/test/soca/3dvar/ocnanal_${CDATE}
+export DATA=${project_binary_dir}/test/soca/3dvar/ocnanal_${PDY}${cyc}
```
CoryMartin-NOAA: See my comment in g-w about this path

test/soca/gw/config.base
AndrewEichmann-NOAA: Isn't that the idea?

CoryMartin-NOAA: fair enough

guillaumevernieres:
@AndrewEichmann-NOAA , the `CI` won't work until your gw PR is merged.

scripts/exgdas_global_marine_analysis_prep.py
guillaumevernieres: Not sure I understand your comment @CoryMartin-NOAA `$COMOUT` doesn't have `ocnanal_...` in it ... Or do you mean anl_out should be something else?

_I'm being generous and decided to share some of my problems with @AndrewEichmann-NOAA_

CoryMartin-NOAA: if you look at like the existing atmos GDAS ROTDIR, there is no subdirectory for analysis stuff, it is all there together. Not sure this is a 'requirement' or not, just thinking of consistency.

guillaumevernieres: OK, saved as an issue @CoryMartin-NOAA .

guillaumevernieres (commented):
The CI worked @AndrewEichmann-NOAA  but you need to address @CoryMartin-NOAA 's comment about the missing `$RUN` as a prefix. Looks good otherwise.

guillaumevernieres:
I'm taking the liberty to push to your branch @AndrewEichmann-NOAA ... Needs a few minor updates.

test/soca/runtime_vars.sh
guillaumevernieres: We have an issue to address the extra dir level, if that's what you were reffering to @CoryMartin-NOAA , I removed PDY from the name if that's what your comment was about.

RussTreadon-NOAA (approved):
Changes look straightforward.  Approve pending successful GW-RT on Hera and Orion.

AndrewEichmann-NOAA:
For clarification, is this the comment in https://github.com/NOAA-EMC/global-workflow/pull/1191 that needs to be addressed?

[CoryMartin-NOAA](https://github.com/CoryMartin-NOAA) (https://github.com/NOAA-EMC/global-workflow/pull/1191#discussion_r1051181907)
I think @aerorahul wants instead this to be `${RUN}ocnanal_${cyc}` per a conversation he and I had yesterday afternoon

Member
@[aerorahul](https://github.com/aerorahul) aerorahul (https://github.com/NOAA-EMC/global-workflow/pull/1191#discussion_r1051188752)
Yes. Thanks @CoryMartin-NOAA.
`${PDY}` is not available at this time, so the issue is correct on that, that `${CDATE}` should become `${cyc}`.

I also realized yesterday, that our run directories will get clobbered when gdas and gfs cycles start. To distinguish between those, it is best to prepend ocnanal with a `${RUN}`.

`${RUN} `is set upstream of the job in the ecflow/rocoto job-card. EMC internally translates that to `${CDUMP}`, which is also not available at this point in the j-job.

guillaumevernieres:
I forgot if I added `RUN` yesterday @AndrewEichmann-NOAA , you might need to issue a bugfix.

aerorahul:
@AndrewEichmann-NOAA @guillaumevernieres 
I don't think so.  This was merged yesterday and it looks right in `develop`.
```
export DATA=${DATA:-${DATAROOT}/${RUN}ocnanal_${cyc}}
```
I think the comment has been addressed.  Thank you for doing so.

PR #256, opened 2022-12-20
Simple stand alone application to run a suite of jjobs (soca oriented only for now)

guillaumevernieres:
## Description
As the title says. See description here: [g-w/1194](https://github.com/NOAA-EMC/global-workflow/issues/1194)

Given a yaml file, write and execute a script that would execute (sbatch or run in place) a sequence of jjobs. This works for `soca`, it could be generalized but I'm out of time for now. This PR is the first steps towards a more generic application. Once we can do the same thing with the ATM/Aero/Land, we'll move the code to the global-workflow.

The CI does not cover this code yet, maybe in the future, but soca is currently over tested ...


RussTreadon-NOAA (commented):
Seems OK but `run_jjobs.py` needs to be tested on Hera and Orion.   Eventually need to generalize beyond slurm to support other workload managers.    

Could a similar result be achieved via ctest?   For example, `ctest -R test test_gdasapp_atm_jjob_run` executes the atm 3dvar prep, run, and post jobs.   `DEPENDS` could be added to `CMakeList.txt` to enforce the desired sequence in which jobs are run.

This more a question than a suggestion.  I'm neither a cmake nor a ctest guru.

guillaumevernieres:
> Seems OK but `run_jjobs.py` needs to be tested on Hera and Orion. Eventually need to generalize beyond slurm to support other workload managers.
> 
> Could a similar result be achieved via ctest? For example, `ctest -R test test_gdasapp_atm_jjob_run` executes the atm 3dvar prep, run, and post jobs. `DEPENDS` could be added to `CMakeList.txt` to enforce the desired sequence in which jobs are run.
> 
> This more a question than a suggestion. I'm neither a cmake nor a ctest guru.

Noted for slurm and HPC @RussTreadon-NOAA . I'm selfishly answering the marine DA needs of the moment, which is to run soca stand alone to evaluate/develop our static B. This would be replacing what I added in `run_jedi_exe.py`.

Yes, it's the same process that we went through for the recent jjob ctests for the atm and soca, however, that work is hardwired to our ctests and to a resolution. I could see how we could use this new application to cleanup our tests, but I don't really see it as necessary for now. I need an application that can be used and easily manipulated through a yaml file to test various resolutions/configurations, similar to what run_jedi_exe.py does. What I don't understand yet is what parts need to be more generic to be able to implement it for other applications, but at this point I don't know if this is something that is needed?

guillaumevernieres:
> Seems OK but `run_jjobs.py` needs to be tested on Hera and Orion. Eventually need to generalize beyond slurm to support other workload managers.
> 
> Could a similar result be achieved via ctest? For example, `ctest -R test test_gdasapp_atm_jjob_run` executes the atm 3dvar prep, run, and post jobs. `DEPENDS` could be added to `CMakeList.txt` to enforce the desired sequence in which jobs are run.
> 
> This more a question than a suggestion. I'm neither a cmake nor a ctest guru.

@RussTreadon-NOAA , I added 3 `yaml` examples to run of `containers` and `hera` (C48-5deg/C384-0.25deg).

ush/soca/run_jjobs.py
```diff
@@ -0,0 +1,155 @@
...
+    # Dump the configs in a separate yaml file
+    with open("overwrite_defaults.yaml", "w") as f:
+        yaml.safe_dump(exp_config['setup_expt config'], f)
+
+    # Setup the experiment
+    bash_script.write("\n")
+    bash_script.write("# Setup the experiment\n")
+
+    setupexpt = "${HOMEgfs}/workflow/setup_expt.py cycled "
+    # Most of the args keys are not used to run the jjobs but are needed to run setup_expt.py
```
RussTreadon-NOAA, line 82: Valid comment.  Changes should be eventually made on the g-w side to simplify all this.   I think that's the goal of the yaml input file but it seems we still need to pass other arguments.   Hopefully this requirement will change in the future.

ush/soca/run_jjobs.py
```diff
@@ -0,0 +1,155 @@
...
+for key, value in var2replace.items():
+    newconfigbase = newconfigbase.replace(f'{key}=', f'{key}={value} #')
+with open(configbase, 'w') as f:
+    f.write(newconfigbase)
+
+# Generate the script that runs the jjobs
+with open("run_jjobs.sh", "w") as bash_script:
+    bash_script.write("#!/usr/bin/env bash\n")
+    if machine != "container":
+        slurm_header(exp_config['job options'], bash_script)
```
RussTreadon-NOAA, line 135: This is OK since at present we can only run JEDI on Hera and Orion.  Both systems use slurm.   We will need to generalize this when we can run JEDI on WCOSS2.  WCOSS2 uses PBS.

RussTreadon-NOAA (dismissed):
Minor comments - nothing requiring changes prior to approving this PR.

Approve as is.

ShastriPaturi (approved):
The fixes worked.
Thanks @guillaumevernieres. I also see the 1-hourly fgat files.

PR #265, opened 2023-01-06
Minor updates related soca running in the g-w

guillaumevernieres:
- Cleaned-up the bash function that creates the MOM6 increment from the JEDI/SOCA increment
- dump the MOM6 incremement in `COMOUT`
- added a full path to `r2d2_config.yaml`, for some reason not specifying the full path was causing problems. No clue why, it was fine before and the `CI` never picked up that issue. Adding the full path and moving on to more interesting problems

### DIFF
--- a/scripts/exgdas_global_marine_analysis_prep.py
+++ b/scripts/exgdas_global_marine_analysis_prep.py
@@ -171,7 +171,7 @@ ufsda.symlink(anl_out, os.path.join(anl_dir, 'Data'), remove=False)
 logging.info(f"---------------- Stage observations")
 
 # setup the archive, local and shared R2D2 databases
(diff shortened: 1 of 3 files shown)


AndrewEichmann-NOAA (commented):
Looks reasonable. At some point utility executables should probably be variable-ized in the manner of NLN, but that's a tale for a different PR

guillaumevernieres:
> Looks reasonable. At some point utility executables should probably be variable-ized in the manner of NLN, but that's a tale for a different PR

Yes, for sure ... Can I get a green check @AndrewEichmann-NOAA ?

AndrewEichmann-NOAA (approved):

PR #268, opened 2023-01-10
remove output before db test, tweaks to bufr2ioda test

AndrewEichmann-NOAA:
pre-cleans soca obsdb test, plus some refinement of bufr2ioda ctest

### DIFF
--- a/test/soca/test_bufr2ioda.sh
+++ b/test/soca/test_bufr2ioda.sh
@@ -7,10 +7,18 @@ OBSSOURCE=${2}
 CMD=${PROJECT_BINARY_DIR}/bin/bufr2ioda.x
 OBSYAML=${PROJECT_BINARY_DIR}/test/testinput/${OBSSOURCE}.yaml
 
-rm -fv ${PROJECT_BINARY_DIR}/test/testoutput/${OBSSOURCE}_20180415.nc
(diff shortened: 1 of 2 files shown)


test/soca/test_bufr2ioda.sh
```diff
@@ -4,13 +4,21 @@ set -x
...
+rm -v ${PROJECT_BINARY_DIR}/test/${OUTFILE}
+
 ${CMD} ${OBSYAML}
+rc=$?
 
+ls ${PROJECT_BINARY_DIR}/test/${OUTFILE}
+ra=$?
+rc=$((rc+ra))
 
+export err=$rc; err_chk
```
guillaumevernieres: `err_chk` is not defined. I don't know where this is coming from, global-workflow? I think It's not needed here.

AndrewEichmann-NOAA: It's not needed, but oddly enough it does work. I never did figure out where it came from.

guillaumevernieres:
looks like your prep and run step failed @AndrewEichmann-NOAA

AndrewEichmann-NOAA:
@guillaumevernieres Now I got it - I accidentally changed the DB root directory, which made the automatic tests fail, my build was still using the previously created DB in the correct directory, which is why prep and run worked for me. Example of the evil I'm trying to correct

guillaumevernieres (approved):
:+1:

PR #274, opened 2023-01-12
B-matrix diagnostics

### LINKED ISSUE #273: B-matrix impulse response
## Description
Diagnose the B-matrix used in the variational application, starting by looking at it's impulse response.

guillaumevernieres:
- fixes #273 

The next step is to create the associated j-job in the g-w.

### DIFF
--- a/parm/soca/berror/saber_blocks.yaml
+++ b/parm/soca/berror/saber_blocks.yaml
@@ -26,7 +26,7 @@ saber outer blocks:
     input fields:
     - parameter: StdDev
       file:
-        read_from_file: 1
+        read_from_file: 3
         basename: ./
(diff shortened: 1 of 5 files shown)


AndrewEichmann-NOAA (approved):

guillaumevernieres:
@RussTreadon-NOAA , @CoryMartin-NOAA : I added a python scripts, under `ush/ufsda`, that will generate a "dirac" yaml file from the var.yaml. Is that the correct location? Should it be under `ush`?

ShastriPaturi (approved):
Looks Good.

CoryMartin-NOAA:
@guillaumevernieres sorry I'm late seeing this, I would suggest not under `ufsda` as that is a python module (i.e. you `import ufsda`) but this is fine for now. The `ufsda` directory is likely going away at some point when we migrate to workflow-based tools for YAML, stage, etc.

PR #297, opened 2023-01-26
Use new convention for AOD

CoryMartin-NOAA:
Change `aerosol_optical_depth` to `aerosolOpticalDepth` in 2 YAML files


andytangborn (approved):
Looks good.

PR #301, opened 2023-01-26
Generalize run_ufo_hofx_test.sh

ADCollard:
Generalized run_ufo_hofx_test.sh to work with both orion and hera and to input instrument name

### DIFF
--- a/ush/ufoeval/run_ufo_hofx_test.sh
+++ b/ush/ufoeval/run_ufo_hofx_test.sh
@@ -39,19 +39,22 @@ exename=test_ObsFilters.x
 #-------------- Do not modify below this line ----------------
 # paths that should only be changed by an expert user
 
+dataprocdate=20230126 # Production date of test data
+
 obtype_short=${obtype:0:4}
 if [ $obtype_short = "cris" ] || [ $obtype_short = "iasi" ] || [ $obtype_short = "hirs" ] || [ $obtype_short = "sevi" ] || \
-   [ $obtype_short = "avhr" ] || [ $obtype_short = "mhs_" ] || [ $obtype_short = "ssmi" ]; then
+   [ $obtype_short = "avhr" ] || [ $obtype_short = "mhs_" ] || [ $obtype_short = "ssmi" ] || [ $obtype_short = "amsu" ] || \
+   [ $obtype_short = "atms" ]; then
    radiance="YES"
 else 
    radiance="NO"
 fi
 
 if [ $machine = orion ]; then
-    export Datapath='/work2/noaa/da/cmartin/UFO_eval/data/gsi_geovals_l127/nofgat_aug2021/20220816' 
+    export Datapath='/work2/noaa/da/cmartin/UFO_eval/data/gsi_geovals_l127/nofgat_aug2021/'$dataprocdate 
     FixDir=/work2/noaa/da/cmartin/GDASApp/fix
 elif [ $machine = hera ]; then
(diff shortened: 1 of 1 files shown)


ush/ufoeval/run_ufo_hofx_test.sh
```diff
@@ -10,21 +10,57 @@
...
 
 #-------------- Do not modify below this line ----------------
 # paths that should only be changed by an expert user
-GeoDir=/work2/noaa/da/cmartin/UFO_eval/data/gsi_geovals_l127/nofgat_aug2021/20220816/geovals/
-ObsDir=/work2/noaa/da/cmartin/UFO_eval/data/gsi_geovals_l127/nofgat_aug2021/20220816/obs/
-BCDir=/work2/noaa/da/cmartin/UFO_eval/data/gsi_geovals_l127/nofgat_aug2021/20220816/bc/
-FixDir=/work2/noaa/da/cmartin/GDASApp/fix/
+
+obtype_short=${obtype:0:4}
+if [ $obtype_short = "cris" ] || [ $obtype_short = "iasi" ] || [ $obtype_short = "hirs" ] || [ $obtype_short = "sevi" ] || \
```
CoryMartin-NOAA, line 45: amsu? atms?

ush/ufoeval/run_ufo_hofx_test.sh
```diff
@@ -10,21 +10,57 @@
...
+   [ $obtype_short = "avhr" ] || [ $obtype_short = "mhs_" ] || [ $obtype_short = "ssmi" ]; then
+   radiance="YES"
+else 
+   radiance="NO"
+fi
+
+if [ $machine = orion ]; then
+    export Datapath='/work2/noaa/da/cmartin/UFO_eval/data/gsi_geovals_l127/nofgat_aug2021/20220816'
```
CoryMartin-NOAA: Use the new date on this one too, and also perhaps something like `$dataprocdate` that can be used in both of these?

ush/ufoeval/run_ufo_hofx_test.sh
```diff
@@ -76,10 +118,14 @@ config:
 EOF
 $GDASApp/ush/genYAML --config $workdir/temp.yaml
```
CoryMartin-NOAA, line 122: While we are modifying this script, what about a check for exit code after each key step so that it quits and doesn't keep going? or is that overkill?

ush/ufoeval/run_ufo_hofx_test.sh
ADCollard: I guess I did not finish setting this up.  I will add more instruments, we just need to ensure all in the test yamls are represented here.

ADCollard: I only added ATMS and AMSU-A.  I suggest if any more are missing we add them as the YAMLs are added

ush/ufoeval/run_ufo_hofx_test.sh
ADCollard: Done

ush/ufoeval/run_ufo_hofx_test.sh
ADCollard: Not overkill.  Very useful not to have EVA just ploughing ahead.  

Done

ADCollard:
@CoryMartin-NOAA Are my changes OK with you?

CoryMartin-NOAA (approved):
Thanks @ADCollard

PR #312, opened 2023-02-01
Update sfc, sfcship and sondes_ps

ADCollard:
Update sfc, sfcship and sondes_ps

run_ufo_hofx_test.sh <yaml> does not have any IODA errors for these files.


CoryMartin-NOAA (approved):
¡muy bien!

PraveenKumar-NOAA (approved):
Looks good.

PR #323, opened 2023-02-03
testing/satwind.yaml fixed for passing GDASApp/UFO 

BrettHoover-NOAA:
The variables `qualityInformationWithoutForecast` and `windComputationMethod` are used in a set of 5 satwind YAML filters, each as part of a `Bounds Check`. The updated IODA conventions now produces these variables as integer-values rather than float-values, which apparently breaks `Bounds Check` since it requires float-values.

The fix here is to use `Perform Action` instead of `Bounds Check` and flip the filtering conditions, e.g.:
```
  # Reject obs with qualityInformationWithoutForecast < 85
  - filter: Bounds Check
   ...
    test variables:
    - name: MetaData/qualityInformationWithoutForecast 
    minvalue: 85.
    maxvalue: 100.
    action:
      name: reject
```
becomes:
```
  # Reject obs with qualityInformationWithoutForecast < 85
  # CLEARED
  - filter: Perform Action
  ...
    test variables:
    - name: MetaData/qualityInformationWithoutForecast 
    is_not_in: 85-100
    action:
      name: reject
```
The other significant change made to the YAML is to replace the `minvalue maxvalue` formulation of subset-filters for `satelliteIdentifier` with `is_in` formulation instead. This is done because `satelliteIdentifier` is now also an integer-value, and while it is not explicitly necessary to make this change to get the YAML filters to work, it is a more explicit formulation for making it clear that `satelliteIdentifier` is an integer-value and reduces the number of lines in the YAML file.


ADCollard (approved):
Looks good!  Thanks!

ADCollard:
@CoryMartin-NOAA If you would like to merge this, I will proceed with a pull request for the full [feature/sprint-ioda-converters](https://github.com/NOAA-EMC/GDASApp/tree/feature/sprint-ioda-converters) branch.

CoryMartin-NOAA:
@ADCollard I defer to you, if you think we are ready to merge to develop, let's do it

PR #329, opened 2023-02-06
Fixed broken Perform Action filters

BrettHoover-NOAA:
Small update to the parm/atm/obs/testing/satwind.yaml file - there were a few filters changed from `Bounds Check` formulation to `Perform Action` formulation in the last PR that ended up not working properly, apparently due to the use of `test variables:` in `Bounds Check` not translating to `Perform Action`. The effect appeared to be to skip the filter being applied to `test variables` entirely, but not generate an error.

These filters have been modified to exclusively use the `where:` formulation for subsetting observations to reject in a `Perform Action` filter. I have tested this change and I can now re-create the level of GSI/UFO acceptance that we had prior to updating IODA variable name conventions.


BrettHoover-NOAA:
Requesting a review on this just to make sure I'm not screwing anything up with your plans with this feature branch.

ADCollard (approved):
Good catch, @BrettHoover-NOAA

CoryMartin-NOAA (approved):

PR #338, opened 2023-02-07
Feature/obs tests

ADCollard:
This PR fixes the EVA yaml for radiances, adds flexibility to [ush/ufoeval/run_ufo_hofx_test.sh](https://github.com/NOAA-EMC/GDASApp/blob/feature/obs_tests/ush/ufoeval/run_ufo_hofx_test.sh) to run a reduced EVA suite or none at all; and introduces the [test_yamls.sh](https://github.com/NOAA-EMC/GDASApp/blob/feature/obs_tests/ush/ufoeval/test_yamls.sh) script which loops through the YAMLs in [parm/atm/obs/testing](https://github.com/NOAA-EMC/GDASApp/tree/feature/obs_tests/parm/atm/obs/testing) and reports if they pass.

### DIFF
--- a/ush/ufoeval/run_ufo_hofx_test.sh
+++ b/ush/ufoeval/run_ufo_hofx_test.sh
@@ -14,10 +14,11 @@
 usage() {
   set +x
   echo
-  echo "Usage: $0 instrument [cycle] [-x] [-s] [-h]"
+  echo "Usage: $0 [-c cycle] [-x] [-s] [-h] instrument [cycle]"
   echo
-  echo "  -x  run eva DEFAULT=YES"
-  echo "  -s  just produce eva stats plots DEFAULT=YES"
+  echo "  -c  cycle to run DEFAULT=2021080100"
+  echo "  -x  don't run eva DEFAULT=run eva"
+  echo "  -s  just produce eva stats plots DEFAULT=produce lots of plots"
   echo "  -h  display this message and quit"
   echo
   exit 1
@@ -25,38 +26,36 @@ usage() {
 
 # ==============================================================================
 
-
-#--------------- User modified options below -----------------
-obtype=$1
-cycle=${2:-2021080100}
-
-#if [ $# -lt 1 ] || [ $# -gt 2 ]; then
-#   echo "Incorrect number of arguments"
-#   echo "Usage:"
-#   echo $0 instrument [cycle]
-#   exit 1
-#fi
-
-RUN_EVA=YES
-EVA_STATS_ONLY=NO
+cycle=2021080100
+run_eva=YES
+eva_stats_only=NO
  
-while getopts "hsx" opt; do
-  echo $opt
+while getopts "c:hsx" opt; do
   case $opt in
+    c)
+      cycle=$OPTARG
+      ;;
     x)
-      RUN_EVA=NO 
+      run_eva=NO 
       ;;
     s)
-      EVA_STATS_ONLY=YES
+      eva_stats_only=YES
       ;;
     h|\?|:)
       usage
       ;;
   esac
 done
+shift $((OPTIND - 1))
 
-echo "RUN_EVA, EVA_STATS_ONLY " $RUN_EVA $EVA_STATS_ONLY
-exit 0
+if [ $# -lt 1 ] || [ $# -gt 2 ]; then
+   echo "Incorrect number of arguments"
+   usage 
+fi
+
+obtype=$1
+
+#--------------- User modified options below -----------------
 
 machine=orion
 #GDASApp=/scratch1/NCEPDEV/da/$LOGNAME/git/GDASApp/ # Change this to your own branch
@@ -172,9 +171,20 @@ cd $workdir
 
 if [ $? -ne 0 ]; then
    echo "*************** Running UFO failed ****************"
-   echo "Running EVA for diagnostic purposes"
+   RC=1
+   if [ $run_eva == YES ]; then
+       echo "Running EVA for diagnostic purposes"
+   else
+       exit 1
+   fi
 else
    echo "Running EVA"
+   RC=0
+   if [ $run_eva == YES ]; then
+       echo "Running EVA"
+   else
+       exit 0
+   fi
 fi
 
 # Load EVA modules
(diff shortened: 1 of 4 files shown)


ADCollard:
Example output:

```
(gdasapp) Orion-login-2[151] acollard$ ./test_yamls.sh
aircraft.yaml Fails (boo!)
amsua_n19.yaml Fails (boo!)
satwind.yaml Fails (boo!)
sfcship.yaml Fails (boo!)
sfc.yaml Fails (boo!)
sondes_ps.yaml Fails (boo!)
sondes.yaml Fails (boo!)
ssmis_f17.yaml Fails (boo!)
ssmis_f18.yaml Fails (boo!)
```

guillaumevernieres (dismissed):
:+1: for the `boo` \ `yay` output, which is admittedly the only thing I have reviewed in this PR. 
Are you planning to add the tests to the `ci` eventually?

ADCollard:
> 👍 for the `boo` \ `yay` output, which is admittedly the only thing I have reviewed in this PR. Are you planning to add the tests to the `ci` eventually?

I believe @CoryMartin-NOAA is keep to incorporate this into the CI.

ush/ufoeval/run_ufo_hofx_test.sh
```diff
@@ -9,18 +9,55 @@
...
 #-------------------------------------------------------------
-#--------------- User modified options below -----------------
-obtype=$1
-cycle=${2:-2021080100}
+
+# ==============================================================================
+usage() {
+  set +x
+  echo
+  echo "Usage: $0 [-c cycle] [-x] [-s] [-h] instrument [cycle]"
```
CoryMartin-NOAA: Is this usage statement correct? Why `[cycle]` at the end?

ush/ufoeval/run_ufo_hofx_test.sh
```diff
@@ -9,18 +9,55 @@
...
+      eva_stats_only=YES
+      ;;
+    h|\?|:)
+      usage
+      ;;
+  esac
+done
+shift $((OPTIND - 1))
 
 if [ $# -lt 1 ] || [ $# -gt 2 ]; then
```
CoryMartin-NOAA: can you not do instrument, cycle, and an eva option? that would be 3?

ush/ufoeval/run_ufo_hofx_test.sh
```diff
@@ -9,18 +9,55 @@
...
-   exit 1
+   usage 
 fi
 
-machine=hera
+obtype=$1
+
+#--------------- User modified options below -----------------
+
+machine=orion
```
CoryMartin-NOAA: for machine and GDASApp can we do:
`machine=${machine:-orion}` and `GDASApp=${GDASApp:-/path/to/stuff}`? that way an external script can call this directly without needing to edit it? (thinking for CI)

ush/ufoeval/test_yamls.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+# test_yamls.sh 
+# This scripts runs run_ufo_hofx_test.sh for the input YAMLs in ../../parm/atm/obs/testing
+# and outputs pass/fail information
+
+# Work directory
+WORKDIR=$STMP/test_yamls
```
CoryMartin-NOAA: what if the user does not have `$STMP` defined?

ush/ufoeval/run_ufo_hofx_test.sh
```diff
@@ -133,17 +170,32 @@ cd $workdir
...
    echo "*************** Running UFO failed ****************"
-   echo "Running EVA for diagnostic purposes"
+   RC=1
+   if [ $run_eva == YES ]; then
+       echo "Running EVA for diagnostic purposes"
+   else
+       exit 1
+   fi
 else
    echo "Running EVA"
```
CoryMartin-NOAA: Should this echo here? seems it should only do so if the `if` is met below

CoryMartin-NOAA:
> > +1 for the `boo` \ `yay` output, which is admittedly the only thing I have reviewed in this PR. Are you planning to add the tests to the `ci` eventually?
> 
> I believe @CoryMartin-NOAA is keep to incorporate this into the CI.

Regarding implementing this as testing. What do we all think, should this run for every PR as part of the "workflow testing"? Or should it be a bit more sparse and only run daily on develop? I'm fine either way but suspect it might take a nontrivial amount of time to run if it is a `ctest`.

ush/ufoeval/run_ufo_hofx_test.sh
ADCollard: @CoryMartin-NOAA I got a little confused here.  But it should just be 
```
if [ $# -ne 1 ]; then
```

The `shift` takes care of the rest.

Unless I misunderstand what you are asking

ush/ufoeval/test_yamls.sh
ADCollard: The intention was for this line to be edited.  What would you suggest?

ush/ufoeval/run_ufo_hofx_test.sh
CoryMartin-NOAA: the "or greater than 2" part was what I was concerned about

ush/ufoeval/run_ufo_hofx_test.sh
ADCollard: This is not correct (there were many iterations!): `[cycle]` removed

ush/ufoeval/run_ufo_hofx_test.sh
CoryMartin-NOAA: But I see what you're saying, I've never used this like this but it is basically treating the args differently. As long as it works, it's good with me.

ush/ufoeval/test_yamls.sh
CoryMartin-NOAA: This is fine, just was being thorough with the review.

ush/ufoeval/run_ufo_hofx_test.sh
ADCollard: Correct.  Removing line.

CoryMartin-NOAA (approved):
thanks @ADCollard for this and addressing comments!

ush/ufoeval/run_ufo_hofx_test.sh
ADCollard: I made the changes.  Hope you like them.

PR #351, opened 2023-02-12
Feature/eva in soca vrfy

AndrewEichmann-NOAA:
Partially addresses https://github.com/NOAA-EMC/GDASApp/issues/202. Adaptation of feature/eva_emcpy that creates new marine-specific gen_eva_obs_yaml.py -- there are ways to have marine-specific if-then blocks in the original, but I want to at least present a different approach before merging them. One problem is that there appears to be no non-hacky way to determine that the input yaml is a marine yaml which would require different handling - I would expect this could be extracted from the input yaml. 
One twist is that is assumes that the diag files are being read from the COMROT, and so this version expects that to be defined in the environment. Extracting the diag file filenames from the yaml seemed much preferable to alternate methods.

Files added:
ush/eva/gen_eva_obs_yaml_marine.py
ush/eva/marine_gdas_plots.yaml


AndrewEichmann-NOAA:
Don't quite get why tests would fail on this

guillaumevernieres:
That's not what we talked about @AndrewEichmann-NOAA :
1 - Modify the original `gen_eva_obs_yaml.py` so it can handle the `cost function` key, this was mostly done by @mjhossen , so start from his branch.
2 - Write a separate application to reprocess the yamls generated by the script to adjust things like the the path to the diag files and whatever else we see fit. 
3 - We should come up with a simple test for that application too.

guillaumevernieres:
FYI @AndrewEichmann-NOAA , this is @mjhossen 's old branch: `feature/eva_emcpy` .

AndrewEichmann-NOAA:
Closing, supplanted by https://github.com/NOAA-EMC/GDASApp/pull/352

PR #355, opened 2023-02-13
Add script to post-process marine eva files

AndrewEichmann-NOAA:
Currently this takes a directory of yaml files from gen_eva_obs_yaml.py and the template marine_gdas_plots.yaml and the diag file directory, corrects the yamls with the proper diag file paths, and then outputs to another directory. It is anticipated that this will perform other, similar tasks.

File added:
ush/eva/marine_eva_post.py


ush/eva/marine_eva_post.py
```diff
@@ -0,0 +1,55 @@
...
+import logging
+import os
+import socket
+import yaml
+import glob
+
+
+def marine_eva_post(inputdir, outputdir, diagdir):
+    logging.basicConfig(format='%(asctime)s:%(levelname)s:%(message)s', level=logging.INFO, datefmt='%Y-%m-%d %H:%M:%S')
+    inputyamls = glob.glob(os.path.join(inputdir, '*.yaml'))
```
guillaumevernieres: Let's make that function operate on 1 yaml only, while diagdir is going to be common to all, that isn't true for the bounds for example. Simpler to call this function with specific argument for a specific obs type.

guillaumevernieres (commented):
Getting there @AndrewEichmann-NOAA . Just one minor comment.

ush/eva/marine_eva_post.py
AndrewEichmann-NOAA: Ok, sounds safer

guillaumevernieres (commented):
looks good, but I'm tempted to ask for the addition of a bit of testing ...

AndrewEichmann-NOAA:
That might be a story for different PR

guillaumevernieres (approved):
:tada:

PR #359, opened 2023-02-15
Add one more summary plot

ADCollard:
Add a plot to show channel versus the statistics of the final obs error differences between JEDI and GSI.


CoryMartin-NOAA (approved):

CoryMartin-NOAA:
Thanks ADC, will merge now.

PR #366, opened 2023-02-23
Fix stable nightly cron

CoryMartin-NOAA:
Listening to @guillaumevernieres was finally a good idea :-)

By using a fresh branch each time, this seems to have fixed the weird SOCA issue.

This PR will now clone a fresh build of develop, get the latest hashes, attempt to build with them, and then run the tests and update if necessary


ci/stable_driver.sh
```diff
@@ -56,11 +56,14 @@ mkdir -p $stableroot/$datestr
 cd $stableroot/$datestr
 git clone $repo_url
 cd GDASApp
-git checkout feature/stable-nightly
 
 # ==============================================================================
-# merge in develop
-git merge develop
+# run ecbuild to get the repos cloned
```
RussTreadon-NOAA, line 61: Not a request for change but rather information.  What is the purpose of this section of the script?   `stable_driver.sh` executes `run_ci.sh`   Script `run_ci.sh` executes `build.sh`.   Won't `build.sh` clone and build the JEDI repos required by GDASApp?

guillaumevernieres (approved):
:tada:

ci/stable_driver.sh
CoryMartin-NOAA: This clones the repos so that one can run the command on them to get the most recent commits. The ecbuild command is mostly equivalent here to the `./checkout.sh` command in the workflow. There are probably other ways to do this but this was the path of least resistance.

RussTreadon-NOAA: Thank you for the explanation.   We're not just building GDASApp.  We're checking if GDASApp builds with the current head of repo branches required by GDASApp.

RussTreadon-NOAA (approved):
Approve.

PR #370, opened 2023-03-02
Fixing satwind.yaml to treat QI variables as floats

BrettHoover-NOAA:
All this changes is switching the satwind.yaml filters operating on `qualityInformationWithForecast` and `qualityInformationWithoutForecast` to treating these variables as float values. This requires changing from a `Perform Action` filter to a `Bounds Check` filter and providing the QI cutoffs as `minvalue` and `maxvalue` options, rather than using the `is_not_in` option previously used when these variables were treated as integers.

UFO satwind acceptance is unchanged by this update.


CoryMartin-NOAA (approved):

CoryMartin-NOAA:
@ADCollard is this okay to merge in?

PR #372, opened 2023-03-03
Test "genYAML" to generate the yaml file for aero DA

andytangborn:
This PR adds a test to build an aero DA yaml, following the form of the land DA yaml test.


ush/ufsda/genYAML.py
```diff
@@ -72,7 +72,7 @@ def get_runtime_config(config_dict):
     bkg_tstep = str(config_dict.get('BKG_TSTEP', f"PT{assim_freq}H"))
     component_dict = {
         'atmos': 'ATM',
-        'chem': 'AERO',
```
CoryMartin-NOAA: this should stay chem because that is the COMROT structure for NCO.

ush/ufsda/yamltools.py
```diff
@@ -132,6 +136,35 @@ def fv3land_geom_dict(case, levs, ntiles, layout, io_layout):
     }
     return outdict
 
+def fv3aero_geom_dict(case, levs, ntiles, layout, io_layout):
```
CoryMartin-NOAA, line 140: what is different here besides the fieldmetadata file? I don't think we need the 'time invariant' part, right?

test/CMakeLists.txt
```diff
@@ -175,4 +189,6 @@ endif()
 add_subdirectory(atm)
 
 # gdas aerosol tests
-add_subdirectory(aero)
+if (${BUILD_GDASBUNDLE})
```
CoryMartin-NOAA: why this change in this PR?

CoryMartin-NOAA:
@andytangborn a few comments on the code, plus all the tests are failing...

andytangborn:
Ok, but it failed when it was "chem" it gave a key error "aero".  So that
wasn't the right fix then.

ush/ufsda/genYAML.py
andytangborn: I'm getting a key error "aero" otherwise. Changing this to "aero" fixed it. So I need another solution.

ush/ufsda/yamltools.py
andytangborn: I wasn't sure. I can take it out.

test/CMakeLists.txt
andytangborn: If it's not needed, I can remove it.

ush/ufsda/yamltools.py
jiaruidong2017: Yes, the `time invariant` part is land specific.

ush/ufsda/genYAML.py
jiaruidong2017: In `test/genYAML_prep_aero.sh`, try to replace `COMPONENT: aero` with `COMPONENT: chem`.

CoryMartin-NOAA: in `test/genYAML_prep_aero.sh` what if you change it to `chem`? Does it still error? The directory structure will be like `gdas.20230303/12/chem` so we need it to be chem.

test/CMakeLists.txt
CoryMartin-NOAA: It's mainly that if you aren't doing anything to `test/aero` then we shouldn't change that `add_subdirectory` part right now

test/genYAML_prep_aero.sh
```diff
@@ -0,0 +1,24 @@
...
+  layout_x: '8'
+  layout_y: '8'
+  OBS_LIST: ${srcdir}/parm/aero/obs/lists/gdas_aero_prototype.yaml
+  OPREFIX: gdas.t06z.
+  CDATE: '2019061400'
+  PDY: '20190614'
+  cyc: '06'
+  valid_time: '2019-06-14T06:00:00Z'
+  aero_window_length: PT6H
+  CASE: 'C48'
```
CoryMartin-NOAA: isn't this C96 in your sample case?

andytangborn: Yes. I'll change this.

andytangborn: There's also a CASE ENKF 48 below that. Do we even need that since this is for 3dvar?

CoryMartin-NOAA: not really, although that is often (probably incorrectly) used as the analysis resolution

test/CMakeLists.txt
CoryMartin-NOAA: Yeah I would revert this change in this PR, just so that we only touch it later when we add 3dvar or hofx tests

andytangborn: done.

jiaruidong2017 (approved):

CoryMartin-NOAA (approved):

PR #383, opened 2023-03-09
Feature/ufoeval adc

ADCollard:
Fix eva YAMLs to be consistent with changes in [this issue](https://github.com/JCSDA-internal/eva/pull/80).  Bug fix to `ush/ufoeval/run_ufo_hofx_test.sh`


ush/eva/jedi_gsi_compare_conv.yaml
```diff
@@ -279,18 +279,20 @@ diagnostics:
...
+                yloc: -0.10
+                kwargs:
+                  fontsize: 6
+            statistics_variables:
             - n
             - min
             - mean
             - max
             - std
             - name
```
CoryMartin-NOAA: this one still has `name` does that work?

ADCollard: > this one still has `name` does that work?

@CoryMartin-NOAA Full disclosure, I was not able to test this YAML as none of the conventional YAMLs are giving valid output file right now.  So this is a bug.

I'm changing this to a draft PR until I find a way to test this one.

But yes, `name` needs to be removed.

ush/eva/jedi_gsi_compare_conv.yaml
```diff
@@ -368,18 +371,19 @@ diagnostics:
         - add_xlabel: 'h(x) JEDI - h(x) GSI'
           add_ylabel: 'Count'
           statistics:
-            data:
-              variable: experiment::HofXDiff::${variable}
-              @CHANNELKEY@
+            fields:
+              - field_name: experiment::HofXDiff::${variable}
+                @CHANNELKEY@
```
nicholasesposito, line 375: Looking over the previous commit that @CoryMartin-NOAA showed me yesterday, it looks like this should be 
`                channel: ${channel}`
but I might be wrong. What is the preferred way to do this?

CoryMartin-NOAA: This is because the python script decides to put that in if it needs to

CoryMartin-NOAA (approved):

ADCollard:
I have fixed the issues with issues with the conventional eva yaml and tidied up directory structure.  Also updated the date for testing in `ush/ufoeval/run_ufo_hofx_test.sh`.

Ready for review.

PR #389, opened 2023-03-10
Add OBS_LIST to handle different observations in different cycles.

jiaruidong2017:
Add the OBS_LIST below to handle the snow DA with different observations in the different cycles in the global-workflow. 

- gdas_t00z_land_prototype.yaml
- gdas_t06z_land_prototype.yaml
- gdas_t12z_land_prototype.yaml
- gdas_t18z_land_prototype.yaml


CoryMartin-NOAA (changes requested):
Rather than 3 identical lists and a 4th, how about two lists:
1) gdas_land_adpsfc_only.yaml
2) gdas_land.yaml
(or something like that)
and let the script decide based on date

jiaruidong2017:
The configure file `parm/config/config.landanl` in the global-workflow is used to select the observation list as below: 

```
export OBS_LIST=${HOMEgfs}/sorc/gdas.cd/parm/land/obs/lists/gdas_t${cyc}z_land_prototype.yaml
```
It will be easier to have four files by using `t${cyc}z` as above to identify the different four cycles.

CoryMartin-NOAA:
@jiaruidong2017 `$cyc` should already be defined before `config.landanl` is sourced. Therefore you can do something like:
```
OBS_DIR=foo
if [[ "$cyc" == "18" ]]; then
    OBS_DIR=bar
fi
```
There is precedence for stuff like this, like in `config.anal`

jiaruidong2017:
Thanks @CoryMartin-NOAA . I checked and found that the `${cyc}` together with `${HOMEgfs}` was already defined before `config.landanl` is sourced in the global-workflow. 

Sorry, I didn't understand your suggestions below: 

```
OBS_DIR=foo
if [[ "$cyc" == "18" ]]; then
    OBS_DIR=bar
fi
There is precedence for stuff like this, like in config.anal
```
Would you please explain a little more? Thanks.

CoryMartin-NOAA:
@jiaruidong2017 which part do you need explained further?

jiaruidong2017:
@CoryMartin-NOAA Do you want me to do something like below in the `config.landanl`?
```
OBS_LIST=${HOMEgfs}/sorc/gdas.cd/parm/land/obs/lists/gdas_land_adpsfc_only.yaml
if [[ "$cyc" == "18" ]]; then
    OBS_LIST=${HOMEgfs}/sorc/gdas.cd/parm/land/obs/lists/gdas_land.yaml
fi
```

CoryMartin-NOAA:
@jiaruidong2017 yes

jiaruidong2017:
Okay, I will try this in the global-workflow and let you know then. 
I confirmed that my previous approach worked in the global -workflow.

jiaruidong2017:
@CoryMartin-NOAA I confirmed that it worked in the global-workflow by adding the `gdas_land_adpsfc_only.yaml` for t00z, t06z and t12z cycles  and using original `gdas_land_prototype.yaml` for t18z cycle as below: 

```
obs_list_name=gdas_land_adpsfc_only.yaml
if [[ "${cyc}" == "18" ]]; then
    obs_list_name=gdas_land_prototype.yaml
fi

export OBS_LIST=${HOMEgfs}/sorc/gdas.cd/parm/land/obs/lists/${obs_list_name}
```

CoryMartin-NOAA (approved):

PR #394, opened 2023-03-14
Non Agile PR ... Sorry. A few fixes to make the cycling work properly 

### LINKED ISSUE #393: Remove empty obs spaces
### Description
What we have in `obs_list.yaml` is what we wish to assimilate/simulate but the staged obs will rarely have the entire wish list available. The result is the jedi exe failing.
 
### Solution
There are multiple ways to do this. I'm thinking about culling the obs space based on what was actually staged. Any other suggestions @RussTreadon-NOAA , @CoryMartin-NOAA , @AndrewEichmann-NOAA , ... and others ...

A simple python application:
```
input: 
 - var/hofx/letkf.yaml
 - path to where the obs were staged
output:
- new var.yaml with obs spaces pointing to non-existent files removed
```

guillaumevernieres:
### Description
I could split that PR into multiple simpler PR's if the reviewers insist ... But I don't think it's too bad. 22 files touched, most of which are updates to obs yamls.

### Work done
- Updates to a few obs yamls (sst and adt) to allow cycling with a realistic number of observations
- I added a subclass to `YAMLFile` that adds one method that removes the obs spaces that are pointing to a ioda file that does not exists. Not exactly what we talked about @CoryMartin-NOAA , but I couldn't figure out a simple way to re-generate the obs yaml list after the staging of the obs. fixes #393 
- Added one obs yaml that will point to a non existent file in the test obs list to test the above
- Fixed the dumb way I was selecting indices for the dirac locations
- temporary removal of bump and (u,v)


parm/soca/obs/obs_list.yaml
```diff
@@ -4,3 +4,4 @@ observers:
 - !INC ${OBS_YAML_DIR}/sss_smap.yaml
 - !INC ${OBS_YAML_DIR}/sst_noaa19_l3u.yaml
 - !INC ${OBS_YAML_DIR}/icec_emc.yaml
+- !INC ${OBS_YAML_DIR}/adt_3a_egm2008.yaml
```
CoryMartin-NOAA: no end of line!

scripts/exgdas_global_marine_analysis_prep.py
```diff
@@ -357,13 +357,16 @@ def find_bkgerr(input_date, domain):
     logging.info(f"using default SABER blocks yaml")
     os.environ['SABER_BLOCKS_YAML'] = os.path.join(gdas_home, 'parm', 'soca', 'berror', 'saber_blocks.yaml')
 
+# substitute templated variables in the var config
 logging.info(f"{config}")
-varconfig = YAMLFile(path=var_yaml_template)
+varconfig = ufsda.yamltools.DAYAMLFile(path=var_yaml_template)
```
CoryMartin-NOAA, line 362: This is a step backwards IMO, we are going towards the `pygw` YAML tools and this goes back to the old `ufsda` tools...

ush/ufsda/yamltools.py
```diff
@@ -1,9 +1,48 @@
...
 from solo.yaml_file import YAMLFile
 from solo.template import TemplateConstants, Template
 from ufsda.misc_utils import isTrue
+import pygw
+
+logging.basicConfig(format='%(asctime)s:%(levelname)s:%(message)s',
+                    level=logging.INFO, datefmt='%Y-%m-%d %H:%M:%S')
+
+
+class DAYAMLFile(pygw.yaml_file.YAMLFile):
```
CoryMartin-NOAA, line 14: if you are going to go this route, I suggest a "cleanup" function that removes the YAML pieces if the file doesn't exist that runs on an existing YAML file after it is created. However this is not what I would suggest as a permanent solution.

I understand this workaround is due to using R2D2... but we aren't going to be using R2D2 in GFS/GDAS at least not initially. So should we be moving away from R2D2 sooner rather than later. Should this be the catalyst? 

I would much rather create the `obs_list` on the fly based on available observations and then go create the YAML file. What about creating a R2D2 YAML as an intermediate if you wish to go that route?

scripts/exgdas_global_marine_analysis_prep.py
guillaumevernieres: I don't see why @CoryMartin-NOAA , we're clearly in need of DA specific yaml manipulation tools, so a subclass of YAMLFile seems appropriate.

ush/ufsda/yamltools.py
CoryMartin-NOAA: Just to clarify @guillaumevernieres I am okay with this for the short term, but I think we need a more elegant solution for all components eventually (or frankly, fix JEDI to not error if the obs are missing...)

guillaumevernieres:
Closing this, I'll resubmit a new PR with your suggested changes @CoryMartin-NOAA .

PR #399, opened 2023-03-22
minor tweaks to autodetect machine and purge/reset modules correctly

aerorahul:
This PR has minor updates to `build.sh` to bring it up to being consistent with other GFS components such as the ufs-weather-model, GSI, etc.

Specifically,
- auto detects the machine the build is for
- purges/resets the modules based on the machine

### DIFF
--- a/ush/module-setup.sh
+++ b/ush/module-setup.sh
@@ -1,0 +1,127 @@
+#!/bin/bash
+set -eu
+
+if [[ $MACHINE_ID = jet* ]] ; then
+    # We are on NOAA Jet
+    if ( ! eval module help > /dev/null 2>&1 ) ; then
+        source /apps/lmod/lmod/init/bash
+    fi
+    module purge
+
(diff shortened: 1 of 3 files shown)


ush/module-setup.sh
```diff
@@ -0,0 +1,127 @@
...
+    module purge
+
+elif [[ $MACHINE_ID = s4* ]] ; then
+    # We are on SSEC Wisconsin S4
+    if ( ! eval module help > /dev/null 2>&1 ) ; then
+        source /usr/share/lmod/lmod/init/bash
+    fi
+    module purge
+
+elif [[ $MACHINE_ID = wcoss_cray ]] ; then
```
CoryMartin-NOAA: I know this is a copy/paste, but might as well not include wcoss_cray and dell, etc.

aerorahul: that it was indeed.
removed.

CoryMartin-NOAA (approved):

Issue #403, opened 2023-03-23
Incorrect directory names in GDASApp/fix/gsibec


RussTreadon-NOAA:
According to logic in the following g-w scripts
- `scripts/exglobal_atmos_analysis.sh`
- `scripts/exglobal_atmos_sfcanl.sh`
- `scripts/exgdas_enkf_sfc.sh`
- `ush/forecast_postdet.sh`
- `ush/forecast_predet.sh`
- `ush/gaussian_sfcanl.sh`

we compute the number of longitude and latitudes points on the gaussian grid as
```
  res=$(echo $CASE |cut -c2-5)
  resp=$((res+1))
  npx=$resp
  npy=$resp
  LONB_CASE=$((4*res))
  LATB_CASE=$((2*res))
```
Given this, 
- C768 corresponds to 3072 x 1536 (lon, lat)
- C384 corresponds to 1536 x 768
- C192 corresponds to 768 x 384
- C96 corresponds to 384 x 192
- C48 corresponds to 192 x 96

The above mappings between the cube sphere resolution (`CASE`) and the number of gaussian grid longitude and latitude points is at odds with what we find in `gsibec` directories.    

Looking in `/scratch1/NCEPDEV/da/Cory.R.Martin/GDASApp/fix/gsibec` we have
- C768 contains 1536 x 768
- C384 contains 768 x 384
- C192 contains 384 x 192
- C96 contains 192 x 96

which implies a mapping of
```
  res=$(echo $CASE |cut -c2-5)
  LONB_CASE=$((2*res))
  LATB_CASE=$((1*res))
```

I provided the `gsibec` files to @CoryMartin-NOAA .   I think my mapping of the `gsibec` files into case directories is wrong.    It seems we should

- `mv C96 C48`
- `mv C192 C96`
- `mv C384 C192`
- `mv C768 C384`

in `GDASApp/fix/gsibec`.

This issue is opened to share this finding.  If the recommended change is correct and we move existing `gsibec` directories to new names, developers running `fv3jedi_var.x` with `gisbec` staticb will need to update their configurations.   The same comment applies to GDASApp ctests running `fv3jedi_var.x` with `gsibec` staticb

RussTreadon-NOAA:
Further study suggests that it may be better to retain the directory structure currently in `GDASApp/fix/gsibec`.    With the current directory structure 

- `CASE` specifies the background resolution
- `CASE_ANL` specifies the resolution on which to perform the analysis

[...]

PR #405, opened 2023-03-24
Add 3denvar option to run_jedi_exe.py

### LINKED ISSUE #402: Extend run_jedi_exe.py to launch 3denvar job
Currently `ush/run_jedi_exe.py` can launch fv3-jedi jobs for the following configurations:
- 3dhofx
- 4dhofx
- 3dvar
- letkf

A hybrid variational option is missing.  This issue is opened to extend `run_jedi_exe.py` to launch `fv3jedi_var.x` in 3denvar mode.

RussTreadon-NOAA:
The changes in this PR add the option to launch `fv3jedi_var.x` in 3denvar mode via `ush/run_jedi_exe.py`.   This functionality is useful since operational GSI-based global and regional DA systems run in a hybrid model in which the static background error is augmented with an ensemble component (envar).   

This PR adds four new files:
- `parm/atm/berror/hybvar_gsibec.yaml` - hybrid background error yaml file using `gsibec` for static-B along with a 10 member ensemble
- `test/atm/run_jedi_exe_3denvar.sh` - ctest using `run_jedi_exe.py` to launch `fv3jedi_var.x` in 3denvar mode 
- `ush/examples/run_jedi_exe/3denvar_hera.yaml` - configuration file used by `run_jedi_exe.py` to run 3denvar sample case on Hera
- `ush/examples/run_jedi_exe/3denvar_orion.yaml` - configuration file used by `run_jedi_exe.py` to run 3denvar sample case on Orion

Eleven existing files are modified:
- `CMakeLists.txt` - update gsibec to tag 1.0.7.   Previously pointed at head of gsibec develop.  This is problematic since the head of gsibec develop is not always function and/or requires specific hashes of other JEDI components.
- `parm/atm/obs/config/sondes.yaml` - replace `surface_geometric_height` with `surface_geopotential_height` to avoid `fv3jedi_var.x` segmentation fault
- `test/atm/CMakeLists.txt` - add `test_gdasapp_run_jedi_exe_3denvar`
- `test/genYAML_prep.sh` - replace `CASE_ENKF` with `CASE_ANL`
- `test/genYAML_prep_aero.sh` - replace `CASE_ENKF` with `CASE_ANL`
- `test/genYAML_prep_land.sh` - replace `CASE_ENKF` with `CASE_ANL`
- `ush/run_jedi_exe.py` - remove `atmos` from enkfgdas root path.   This change is necessitate by changes in the g-w GFS directory structure.   The `atmos` directory is now located in `memXXX` directories.
- `ush/ufsda/genYAML.py` - replace `CASE_ENKF` with `CASE_ANL`
- `ush/ufsda/misc_utils.py` - replace `CASE_ENKF` with `CASE_ANL`
- `ush/ufsda/stage.py` - add `atmos` to the `memXXX` path for ensemble members, enhance error checking for `dohybvar = .true.` option, add logic to ensure correct `gsibec` static-B is linked to the run directory.
- `ush/ufsda/ufs_yaml.py` - replace `CASE_ENKF` with `CASE_ANL`

Fixes #402

### DIFF
--- a/test/atm/run_jedi_exe_3denvar.sh
+++ b/test/atm/run_jedi_exe_3denvar.sh
@@ -1,0 +1,118 @@
+#!/bin/bash
+set -x
+
+bindir=$1
+srcdir=$2
+
+# Identify machine
+if [[ -d /scratch1 ]] ; then
+    machine="hera"
+elif [[ -d /work ]] ; then
+    machine="orion"
+else
+    echo "UNSUPPORTED MACHINE.  ABORT"
+    exit 99
+fi
+
+# Load modules
+set +x
+module use ${srcdir}/modulefiles
+module load GDAS/${machine}
+set -x
+module list
+
+# Set machine dependent variables
+if [ "$machine" = "hera" ] ; then
+    partition="hera"
+    gdasfix="/scratch1/NCEPDEV/da/Cory.R.Martin/GDASApp/fix"
+elif [ "$machine" = "orion" ]; then
+    partition="debug"
+    gdasfix="/work2/noaa/da/cmartin/GDASApp/fix"
+fi
+
+# Create test run directory
+mkdir -p ${bindir}/test/atm/global-workflow/testrun/gdas_single_test_3denvar
+cd ${bindir}/test/atm/global-workflow/testrun/gdas_single_test_3denvar
+
+# Create input yaml
+cat > ./3denvar_example.yaml << EOF
+working directory: ./
+GDASApp home: ${srcdir}
+GDASApp mode: variational
+template: ${srcdir}/parm/atm/variational/3dvar_dripcg.yaml
+config:
+  berror_yaml: ${srcdir}/parm/atm/berror/hybvar_gsibec.yaml
+  obs_dir: obs
+  diag_dir: diags
+  crtm_coeff_dir: crtm
+  bias_in_dir: obs
+  bias_out_dir: bc
+  obs_yaml_dir: ${srcdir}/parm/atm/obs/config
+  executable: ${bindir}/bin/fv3jedi_var.x
+  obs_list: ${srcdir}/parm/atm/obs/lists/gdas_prototype_3d.yaml
+  gdas_fix_root: ${gdasfix}
+  atm: true
+  layout_x: 1
+  layout_y: 1
+  atm_window_length: PT6H
+  valid_time: 2021-12-21T06:00:00Z
+  dump: gdas
+  case: C96
+  case_anl: C48
+  case_enkf: C48
+  staticb_type: gsibec
+  dohybvar: true
+  levs: 128
+  nmem: 10
+  comin_ges: /scratch1/NCEPDEV/da/Russ.Treadon/GDASApp/cases
+  interp_method: barycentric
+job options:
+  machine: ${machine}
+  account: da-cpu
+  queue: debug
+  partition: ${partition}
+  walltime: '30:00'
+  ntasks: 6
+  modulepath: ${srcdir}/modulefiles
+EOF
+
+# Execute run_jedi_exe.py
+if [ -e stdout.txt ]; then
+    rm -f stdout.txt
+fi
+${srcdir}/ush/run_jedi_exe.py -c ./3denvar_example.yaml > stdout.txt 2>&1
+rc=$?
+if [ $rc -ne 0 ]; then
+    exit $rc
+fi
+
+# Wait and ensure buffers flushed to disk
+sleep 5
+sync stdout.txt
+
+# Check for job submission error
+error=$(grep "sbatch: error" stdout.txt | wc -l)
+if [ $error -ne 0 ]; then
+    rc=$error
+    exit $rc
+fi
+
+# Cancel submitted job
+jobid=$(grep "Submitted" stdout.txt | awk -F' ' '{print $4}')
+scancel $jobid
+rc=$?
+if [ $rc -ne 0 ]; then
+    exit $rc
+fi
+
+# Check for valid yaml files
+ylist="3denvar_example.yaml gdas_variational.yaml"
+for yfile in $ylist; do
+    python3 -c 'import yaml, sys; yaml.safe_load(sys.stdin)' < $yfile
+    rc=$?
+    if [ $rc -ne 0 ]; then
+	exit $rc
+    fi
+done
+
+exit $rc
--- a/test/atm/run_jedi_exe_3denvar.sh
+++ b/test/atm/run_jedi_exe_3denvar.sh
@@ -25,11 +25,9 @@ module list
 if [ "$machine" = "hera" ] ; then
     partition="hera"
     gdasfix="/scratch1/NCEPDEV/da/Cory.R.Martin/GDASApp/fix"
-    cominges="/scratch1/NCEPDEV/da/Russ.Treadon/GDASApp/cases"
 elif [ "$machine" = "orion" ]; then
     partition="debug"
     gdasfix="/work2/noaa/da/cmartin/GDASApp/fix"
-    cominges="/work2/noaa/da/rtreadon/GDASApp/cases"
 fi
 
 # Create test run directory
@@ -66,7 +64,6 @@ config:
   dohybvar: true
   levs: 128
   nmem: 10
-  comin_ges: ${cominges}
   interp_method: barycentric
 job options:
   machine: ${machine}
--- a/test/atm/run_jedi_exe_3denvar.sh
+++ b/test/atm/run_jedi_exe_3denvar.sh
@@ -25,9 +25,11 @@ module list
 if [ "$machine" = "hera" ] ; then
     partition="hera"
     gdasfix="/scratch1/NCEPDEV/da/Cory.R.Martin/GDASApp/fix"
+    cominges="/scratch1/NCEPDEV/da/Russ.Treadon/GDASApp/cases"
 elif [ "$machine" = "orion" ]; then
     partition="debug"
     gdasfix="/work2/noaa/da/cmartin/GDASApp/fix"
+    cominges="/work2/noaa/da/rtreadon/GDASApp/cases"
 fi
 
 # Create test run directory
@@ -64,7 +66,7 @@ config:
   dohybvar: true
   levs: 128
   nmem: 10
-  comin_ges: /scratch1/NCEPDEV/da/Russ.Treadon/GDASApp/cases
+  comin_ges: ${cominges}
   interp_method: barycentric
 job options:
   machine: ${machine}
--- a/ush/run_jedi_exe.py
+++ b/ush/run_jedi_exe.py
@@ -75,8 +75,8 @@ def run_jedi_exe(yamlconfig):
     os.environ['assim_freq'] = str(assim_freq)
     oprefix = executable_subconfig['dump'] + ".t" + str(cyc) + "z."
     gprefix = executable_subconfig['dump'] + ".t" + str(gcyc) + "z."
-    comin = executable_subconfig.get('comin_ges', './')
-    comin_ges_ens = os.path.join(comin, 'enkfgdas.' + str(gPDY), str(gcyc))
+    comin = executable_subconfig.get('gdas_fix_root', './')
+    comin_ges_ens = os.path.join(comin, 'cases', 'enkfgdas.' + str(gPDY), str(gcyc))
 
     single_exec = True
     var_config = {
--- a/ush/run_jedi_exe.py
+++ b/ush/run_jedi_exe.py
@@ -77,7 +77,7 @@ def run_jedi_exe(yamlconfig):
     oprefix = executable_subconfig['dump'] + ".t" + str(cyc) + "z."
     gprefix = executable_subconfig['dump'] + ".t" + str(gcyc) + "z."
     comin = executable_subconfig.get('comin_ges', './')
-    comin_ges_ens = os.path.join(comin, 'enkfgdas.' + str(gPDY), str(gcyc), 'atmos')
+    comin_ges_ens = os.path.join(comin, 'enkfgdas.' + str(gPDY), str(gcyc))
 
     single_exec = True
     var_config = {
--- a/ush/run_jedi_exe.py
+++ b/ush/run_jedi_exe.py
@@ -8,8 +8,6 @@ import subprocess
 import sys
 import yaml
 
-from ufsda.genYAML import genYAML
-
 
 def export_envar(yamlfile, bashout):
 
@@ -59,6 +57,7 @@ def run_jedi_exe(yamlconfig):
     import ufsda
     from ufsda.misc_utils import calc_fcst_steps
     from ufsda.stage import gdas_single_cycle, gdas_fix, background_ens, atm_obs, bias_obs
+    from ufsda.genYAML import genYAML
 
     # compute config for YAML for executable
     executable_subconfig = all_config_dict['config']
--- a/ush/ufsda/stage.py
+++ b/ush/ufsda/stage.py
@@ -1,7 +1,6 @@
 from r2d2 import fetch
 from solo.basic_files import mkdir
 from solo.date import Hour, DateIncrement, date_sequence
-from solo.logger import Logger
 from solo.stage import Stage
 from solo.configuration import Configuration
 from solo.nice_dict import NiceDict
@@ -308,29 +307,7 @@ def gdas_single_cycle(config):
             ufsda.r2d2.fetch(r2d2_config)
     # if hybvar, link ensemble members to DATA
     if config['DOHYBVAR']:
-        for imem in range(1, config['NMEM_ENKF']+1):
-            memchar = f"mem{imem:03d}"
-            logging.info("Stage hybvar ens {memchar}")
-            rst_dir = os.path.join(config['COMIN_GES_ENS'], memchar, 'atmos', 'RESTART')
-            ges_dir = os.path.join(config['COMIN_GES_ENS'], memchar, 'atmos', 'RESTART_GES')
-            jedi_bkg_mem = os.path.join(config['DATA'], 'ens')
-            jedi_bkg_dir = os.path.join(config['DATA'], 'ens', memchar)
-
-            mkdir(jedi_bkg_mem)
-
-            # copy RESTART to RESTART_GES
-            if not os.path.exists(ges_dir):
-                try:
-                    shutil.copytree(rst_dir, ges_dir)
-                except FileExistsError:
-                    shutil.rmtree(ges_dir)
-                    shutil.copytree(rst_dir, ges_dir)
-            try:
-                os.symlink(ges_dir, jedi_bkg_dir)
-            except FileExistsError:
-                os.remove(jedi_bkg_dir)
-                os.symlink(ges_dir, jedi_bkg_dir)
-
+        background_ens(config)
 
 def background(config):
     """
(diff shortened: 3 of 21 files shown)


ush/ufsda/stage.py
```diff
@@ -33,21 +33,35 @@ def gdas_fix(input_fix_dir, working_dir, config):
...
+    if staticb_source in ['bump']:
         ufsda.disk_utils.symlink(os.path.join(input_fix_dir, staticb_source, case_anl),
                                  config['fv3jedi_staticb_dir'])
 
+    # link gsibec staticb
+    if staticb_source in ['gsibec']:
+        if dohybvar:
+            case_gsibec = case_anl
+        else:
+            case_gsibec = case
```
kevindougherty-noaa:
```
case_gsibec = case_anl if dohybvar else case
```

ush/ufsda/stage.py
```diff
@@ -303,6 +317,30 @@ def gdas_single_cycle(config):
             target_file = target_file.replace('nc4', 'txt')
             r2d2_config['target_file_fmt'] = target_file
             ufsda.r2d2.fetch(r2d2_config)
+    # if hybvar, link ensemble members to DATA
+    if config['DOHYBVAR']:
+        for imem in range(1, config['NMEM_ENKF']+1):
+            memchar = f"mem{imem:03d}"
+            print(f"background_ens:  stage {memchar}")
```
kevindougherty-noaa: Do you want print statements or should this be outputted to a logger file?

kevindougherty-noaa (commented):
Just a couple of quick remarks. I will take a more in-depth look first thing Monday.

ush/ufsda/stage.py
RussTreadon-NOAA: Thanks.  I didn't realize the if-else structure could be collapsed into a single line.   The suggested line is succinct and readable.   

I made the change.   Subsequent testing revealed that the original logic is not correct.  `gsibec` requires that we use `case` independent of the value of `dohybvar`.   Neither the original logic nor suggested revision are necessary.   

The logical test was removed and `case` is used for all `gsibec` runs.  Change committed at 0ba33d8.

ush/ufsda/stage.py
RussTreadon-NOAA: Good suggestion.  `print` was replaced with `logging.info` at 0ba33d8.

Much to my surprise, no output is generated by `logging.info` when `run_jedi_exe.py`   I added a `print` statement before `logging.info` and only the `print` was echoed to stdout.

A check of `ush/run_jedi_exe.py` shows other instances of `logging.info` which do not generate printout when the script is executed.   This is odd because `run_jedi_exe.py` sets the logging level to `INFO`
```
def run_jedi_exe(yamlconfig):
    logging.basicConfig(format='%(asctime)s:%(levelname)s:%(message)s', level=logging.INFO, datefmt='%Y-%m-%d %H:%M:%S')
    # open YAML file to get config
    try:
        with open(yamlconfig, 'r') as yamlconfig_opened:
            all_config_dict = yaml.safe_load(yamlconfig_opened)
        logging.info(f'Loading configuration from {yamlconfig}')
```
When `run_jedi_exe.py` is executed, the `Loading configuration from ...` message is _not_ printed.   The only printout I see when executing `run_jedi_exe.py` is from R2D2.   I do not understand why we see this behavior.   Any ideas @CoryMartin-NOAA?

RussTreadon-NOAA: As a test, I commented out 
```
from ufsda.genYAML import genYAML
```
at the top of `run_jedi_exe.py`.    With this line commented out, `run_jedi_exe.py` generated info log messages up to the point where the script referenced `genYAML`.   

```
(gdasapp) Hera(hfe05):/scratch1/NCEPDEV/da/Russ.Treadon/git/global-workflow/test2/sorc/gdas.cd/ush$ ./run_jedi_exe.py -c examples/run_jedi_exe/3denvar_hera.yaml
2023-03-25 20:50:42:INFO:Loading configuration from examples/run_jedi_exe/3denvar_hera.yaml
2023-03-25 20:50:42:INFO:Created working directory /scratch2/NCEPDEV/stmp1/Russ.Treadon/gdas_single_test_3denvar.case_anl                              _c48
2023-03-25 20:50:44:INFO:Note: detected 80 virtual cores but NumExpr set to maximum of 64, check "NUMEXPR_MAX_THREADS" en                              vironment variable.
2023-03-25 20:50:44:INFO:Note: NumExpr detected 80 cores but "NUMEXPR_MAX_THREADS" not set, so enforcing safe limit of 8.
2023-03-25 20:50:57:INFO:Using yamlconfig examples/run_jedi_exe/3denvar_hera.yaml
Traceback (most recent call last):
  File "./run_jedi_exe.py", line 177, in <module>
    run_jedi_exe(args.config)
  File "./run_jedi_exe.py", line 139, in run_jedi_exe
    genYAML(yamlconfig, output=output_file)
NameError: name 'genYAML' is not defined
```

Why does importing `genYAML` from `ufsda.genYAML` turn off info level logging?

RussTreadon-NOAA: The following `import` section is inside `def run_jedi_exe(yamlconfig):` in `run_jedi_exe.py`
```
    # add your ufsda python package to path
    ufsda_path = os.path.join(all_config_dict['GDASApp home'], 'ush')
    sys.path.append(ufsda_path)
    import ufsda
    from ufsda.misc_utils import calc_fcst_steps
    from ufsda.stage import gdas_single_cycle, gdas_fix, background_ens, atm_obs, bias_obs
```
I removed `from ufsda.genYAML import genYAM` from the top of `run_jedi_exe.py` and added it to the above `import` section.    With this change, `run_jedi_exe.py` runs to completion with info messages written to stdout.    This change was committed at [`f57fae0`](https://github.com/RussTreadon-NOAA/GDASApp/commit/f57fae0199ccdf2ff863623cb52ff9cb97f99210)

I don't understand why this change yields the desired behavior.

CoryMartin-NOAA: Is it possible it was using a different version of `genYAML` when it was imported before the `sys.path.append` section? I wonder if that has anything to do with it. Otherwise this is very strange indeed.

RussTreadon-NOAA: Various tests this morning adding `print` to see what's loaded or moving `import` around did not uncover any explanation for the observed behavior.   `run_jedi_exe.py` at [`f57fae0`](https://github.com/RussTreadon-NOAA/GDASApp/commit/f57fae0199ccdf2ff863623cb52ff9cb97f99210) runs with log statements correctly written to stdout.

RussTreadon-NOAA:
Rebuilt GDASApp 3/27 evening.  Much to my surprise 3dvar and 3denvar jobs launched by `run_jedi_exe.py` failed.    

The GDASApp build clones the head of gsibec develop.   
```
 # Build GSI-B
   option(BUILD_GSIBEC "Build GSI-B" OFF)
   if(BUILD_GSIBEC)
    ecbuild_bundle( PROJECT gsibec GIT "https://github.com/GEOS-ESM/GSIbec.git" BRANCH develop )
   endif()
```

I reached out to Ricardo Todling (GMAO).  He advised that we use the gsibec hash referenced in saber.   This is gsibec tag 1.0.7.   Update GDASApp top level `CMakeLists.txt` to clone
```
 # Build GSI-B
   option(BUILD_GSIBEC "Build GSI-B" OFF)
   if(BUILD_GSIBEC)
    ecbuild_bundle( PROJECT gsibec GIT "https://github.com/GEOS-ESM/GSIbec.git" TAG 1.0.7 )
   endif()
```
Rebuilt GDASApp.   The 3dvar and 3denvar tests successfully ran to completion with gsibec tag 1.0.7.   Commit this change `feature/3denvar` at 018cf8b.

RussTreadon-NOAA:
**Testing**

- Automated GW-RT tests of `feature/3denvar` at [741fd8f](https://github.com/NOAA-EMC/GDASApp/pull/405/commits/741fd8fded8fc28532a8bfc5b9e2882d340d828a) pass on Hera and Orion.   
- Script `run_jedi_exe.py` successfully launches and runs 3dvar, 3denvar, 3dhofx, 4dhofx, and letkf on Hera and Orion.

The changes in this PR are ready for review.

CoryMartin-NOAA:
@RussTreadon-NOAA do we wish to review/merge this PR before we have fixed the genYAML issues? I'm fine with that approach but wanted to check with you before I proceed with reviewing this.

RussTreadon-NOAA:
@CoryMartin-NOAA , if I had a quick fix for the issue documented in issue #408, I'd add it to this PR. 
 I don't.   Given this, I'm inclined to move forward with this PR.  Resolution of the problem noted in issue #408 can be the subject of a new PR.

CoryMartin-NOAA:
@RussTreadon-NOAA works for me, more agile!

test/atm/run_jedi_exe_3denvar.sh
```diff
@@ -0,0 +1,120 @@
...
+module use ${srcdir}/modulefiles
+module load GDAS/${machine}
+set -x
+module list
+
+# Set machine dependent variables
+if [ "$machine" = "hera" ] ; then
+    partition="hera"
+    gdasfix="/scratch1/NCEPDEV/da/Cory.R.Martin/GDASApp/fix"
+    cominges="/scratch1/NCEPDEV/da/Russ.Treadon/GDASApp/cases"
```
CoryMartin-NOAA: @RussTreadon-NOAA do you want me to put the cases in the fix directory or would you like to be the case maintainer? I'm fine either way.

RussTreadon-NOAA: My preference is to keep things together in `gdasfix`.   Long term `gdasfix` could be in a role account directory so those in the group could share responsibility for maintaining `gdasfix`.

CoryMartin-NOAA: agreed, I'll copy over after lunch.

CoryMartin-NOAA: Now:
`cominges=${gdasfix}/cases`

RussTreadon-NOAA: Thank you @CoryMartin-NOAA .   Given this change the following have been updated to no longer reference `cominges` or `comin_ges`:

- `test/atm/run_jedi_exe_3denvar.sh`
- `test/atm/run_jedi_exe_letkf.sh`
- `ush/examples/run_jedi_exe/3denvar_hera.yaml`
- `ush/examples/run_jedi_exe/3denvar_orion.yaml`
- `ush/examples/run_jedi_exe/letkf_hera.yaml`
- `ush/examples/run_jedi_exe/letkf_orion.yaml`
- `ush/run_jedi_exe.py`

These changes were committed to `feature/3denvar` at [00a28d3](https://github.com/RussTreadon-NOAA/GDASApp/commit/00a28d3cc5cb1adaae267b1cd2a0b24bd1251b0c).

ush/ufsda/stage.py
```diff
@@ -33,21 +33,31 @@ def gdas_fix(input_fix_dir, working_dir, config):
...
+    # set layers
     layers = int(config['LEVS'])-1
 
     # figure out staticb source
     staticb_source = config.get('STATICB_TYPE', 'gsibec')
 
-    # link staticb
-    if staticb_source in ['bump', 'gsibec']:
+    # link bump staticb
+    if staticb_source in ['bump']:
```
CoryMartin-NOAA: @RussTreadon-NOAA I think I remember seeing some discussion about the gsibec fix file directory structure? 

what about going back to `if staticb_source in ['bump', 'gsibec']:` but making sure `case_anl` is consistent regardless of BUMP/GSIBEC? I would think we should be able to have independent `case` and `case_anl` if possible.

RussTreadon-NOAA: Hmm, I'll need to think about this.   The current approach gets back to closed issue #403.

The way gsibec is implemented in UFS-DA we need the gsibec `case` to equal the background `case`.   The `case` gsibec may be used to generate a variational analysis on the `case_anl` grid.    For example, `fv3jedi_var.x` can read in `case=C96` tiles and then perform a gsibec 3dvar analysis on the `case_anl=C96` or `case_anl=C48` grid.   

The gsibec logic in `stage.py` is designed for this flexibility.   gsibec is associated with `case` while the analysis resolution is specified by `case_anl`.

We could refactor this section of `stage.py` so single `ufsda.disk_utils.symlink` line handles both bump and gsibec.   Would the following work & be preferable?

```
    # link staticb
    case_berror = case if staticb_source in ['gsibec'] else case_anl
    ufsda.disk_utils.symlink(os.path.join(input_fix_dir, staticb_source, case_berror),
                               config['fv3jedi_staticb_dir'])
```

CoryMartin-NOAA: yeah I think this is exactly what I had in mind, this approach reduces the amount of redundant code in the "if-block"

RussTreadon-NOAA: Got it.  The suggested change was made and committed at 581bdd8.

ush/ufsda/stage.py
```diff
@@ -303,6 +306,30 @@ def gdas_single_cycle(config):
...
+            logging.info("Stage hybvar ens {memchar}")
+            rst_dir = os.path.join(config['COMIN_GES_ENS'], memchar, 'atmos', 'RESTART')
+            ges_dir = os.path.join(config['COMIN_GES_ENS'], memchar, 'atmos', 'RESTART_GES')
+            jedi_bkg_mem = os.path.join(config['DATA'], 'ens')
+            jedi_bkg_dir = os.path.join(config['DATA'], 'ens', memchar)
+
+            mkdir(jedi_bkg_mem)
+
+            # copy RESTART to RESTART_GES
+            if not os.path.exists(ges_dir):
```
CoryMartin-NOAA: Last comment before approval (sorry!)

So this section checks to see if the path doesn't exist, and then the try/except also does the `FileExistsError`. Is this just to be extra cautious in case it is created in this split second?
Or is this not doing what you want it to do? What about if the directory exists but it is not 'correct'? In this case it would not even get to the try/except and copy section.

RussTreadon-NOAA: @CoryMartin-NOAA , I confess to not putting much thought into this block.   I simply copied it from `background` in `stage.py` and placed it in a loop over the members.  I can't recall why I added the test `if not os.path.exists(ges_dir)`.   

This is the good thing about reviews.  They make you go back and think.  

Looking more closely I see that `background_ens` does exactly the same thing as the section of script above with the exception that `background_ens` creates member analysis directories.  This is necessary for enkf runs.   It's not necessary for EnVar runs.   

Instead of maintaining two blocks of script which basically do the same, I replaced the section of scripting discussed in this comment with `background_ens(config)`.  Logic has been added to `background_ens` to skip creation of member analysis directories when running EnVar.  

I did not commit this change yet because I'm not sure why we have `RESTART` and `RESTART_GES`   I don't see a need for `RESTART_GES`.   Do you recall why we have `RESTART_GES`?

CoryMartin-NOAA: The only thing that I can think of is if the analysis (RESTART) files are updated directly rather than passing increments. So then you keep the 'old' RESTART files as the guess and the new ones get updated and then used for the next forecast. We do this in the aerosol DA work right now because the model can't read the aerosol increments, but we can use the existing infrastructure for Gaussian increments after regridding for the ATM app, so no, I don't really see a point in having a `RESTART_GES` either.

RussTreadon-NOAA: Using `background_ens` for EnVar and LETKF proved to be more involved than I anticipated.

`/parm/atm/lgetkf/lgetkf.yaml` includes `/RESTART` in the local path to each ensemble member.   `/parm/atm/berror/hybvar_gsibec.yaml` and  `parm/atm/berror/hybvar_bump.yaml` do not include `/RESTART` in the path.  I opted to remove `/RESTART` from `lgetkf.yaml`.   

There's another difference in the local path to the ensemble backgrounds for EnVar and LETKF.   The EnVar path use `ens`.   The LETKF path uses  `bkg`.   Logic was added to `background_ens` in `stage.py` to handle this.  

These changes were committed at 3904f75.

I currently have
```
    dohybvar = config['DOHYBVAR']
    bkgdir = 'ens'
    if not dohybvar: bkgdir = 'bkg'
```
in `background_ens`.   Is there a clever way to compress this to fewer lines?   My google searches have not yet turned up anything.

CoryMartin-NOAA: No I think that's about as compressed as it can get, unless you only use `dohybvar` once, then you can replace it with `config['DOHYBVAR']` on the last line and remove the first?

RussTreadon-NOAA: Good suggestion.  Decide to directly use `config['DOHYBVAR']`.  Also clean up comments.  These changes committed at 854ed4b.

Manually reran 3dvar, 3denvar, letkf, 3dhofx, 4dhofx via `run_jedi_exe.py` on Orion.  All successfully run to completion.  

Manually ran 49 ctests within g-w `develop`.   All tests pass except one.  `test_gdasapp_atm_jjob_ens_run` failed.   Need to update g-w `scripts/exgdas_global_atmos_ensanal_run.sh` to account for removal of `RESTART` in ensemble member background path.   This g-w change will be addressed in g-w PR [#1373](https://github.com/NOAA-EMC/global-workflow/pull/1373).

CoryMartin-NOAA: Ok so since we have a chicken/egg problem and a queue of PRs in g-w, do we want to wait for that or merge this one and accept a failing test for the short term?

RussTreadon-NOAA: Ideally we update g-w `develop` at the same time we merge this PR into GDASApp `develop`.   This won't happen.  

My preference is to see this PR enter GDASApp `develop` sooner rather than later.  I'm not keen on maintaining `feature/3denvar` until g-w is ready.  I've been maintaining several g-w branches for a few weeks waiting for PRs [#1370](https://github.com/NOAA-EMC/global-workflow/pull/1370), [#1372](https://github.com/NOAA-EMC/global-workflow/pull/1372), and [#1373](https://github.com/NOAA-EMC/global-workflow/pull/1373) to be reviewed and merged into g-w `develop`.

CoryMartin-NOAA: Works for me, will do another review right now

ush/run_jedi_exe.py
```diff
@@ -76,8 +75,8 @@ def run_jedi_exe(yamlconfig):
     os.environ['assim_freq'] = str(assim_freq)
     oprefix = executable_subconfig['dump'] + ".t" + str(cyc) + "z."
     gprefix = executable_subconfig['dump'] + ".t" + str(gcyc) + "z."
-    comin = executable_subconfig.get('comin_ges', './')
-    comin_ges_ens = os.path.join(comin, 'enkfgdas.' + str(gPDY), str(gcyc), 'atmos')
+    comin = executable_subconfig.get('gdas_fix_root', './')
+    comin_ges_ens = os.path.join(comin, 'cases', 'enkfgdas.' + str(gPDY), str(gcyc))
```
CoryMartin-NOAA, line 79: just double checking you don't need `'atmos'` here any more in the path?

CoryMartin-NOAA: Never mind, I see below how it is used with the members

CoryMartin-NOAA (approved):

CoryMartin-NOAA:
Thanks @RussTreadon-NOAA for all the hard work on this and the back and forth in review, much appreciated!

CoryMartin-NOAA:
Will wait either for a 2nd approval or confirmation about the RT results being as expected before merging

CoryMartin-NOAA:
These two single test failures are expected given earlier discussion with @RussTreadon-NOAA , I will merge this PR in and hope that the g-w PR can be merged relatively soon.

RussTreadon-NOAA:
The Hera and Orion GW-RT failures of `test_gdasapp_atm_jjob_ens_run ` are expected.   

This PR removes `RESTART` from the run directory paths to the ensemble member backgrounds for ensemble DA.  This was done to be consistent with the run directory path to the deterministic backgrounds for variational DA.  The variational DA path does not include `RESTART`.  The run directory paths to the ensemble members for ensemble variational DA also do not include `RESTART`.

g-w `scripts/exgdas_global_atmos_ensanal_run.sh` needs to be updated to account for this change.  The required change will be made in g-w PR [#1373](https://github.com/NOAA-EMC/global-workflow/pull/1373).  Ideally g-w PR [#1373](https://github.com/NOAA-EMC/global-workflow/pull/1373) and GDASApp PR #405 would be merged into their respective `develop` branches around the same time.

CoryMartin-NOAA:
Sorry @RussTreadon-NOAA if I jumped the gun on merging this. Let me try to review https://github.com/NOAA-EMC/global-workflow/pull/1373 later today and see if we can get the ball rolling on that.

RussTreadon-NOAA:
Thank you @CoryMartin-NOAA  for merging this PR into GDASApp `develop`.   

It's OK to pause your review of g-w PR [#1373](https://github.com/NOAA-EMC/global-workflow/pull/1373).   g-w [#1370](https://github.com/NOAA-EMC/global-workflow/pull/1370) and [#1372](https://github.com/NOAA-EMC/global-workflow/pull/1372) are ahead of [#1373](https://github.com/NOAA-EMC/global-workflow/pull/1373).  EIB is currently reviewing PR [#1370](https://github.com/NOAA-EMC/global-workflow/pull/1370).

Now that GDASApp `develop` has been updated, I'll formally update g-w PR [#1373](https://github.com/NOAA-EMC/global-workflow/pull/1373).

PR #411, opened 2023-04-06
change CASE_ENKF to CASE_ANL in UFO testing script

CoryMartin-NOAA:
#405 changed the scripts that compute the analysis resolution for the YAML files from `CASE_ENKF` to `CASE_ANL`.

The UFO driver testing scripts (such as what produce https://emc.ncep.noaa.gov/data_assimilation/JEDI/GDAS/UFO/acceptance/) need to also be updated to define these variables in its environment, otherwise it will fail.

This PR makes that change for ush/ufoeval/run_ufo_hofx_test.sh

### DIFF
--- a/ush/ufoeval/run_ufo_hofx_test.sh
+++ b/ush/ufoeval/run_ufo_hofx_test.sh
@@ -121,7 +121,7 @@ export cyc=${CDATE:8:2}
 export gPDY=${GDATE:0:8}
 export gcyc=${GDATE:8:2}
 export CASE="C768"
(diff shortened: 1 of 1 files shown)


CoryMartin-NOAA:
I should note that the UFO tests don't actually need this variable (there is no analysis resolution) but the YAML generation code requires it (that should probably be also fixed later)

RussTreadon-NOAA (approved):
Sorry for not catching this in my PR from earlier in the week.   Change is correct.  

Approve.

Issue #376, opened 2023-03-06
Add start of infrastructure for BUFR to IODA observation processing to GDASApp


CoryMartin-NOAA:
Through a lot of effort by Ron and others, there is a bufr2ioda executable in ioda-converters that can convert BUFR files to netCDF4/HDF5 IODA files based on input YAML configurations.

While this will be an iterative process to get every field needed (and with the correct format, names, units, etc.), this issue is to track the initial inclusion of this infrastructure into GDASApp.

- [ ] Templated YAML for PrepBUFR
- [x] Templated YAML for ATMS
- [ ] Templated YAML for AMSU-A
- [x] Script to generate YAMLs
- [x] Script to run executable

guillaumevernieres:
FYI @CoryMartin-NOAA : https://github.com/NOAA-EMC/GDASApp/blob/fede25cb41932e339671258d0a0af7913320abf2/test/soca/CMakeLists.txt#L14

We could just move the bufr2ioda tests out of "soca", the testing "infrastructure" is already there and was added by @AndrewEichmann-NOAA and @ShastriPaturi a while back.

CoryMartin-NOAA:
@guillaumevernieres yeah I have testing too for one ob type to make sure it works, this issue is to include YAMLs and have infrastructure to process a full cycle's worth of obs from BUFR to IODA in one go, do you have anything to do the latter yet?

guillaumevernieres:
> @guillaumevernieres yeah I have testing too for one ob type to make sure it works, this issue is to include YAMLs and have infrastructure to process a full cycle's worth of obs from BUFR to IODA in one go, do you have anything to do the latter yet?

I'm not quite sure what the above means @CoryMartin-NOAA ... So probably not! What's in the soca ctest only runs the bufr to ioda converter for a few of the marine obs, so just the same thing that you currently are doing.

PR #442, opened 2023-04-21
First steps towards BUFR to IODA automatic processing

CoryMartin-NOAA:
This PR partially addresses Issue #376 

This PR adds the first template BUFR2IODA YAML file for ATMS satellite radiance observations. 
It also adds a driver script that can loop through all BUFR files in a provided directory, look to see if there is a corresponding YAML template, and if so, will fill out the template and run `bufr2ioda.x` on that observation type. This is primarily a shell script but also calls a python script to handle the YAML generation.

An example of how to run this driver script:
- First load the GDASApp modules (and this assumes you already have built GDASApp)
- `./run_bufr2ioda.sh 2021080100 gdas /scratch1/NCEPDEV/global/glopara/dump/gdas.20210801/00/atmos/ /scratch2/NCEPDEV/stmp1/Cory.R.Martin/tmp/gdasapp/parm/ioda/ /scratch2/NCEPDEV/stmp1/Cory.R.Martin/gdas.20210801/00/obs`


CoryMartin-NOAA:
Tagging @danholdaway for awareness as Swell may want to implement something similar

guillaumevernieres (approved):
code looks good but I don't know anything about the bufr to ioda converter.

emilyhcliu:
@CoryMartin-NOAA I did not rcheck out the branch and run the scripts.  But the procedures and the AMTS YAML look right to me.

emilyhcliu (approved):

.gitignore
```diff
@@ -159,7 +159,7 @@ fv3-jedi-data/
 fv3-jedi-lm/
 gsibec/
 gsw/
-ioda/
+/ioda/
```
RussTreadon-NOAA, line 162: Curiosity.   Why add a leading `/`?   What does this do?

parm/ioda/bufr_atms.yaml
```diff
@@ -0,0 +1,176 @@
...
+            query: "*/SAID"
+
+          satelliteInstrument:
+            query: "*/SIID"
+
+          # The type should be integer.
+          # However, UFO/CRTM requred this variable to be float
+          fieldOfViewNumber:
+            query: "*/FOVN"
+            type: float
```
RussTreadon-NOAA, line 34: Curious.  When will ufo and/or crtm be updated to allow FOVN to be an integer as it should be?

parm/ioda/bufr_atms.yaml
```diff
@@ -0,0 +1,176 @@
...
+                hour: "*/HOUR"
+                minute: "*/MINU"
+                second: "*/SECO"
+
+        splits:
+          satId:
+            category:
+              variable: satelliteIdentifier
+              map:
+                _224: npp
```
RussTreadon-NOAA, line 81: We identify different `satelliteIdentifiers` in the yaml but these different `satelliteIdentifiers` share many parameters in common.   What happens if a future atms instrument has slightly different parameters (e.g., `scanStart` or `scanStep`)?

RussTreadon-NOAA (commented):
While not a requirement for this PR, should we, at some point, add a bufr2ioda ctest to GDASApp?

I did not exercise `run_bufr2ioda.sh`.   I assume it along with the python script and yaml work as intended.

.gitignore
CoryMartin-NOAA: if you don't do that, it will also ignore subdirectories named `ioda/` (like `parm/ioda`), this tells it to just ignore the root `/ioda/` directory.

parm/ioda/bufr_atms.yaml
CoryMartin-NOAA: good question, would defer to @emilyhcliu

parm/ioda/bufr_atms.yaml
CoryMartin-NOAA: @emilyhcliu any thoughts on this?

.gitignore
RussTreadon-NOAA: Thanks!  I learned something new.

ShastriPaturi (commented):
@CoryMartin-NOAA can you share a ATMS bufr file for me to test?

CoryMartin-NOAA:
@ShastriPaturi `/scratch1/NCEPDEV/global/glopara/dump/gdas.20210801/00/atmos/gdas.t00z.atms.tm00.bufr_d`

ShastriPaturi (approved):
Works like a charm.
This needs to be adapted for the marine BUFR files.
Thanks @CoryMartin-NOAA

PR #446, opened 2023-04-26
update aero bump yaml based on updated fv3jedi and global-workflow

ypwang19:
This PR updates the template staticb_bump.yaml in parm/aero/berror for aerosol DA based on the updated BUMP in fv3-jedi and the updated global-workflow.

- replace key word "**bump**" to "**read**".
- change **datetime: 2016-06-30T00:00:00Z** to **datetime: '2016-06-30T00:00:00Z'**, because JSON in g-w can not work with type datetime.

### DIFF
--- a/parm/aero/berror/staticb_bump.yaml
+++ b/parm/aero/berror/staticb_bump.yaml
@@ -1,7 +1,7 @@
 covariance model: SABER
 saber central block:
   saber block name: BUMP_NICAS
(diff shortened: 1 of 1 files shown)


CoryMartin-NOAA (approved):
looks good, thanks!

andytangborn (approved):
Looks good to me too.

PR #447, opened 2023-04-26
python post script

AndrewEichmann-NOAA:
Moves functionality of jobs/JGDAS_GLOBAL_OCEAN_ANALYSIS_POST in g-w to python script in GDASApp.  Will require change to jjob script to be used, but should be safe to have in develop branch.

Adds one file:
scripts/exgdas_global_marine_analysis_post.py

Tested with soca global-workflow ctests, test_gdasapp_soca_JGDAS_GLOBAL_OCEAN_ANALYSIS_POST and subsequent test_gdasapp_soca_JGDAS_GLOBAL_OCEAN_ANALYSIS_VRFY both pass.


scripts/exgdas_global_marine_analysis_post.py
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env python3
+################################################################################
+#  UNIX Script Documentation Block
+#                      .                                             .
+# Script name:         exufsda_global_marine_analysis_post.py
+# Script description:  Stages files and generates YAML for UFS Global Marine Analysis
+#
+# Author: Andrew Eichmann    Org: NCEP/EMC     Date: 2023-04-24
+#
+# Abstract: This script stages the marine observations, backgrounds and prepares
```
guillaumevernieres: unfortunate copy/paste for the abst here

guillaumevernieres (commented):
I thought we said a shell script @AndrewEichmann-NOAA ? I'm fine with python, but that seems like more work.

scripts/exgdas_global_marine_analysis_post.py
AndrewEichmann-NOAA: corrected

scripts/exgdas_global_marine_analysis_post.py
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env python3
+################################################################################
+#  UNIX Script Documentation Block
+#                      .                                             .
+# Script name:         exufsda_global_marine_analysis_post.py
+# Script description:  Stages files and generates YAML for UFS Global Marine Analysis
```
guillaumevernieres: Dang ... that needs to be updated too.

AndrewEichmann-NOAA: lil' ol' genius me

guillaumevernieres (approved):
Thanks @AndrewEichmann-NOAA .

PR #461, opened 2023-05-05
add ensemble based bump localization to gsibec hybvar yaml

RussTreadon-NOAA:
This PR is opened to add bump generated localization to `hybvar_gsibec.yaml`.    A user provided ensemble is processed by bump to generate localization values.

Please note the localization added to the ensemble section of `hybvar_gsibec.yaml` is for demonstration purposes only.   Subsequent testing and tuning is required for scientific validation.

Fixes #460

PR #463, opened 2023-05-08
Remove the WORKDIR prior to populating the directory

jiaruidong2017:
The `test_imsproc.sh` does not remove or empty the `WORKDIR` prior to populating the directory. This PR removed the `WORKDIR` so that the `test_gdasapp_land_imsproc` created a clean `WORKDIR` each time when the script runs. (https://github.com/NOAA-EMC/GDASApp/pull/461#issuecomment-1538348274)


RussTreadon-NOAA (approved):
Change tested in a working copy of `test_imsproc.sh` in which `$WORKDIR` existed.  `ctest -R test_gdasapp_land_imsproc` executed and successfully ran to completion.

Approved.

CoryMartin-NOAA (approved):

Issue #477, opened 2023-05-18
gdasocnanalvrfy fails because of directory variable assigned to nothing


AndrewEichmann-NOAA:
gdasocnanalvrfy fails in cycling with the following:
```
Traceback (most recent call last):
  File "/scratch1/NCEPDEV/da/Andrew.Eichmann/fv3gfs/global-workflow/sorc/gdas.cd/scripts/exgdas_global_marine_analysis_vrfy.py", line 212, in <module>
    os.path.join(com_ice_history, f'{RUN}.t{gcyc}z.icef006.nc')]
  File "/scratch1/NCEPDEV/da/python/opt/core/miniconda3/4.6.14/envs/eva/lib/python3.9/posixpath.py", line 76, in join
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```
Investigation shows that while previously `com_ice_history = os.getenv('COM_ICE_HISTORY_PREV') `, `COM_ICE_HISTORY_PREV` is not assigned in the corresponding jjob JGDAS_GLOBAL_OCEAN_ANALYSIS_VRFY, whereas `COM_ICE_HISTORY` is, as follows:
`RUN=${GDUMP} YMD=${gPDY} HH=${gcyc} generate_com -rx COM_ICE_HISTORY:COM_ICE_HISTORY_TMPL`
Furthermore it seems that `COM_ICE_HISTORY_PREV` is assigned to the same thing in `JGDAS_GLOBAL_OCEAN_ANALYSIS_VRFY`, which raises the question whether they both should be one or the other.

AndrewEichmann-NOAA:
Addressed with change to global-workflow

PR #478, opened 2023-05-18
fixing directory assignment bug

AndrewEichmann-NOAA:
Addresses https://github.com/NOAA-EMC/GDASApp/issues/477


guillaumevernieres:
Sorry for only noticing now @AndrewEichmann-NOAA , but the changes need to be in the `global-workflow` here:  
https://github.com/NOAA-EMC/global-workflow/blob/4bebafd61d5a0fc2f52910294204af5f1f51cb21/jobs/JGDAS_GLOBAL_OCEAN_ANALYSIS_VRFY#L19

I made a naming convention mistake a while back, I'm not sure why that one wasn't noticed. 
I forgot the `_PREV` for ice and ocean. It should have been something like this:
```bash
RUN=${GDUMP} YMD=${gPDY} HH=${gcyc} generate_com -rx COM_OCEAN_HISTORY_PREV:COM_OCEAN_HISTORY_TMPL
```

AndrewEichmann-NOAA:
@guillaumevernieres Ok, I'll reverse it. what's the difference between *_PREV and not *_PREV?

AndrewEichmann-NOAA:
@guillaumevernieres Ok, I'll make the change in g-w

AndrewEichmann-NOAA:
To be fixed with https://github.com/NOAA-EMC/global-workflow/pull/1631

PR #483, opened 2023-05-20
Add yaml files and namelist template for land DA in G-W.

jiaruidong2017:
This PR adds 2 files needed to create ensemble members for land DA using JEDI.

`incr.nml.j2` is a Fortran namelist jinja2 template for executable `JEDIINCEXE` (`apply_incr.exe`). 
`prep_run.yaml` is used as `PREP_RUN_LIST` in the global-workflow for copying and processing background data to the ensemble directories. 

Modified `prep_ims.yaml` and `letkf_create_ens.py` to accommodate the implementation in the global-workflow. 

A branch in the global-workflow with the same name is used with this branch in GDASApp.


CoryMartin-NOAA (approved):
Looks fine as far as I can tell

aerorahul:
I have a branch `feature/land_analysis` that has the same updates that go with the [global-workflow PR #1635](https://github.com/NOAA-EMC/global-workflow/pull/1635).
I would appreciate if we clarify some questions before merging this and the global-workflow PRs

parm/land/prep/prep_ims.yaml
```diff
@@ -6,5 +6,5 @@ calcfims:
     - [$(COM_OBS)/$(OPREFIX)IMS4km_to_FV3_mapping.$(CASE)_oro_data.nc, $(DATA)/obs/IMS4km_to_FV3_mapping.$(CASE)_oro_data.nc]
 ims2ioda:
   copy:
-    - [$(DATA)/ims_snow_{{ current_cycle | to_YMDH }}.nc4, $(COM_OBS)/$(OPREFIX)ims_snow.nc4]
+    - [$(DATA)/ims_snow_{{ current_cycle | to_YMDH }}.nc4, $(COM_OBS)/$(OPREFIX)ims_snow_{{ current_cycle | to_YMDH }}.nc4]
```
aerorahul, line 9: Why do we need `YYYYMMDDHH` in the name of the file?

ush/land/letkf_create_ens.py
```diff
@@ -29,7 +29,7 @@
 for (mem, value) in zip(ens_dirs, sign):
     for tt in range(1, 7):
         # open file
-        out_netcdf = os.path.join(workdir, mem, f"{fstub}.sfc_data.tile{tt}.nc")
+        out_netcdf = os.path.join(workdir, mem, "RESTART", f"{fstub}.sfc_data.tile{tt}.nc")
```
aerorahul, line 32: I include this script in `land_analysis.py` in the global-workflow rather than this floating script.

parm/land/prep/prep_run.yaml
```diff
@@ -0,0 +1,31 @@
...
+    - [$(DATA)/anl/landinc.{{ current_cycle | to_YMD }}.{{ current_cycle.strftime('%H') }}0000.sfc_data.tile2.nc, $(DATA)/anl/{{ current_cycle | to_YMD }}.{{ current_cycle.strftime('%H') }}0000.xainc.sfc_data.tile2.nc]
+    - [$(DATA)/anl/landinc.{{ current_cycle | to_YMD }}.{{ current_cycle.strftime('%H') }}0000.sfc_data.tile3.nc, $(DATA)/anl/{{ current_cycle | to_YMD }}.{{ current_cycle.strftime('%H') }}0000.xainc.sfc_data.tile3.nc]
+    - [$(DATA)/anl/landinc.{{ current_cycle | to_YMD }}.{{ current_cycle.strftime('%H') }}0000.sfc_data.tile4.nc, $(DATA)/anl/{{ current_cycle | to_YMD }}.{{ current_cycle.strftime('%H') }}0000.xainc.sfc_data.tile4.nc]
+    - [$(DATA)/anl/landinc.{{ current_cycle | to_YMD }}.{{ current_cycle.strftime('%H') }}0000.sfc_data.tile5.nc, $(DATA)/anl/{{ current_cycle | to_YMD }}.{{ current_cycle.strftime('%H') }}0000.xainc.sfc_data.tile5.nc]
+    - [$(DATA)/anl/landinc.{{ current_cycle | to_YMD }}.{{ current_cycle.strftime('%H') }}0000.sfc_data.tile6.nc, $(DATA)/anl/{{ current_cycle | to_YMD }}.{{ current_cycle.strftime('%H') }}0000.xainc.sfc_data.tile6.nc]
```
aerorahul: Why can't JEDI write it in the format that is expected?
See changes to `letfkoi.yaml` in the branch `feature/land_analysis`.

parm/land/prep/incr.nml.j2
```diff
@@ -0,0 +1,9 @@
+&noahmp_snow
+ date_str = "{{ current_cycle | to_YMD }}",
+ hour_str = "{{ current_cycle.strftime('%H') }}",
+ res = {{ CASE[1:] }},
+ frac_grid = "{{ FRACGRID }}",
```
aerorahul: I would like to see a change in the program `apply_incr.exe`.
`FRACGRID` is expected to be a 3 letter character e.g. `YES`.
The code conditions on `YES` (all uppercase), all other exceptions are treated as `NO` or `.false`.
This means, `yes`, `Yes` are treated as `.false`.
Besides, `FRACGRID` can be inferred based on the contents of `otype` file being read, so ideally, the code can internally resolve if it is processing fractional grid or not.
Thoughts?

aerorahul (changes requested):
Please see the review comments in addition to the one before.

parm/land/prep/prep_run.yaml
CoryMartin-NOAA: This is a good question, I thought we discussed this previously @jiaruidong2017 to modify the JEDI YAML to write out the expected filenames

parm/land/prep/prep_ims.yaml
CoryMartin-NOAA: This is consistent with the other GDASApp 'applications' and is largely a holdover from our attempt at using R2D2 (I think). I suggest we keep this consistent here with the others and change them all in batch. See similarly: https://github.com/NOAA-EMC/GDASApp/blob/cb485bd1c0c5a13fedbda9ff67423b795b7a34d1/parm/atm/obs/config/amsua_n19.yaml#L6

jiaruidong2017: To match the standard observation file name format in the GDASApp.

ush/land/letkf_create_ens.py
jiaruidong2017: Do you mean you have put the script `letkf_create_ens.py` in `land_analysis.py`? The script `letkf_create_ens.py` comes from GDASApp, and this script is a key part in GDASApp for the snow DA tests.

aerorahul: Yes.  See PR #1635 in global-workflow

parm/land/prep/prep_run.yaml
jiaruidong2017: In letkfoi.yaml from GDASApp, the output increment is as below:
```
output increment:
  datapath: $(DATA)/anl
  prefix: landinc
  filetype: fms restart
  filename_sfcd: '{{ current_cycle | to_fv3time }}.sfc_data.nc'
  filename_cplr: '{{ current_cycle | to_fv3time }}.coupler.res'
  state variables: [snwdph,vtype,slmsk]
```

I tried to removed `prefix: landinc`, but it didn't work because it required to have the `prefix` definition and the prefix cannot be empty. I do need your help on this.

CoryMartin-NOAA: It doesn't work? I thought for sure `prefix` was optional...

aerorahul: In that case, I suggest set `prefix: landinc` and remove `xainc` from the destination filename.
This will be consistent w/ `aeroinc` in aerosol DA update.
An update may be required in `apply_incr.exe` to take this updated filename.
In general, minimize file movements and have a plan for how the data "flows" through the system.

parm/land/prep/incr.nml.j2
jiaruidong2017: This will require the help and actions from @barlage and @ClaraDraper-NOAA to make such changes.

aerorahul: I am happy to open a PR if they are willing to accept.  It is trivial to make this correction and update.

jiaruidong2017: @aerorahul Yes, that will be great. Thanks.

parm/land/prep/prep_run.yaml
jiaruidong2017: @CoryMartin-NOAA I will try it again, and let you know then.

parm/land/prep/incr.nml.j2
CoryMartin-NOAA: I think we should double check with Mike and Clara and then have @jiaruidong2017 take the lead on making the changes, as to not put too much burden on @aerorahul

ush/land/letkf_create_ens.py
jiaruidong2017: @CoryMartin-NOAA Do you think it is okay we can keep this letkf_create_ens.py both separately in the GDASApp and in the land_analysis.py from global-workflow?

CoryMartin-NOAA: We should not duplicate efforts. The GDASApp has several tests that use global-workflow scripts, we can make the GDASApp land test that uses this .py file do that instead (once it is in the workflow develop branch)

jiaruidong2017: @CoryMartin-NOAA I need your help to make the GDASApp land test by using the global-workflow script. Thanks.

CoryMartin-NOAA (changes requested):
Let's ensure that we can write out from JEDI the files (and read into the next step) consistently to minimize IO time spent copying/moving files.

parm/land/prep/prep_run.yaml
jiaruidong2017: I made an experiment by removing the `prefix:` from the letkfoi.yaml as below:

```
output increment:
  datapath: $(DATA)/anl
  filetype: fms restart
  filename_sfcd: '{{ current_cycle | to_fv3time }}.xainc.sfc_data.nc'
  filename_cplr: '{{ current_cycle | to_fv3time }}.xainc.coupler.res'
  state variables: [snwdph,vtype,slmsk]
```

The output from JEDI run is as below:
```
20210323.060000.20210323.060000.xainc.coupler.res
20210323.060000.20210323.060000.xainc.sfc_data.tile1.nc
20210323.060000.20210323.060000.xainc.sfc_data.tile2.nc
20210323.060000.20210323.060000.xainc.sfc_data.tile3.nc
20210323.060000.20210323.060000.xainc.sfc_data.tile4.nc
20210323.060000.20210323.060000.xainc.sfc_data.tile5.nc
20210323.060000.20210323.060000.xainc.sfc_data.tile6.nc
```
The output files have doubled times of `20210323.060000`. 

If I kept the option `prefix:` in the `letkfoi.yaml` as below: 
```
output increment:
  datapath: $(DATA)/anl
  prefix:
  filetype: fms restart
  filename_sfcd: '{{ current_cycle | to_fv3time }}.xainc.sfc_data.nc'
  filename_cplr: '{{ current_cycle | to_fv3time }}.xainc.coupler.res'
  state variables: [snwdph,vtype,slmsk]
```
The output is the same as above with double times. 

Do you have any suggestions? Thanks.

jiaruidong2017:
> Let's ensure that we can write out from JEDI the files (and read into the next step) consistently to minimize IO time spent copying/moving files.

See above my experiments.

parm/land/prep/prep_run.yaml
CoryMartin-NOAA: Well that is weird... seems like a bug. But can you not just write:
```
  filename_sfcd: 'xainc.sfc_data.nc'
  filename_cplr: 'xainc.coupler.res'
```
?

jiaruidong2017: @CoryMartin-NOAA Do you want me just try: 
```
  filename_sfcd: 'xainc.sfc_data.nc'
  filename_cplr: 'xainc.coupler.res'
```

CoryMartin-NOAA: yes

jiaruidong2017: It is weird, because it produced the same output as previous.

jiaruidong2017: Please ignore my previous answers. The above settings could produce the correct output as below: 

```
20210323.060000.xainc.coupler.res
20210323.060000.xainc.sfc_data.tile1.nc
 20210323.060000.xainc.sfc_data.tile2.nc
20210323.060000.xainc.sfc_data.tile3.nc
20210323.060000.xainc.sfc_data.tile4.nc
20210323.060000.xainc.sfc_data.tile5.nc
20210323.060000.xainc.sfc_data.tile6.nc
```
It seems if prefix is not defined, the time tag will be added at the beginning by default. Therefore, we can use this settings to avoid copying/renaming files.

ush/land/letkf_create_ens.py
jiaruidong2017: @aerorahul This PR is the companion with the https://github.com/NOAA-EMC/global-workflow/pull/1636. When I use your implementation in your draft PR, this script will be removed.

jiaruidong2017:
This PR will be closed when PR #490 is merged into develop.

jiaruidong2017:
This PR is closed because PR #490  was merged into develop.

Issue #45, opened 2022-04-08
Create initial C96 (C48 analysis res) 3DVar regression test


CoryMartin-NOAA:
Will run 3DVar as part of ctest, compare output increment to staged increment on disk with `nccmp` and return successful if `fv3jedi_var.x` and `nccmp` have exit codes of 0.

CoryMartin-NOAA:
@RussTreadon-NOAA is this something you would like to tackle or would you prefer I do it (probably next week)?

RussTreadon-NOAA:
@CoryMartin-NOAA , I can work on this after prototype GDAS-JEDI cycling is in g-w.   fv3-jedi has 3dvar ctests.   Does it make sense to model our ctest based on these existing tests with our extensions added?

CoryMartin-NOAA:
@RussTreadon-NOAA thanks! Yes, what I used for inspiration was the hyb-3dvar test: https://github.com/JCSDA-internal/fv3-jedi/blob/develop/test/testinput/hyb-3dvar.yaml and just changed the **B**. I think this is a good idea to model ours based on what is in the authoritative repo, since then it will be easier to change our YAMLs if the syntax changes.

RussTreadon-NOAA:
Add self (@RussTreadon-NOAA ) to list of issue assignees

CoryMartin-NOAA:
Here is at least one goal for the sprint for next week @guillaumevernieres @RussTreadon-NOAA

guillaumevernieres:
Or we simply add a test section in our `yamls`?

CoryMartin-NOAA:
@guillaumevernieres sorry if I confused, I mainly meant we want a 3DVar C48 test for atm in our ctest ASAP (I think what you're saying is that OOPS can handle some basic testing capability within the exe?)

guillaumevernieres:
> @guillaumevernieres sorry if I confused, I mainly meant we want a 3DVar C48 test for atm in our ctest ASAP (I think what you're saying is that OOPS can handle some basic testing capability within the exe?)

@CoryMartin-NOAA , sorry for the confusion. Yes, I was just thinking as an alternative to nccmp we could just use what's provided in oops and only store a small ascii file instead of the full increment or state or ...

PR #496, opened 2023-05-31
Adding mhs_metop-b and mhs_n19 testing YAMLs

azadeh-gh:
Added MHS test YAMLs (mhs_metop-b and mhs_n19) to parm/atm/obs/testing in GDASApp to reproduce the GSI (HofX, QC and final obs error).
Bennartz Scattering QC check is used in JEDI YAML file, instead of MHS scattering index error inflation, so the number of obs passing QC in JEDI are less than GSI and the obs errors are different between JEDI and GSI. 
The QC flowchart and validation plots can be found in [Issue #45](https://github.com/NOAA-EMC/JEDI-T2O/issues/45)


ADCollard (approved):

CoryMartin-NOAA (approved):

nicholasesposito (commented):
Looks good!

PR #499, opened 2023-06-07
use jinja2 filter for boolean conversions

aerorahul:
Use a jinja filter to translate python bool to Fortran bool in the land analysis increment executable.
The global-workflow land DA PR https://github.com/NOAA-EMC/global-workflow/pull/1635 has been updated to use this


CoryMartin-NOAA (approved):
simple enough

Issue #498, opened 2023-06-05
marine DA verify task is failing


guillaumevernieres:
I have not check where yet ...

AndrewEichmann-NOAA:
@guillaumevernieres @CoryMartin-NOAA 
With the changes in PR https://github.com/NOAA-EMC/GDASApp/pull/507 in place, the vrfy ctest now fails when it doesn't find the file `build/test/soca/gw/testrun/testjjobs/ROTDIRS/gdas_test/gdas.20180415/12//analysis/atmos/gdas.t12z.cnvstat`. Is there some accounting for its absence, and do we want to leave it to break when it's not found (as opposed to just issuing a warning)?
EDIT: It dawns on me that this does not exist in the ctest - can we provide it? Or make it a warning?

guillaumevernieres:
> @guillaumevernieres @CoryMartin-NOAA With the changes in PR #507 in place, the vrfy ctest now fails when it doesn't find the file `build/test/soca/gw/testrun/testjjobs/ROTDIRS/gdas_test/gdas.20180415/12//analysis/atmos/gdas.t12z.cnvstat`. Is there some accounting for its absence, and do we want to leave it to break when it's not found (as opposed to just issuing a warning)? EDIT: It dawns on me that this does not exist in the ctest - can we provide it? Or make it a warning?

I just checked, that file is `164M` ... How about just issuing a warning?

CoryMartin-NOAA:
Or a test that creates a 'dummy' cnvstat file

guillaumevernieres:
My grepping isn't returning anything. Where are we using that file @AndrewEichmann-NOAA ?

AndrewEichmann-NOAA:
@guillaumevernieres That's just it - it gets produced with the atmospheric analysis and so is not present when you run just soca. Lil' ol' genius me tested it with cycling instead of the ctests.

guillaumevernieres:
> @guillaumevernieres That's just it - it gets produced with the atmospheric analysis and so is not present when you run just soca. Lil' ol' genius me tested it with cycling instead of the ctests.

Well, myself being a super genius, I knew that! I was just wondering where it was used in the ex script. I forgot you added that `diag_statistics.get_diag_stats()` a while back. 
[...]

PR #507, opened 2023-06-13
Changing handling of var.yaml to correct vrfy bug

AndrewEichmann-NOAA:
Currently the vrfy task fails when `ush/eva/gen_eva_obs_yaml.py` tries to ingest `var.yaml` when in encounters the line 304  `pattern: %mem%` because it doesn't like the unquoted % sign. `var.yaml` starts out with this quoted  like so: `pattern: '%mem%'`, but `scripts/exgdas_global_marine_analysis_run.sh ` strips the single quotes out of `var.yaml` apparently to placate `soca_var.x`. This PR changes the handling of `var.yaml` in `scripts/exgdas_global_marine_analysis_run.sh` so that the original is preserved for the vrfy task and its handling of yaml for eva in that it copies the file to a new name to be fed to `soca_var.x`

Tested with the soca ctest suite. vrfy still fails, but at a later point.

### DIFF
--- a/scripts/exgdas_global_marine_analysis_run.sh
+++ b/scripts/exgdas_global_marine_analysis_run.sh
@@ -39,8 +39,9 @@ function clean_yaml()
 
 ################################################################################
 # run 3DVAR FGAT
-clean_yaml var.yaml
-$APRUN_OCNANAL $JEDI_BIN/soca_var.x var.yaml
+cp var.yaml cleaned_var.yaml
+clean_yaml cleaned_var.yaml
+$APRUN_OCNANAL $JEDI_BIN/soca_var.x cleaned_var.yaml
 export err=$?; err_chk
 
 ################################################################################
--- a/scripts/exgdas_global_marine_analysis_run.sh
+++ b/scripts/exgdas_global_marine_analysis_run.sh
@@ -39,9 +39,9 @@ function clean_yaml()
 
 ################################################################################
 # run 3DVAR FGAT
-cp var.yaml cleaned_var.yaml
-clean_yaml cleaned_var.yaml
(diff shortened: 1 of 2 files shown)


AndrewEichmann-NOAA:
Partially addresses https://github.com/NOAA-EMC/GDASApp/issues/498

scripts/exgdas_global_marine_analysis_run.sh
```diff
@@ -39,8 +39,9 @@ function clean_yaml()
 
 ################################################################################
 # run 3DVAR FGAT
-clean_yaml var.yaml
-$APRUN_OCNANAL $JEDI_BIN/soca_var.x var.yaml
+cp var.yaml cleaned_var.yaml
+clean_yaml cleaned_var.yaml
+$APRUN_OCNANAL $JEDI_BIN/soca_var.x cleaned_var.yaml
 export err=$?; err_chk
```
guillaumevernieres, line 45: Seems fine, but I'd like to keep the `var.yaml` as the input to `var.x`. Can you instead copy `var.yaml` to `original_var.yaml` ... or something. You might have to copy  `original_var.yaml` in post.

AndrewEichmann-NOAA: You mean you want the "cleaned" `var.yaml `for the `var.x`?

guillaumevernieres: yes

AndrewEichmann-NOAA:
New version preserves modified `var.yaml` as such, copies to `var_original.yaml`

guillaumevernieres (approved):
:+1:

PR #516, opened 2023-06-26
Updated staticb_bump_aero.yaml to work with latest version of BUMP.  …

andytangborn:
Update of staticb_bump_aero.yaml  to work with latest version of BUMP.  Changed the name to what the workflow is expecting. This is currently working in a cycling experiment.


andytangborn:
@CoryMartin-NOAA I'm guessing that the test is failing because I changed the name of the yaml template, and genYaml probably doesn't know about it.

CoryMartin-NOAA:
@andytangborn indeed, that is the case it looks like. Can you fix that easily?

andytangborn:
Yes, I think I can.

CoryMartin-NOAA (approved):

PR #518, opened 2023-06-29
Fix non-SOCA tests

CoryMartin-NOAA:
As the title says, @AndrewEichmann-NOAA will tackle the remaining tests


AndrewEichmann-NOAA (approved):
Corresponding changes fix the test_gdasapp_soca_concatioda test (PR pending) so I say go for it

PR #537, opened 2023-07-14
Move fieldmetadata to inside version control rather than glopara fix

CoryMartin-NOAA:
I think out of convenience before, the glopara/fix/gdas directories contain several text files that do not need to be maintained by the glopara group, and can instead store them in GDASApp or global-workflow.

This PR is the first step towards that, by moving the FV3-JEDI IO fieldmetadata files into GDASApp. Subsequent work will need to determine if things like FV3 diag tables can be shared between FV3-JEDI and the UFS.


CoryMartin-NOAA:
A companion PR after this one will need to be made to global-workflow to update to this hash + modify scripts to look in the new location.

DavidNew-NOAA (dismissed):
Looks good to me.

aerorahul (dismissed):
looks good to me.

ush/land/genYAML_output_letkfoi.yaml
```diff
@@ -20,7 +20,7 @@ driver:
   save posterior mean increment: 'true'
 geometry:
   akbk: Data/fv3files/akbk127.nc4
-  field metadata override: Data/fieldmetadata/gfs-land.yaml
+  field metadata override: Data/fieldmetadata/fv3jedi_fieldmetadata_restart.yaml
```
jiaruidong2017, line 23: Need change to: 
 field metadata override: $(DATA)/fv3jedi/fv3jedi_fieldmetadata_restart.yaml

jiaruidong2017 (commented):
One concern is that all the field meta data were merged into a single file: `fv3jedi_fieldmetadata_restart.yaml`. Once there is any change to the meta data from fv3-jedi, how does GDASApp make the updates?

CoryMartin-NOAA:
@jiaruidong2017 we just edit the file that I've added here. Not sure I understand the concern (unless I'm missing something)

jiaruidong2017:
@CoryMartin-NOAA What I mean is, do we have a tool which can monitor the changes from `fv3-jedi` and update this file automatically? For example, when the GDASApp was built, the meta data were  already updated in the directory: `$(project_source_dir)/fv3-jedi/test/Data/fieldmetadata/`. Can we simply use or link these files so that we don't need to monitor their changes?

CoryMartin-NOAA:
@jiaruidong2017 I think we want to know when this changes though, not anything automatic. Because this file contains the information on how to read the model state into JEDI. If anything changes here, we would need the model forecasts to change appropriately. We have the tests so that we know if it starts failing that something changed.

jiaruidong2017:
@CoryMartin-NOAA Okay, I understand what you mean `inside version control` from your title. Thanks for answer my questions.

CoryMartin-NOAA:
Tests will fail until the workflow is updated. I have tested this manually on Hera and all tests pass:
```
      Start 1501: test_gdasapp_atm_jjob_var_init
43/49 Test #1501: test_gdasapp_atm_jjob_var_init ........................   Passed   44.30 sec
      Start 1502: test_gdasapp_atm_jjob_var_run
44/49 Test #1502: test_gdasapp_atm_jjob_var_run .........................   Passed   74.40 sec
      Start 1503: test_gdasapp_atm_jjob_var_final
45/49 Test #1503: test_gdasapp_atm_jjob_var_final .......................   Passed   10.10 sec
      Start 1504: test_gdasapp_atm_jjob_ens_init
46/49 Test #1504: test_gdasapp_atm_jjob_ens_init ........................   Passed   43.70 sec
      Start 1505: test_gdasapp_atm_jjob_ens_run
47/49 Test #1505: test_gdasapp_atm_jjob_ens_run .........................   Passed  170.86 sec
      Start 1506: test_gdasapp_atm_jjob_ens_final
48/49 Test #1506: test_gdasapp_atm_jjob_ens_final .......................   Passed   42.79 sec
      Start 1507: test_gdasapp_aero_gen_3dvar_yaml
49/49 Test #1507: test_gdasapp_aero_gen_3dvar_yaml ......................   Passed    0.25 sec

100% tests passed, 0 tests failed out of 49
```

This will need to be approved and merged before the workflow tests will pass, which will require a global-workflow PR, which will be a bit of a chicken/egg problem.

ush/land/genYAML_output_letkfoi.yaml
jiaruidong2017: With the recent commit, this issue was resolved.

jiaruidong2017 (approved):
Looks good to me now.

PR #538, opened 2023-07-17
Refactoring of vrfy - transferred bulk of plotting to sub-script

AndrewEichmann-NOAA:
No functional changes, just moved plotting-specific code from vrfy script to script in ush. Also cleaned up some superfluous variables.


guillaumevernieres (changes requested):
Still a bit of a mess ... Which I created!

1 - For the state figures, we have 4 cases that are generating similar figures, the difference being the bounds, and input files ...

Instead of splitting into `plot_analysis` and `plot_increment`, split into:

- `plot_increment`
- `plot_background`
- `plot_analysis`
- `plot_bkgerr`

Ideally, we should create a simple `ViewState` class that would look something like this:
```python
class ViewState:
    def __init__(self, config_dict):
        self.config = config_dict
        self.stuff = config_dict['stuff']
        ...
        # Where config contains the bounds, filename and whatever else is needed to plot stuff

    def plot(self):
        # Loop over variables, slices (horiz and vertical) and projections ... and whatever else is needed
```
Now instead of having 4 separate function, you have one class that you "instantiate" 4 times with different configurations. In the ex global script, you'd do something like this:
```python
config= ...
viewIncr = ViewState(config)
viewIncr.plot()

config= ...
viewBkg = ViewState(config)
viewBkg.plot()

config= ...
viewAna = ViewState(config)
viewAna.plot()

config= ...
viewBkgErr = ViewState(config)
viewBkgErr.plot()
```
That should remove quite a bit of code and make things a bit more readable.
Hope that makes sense, happy to chat if not.
 
2 - Move the `soca_vrfy.py` into ush/soca

AndrewEichmann-NOAA:
@guillaumevernieres This is a minor thing, but should `ush/diag_statistics.py` go into soca as well? It's run only from our vrfy task

guillaumevernieres:
> @guillaumevernieres This is a minor thing, but should `ush/diag_statistics.py` go into soca as well? It's run only from our vrfy task

If we're the only one using it, I would say yes.

AndrewEichmann-NOAA:
@guillaumevernieres I think you'll like this new version. I've changed my mind about hiding the plot settings, the way this is you can easily tweak the nobs on the front panel, so to speak. I looked at the plots and to the extent that the ctests exercise them they seem the same as develop, but now all of GDASApp appears to be broken, so feh

scripts/exgdas_global_marine_analysis_vrfy.py
```diff
@@ -22,189 +22,147 @@
...
-                     colormap=incr_cmap,
+                     lats=np.arange(-60, 60, 10),
+                     variables_zonal=['Temp', 'Salt'],
+                     variables_horiz=['Temp', 'Salt', 'ave_ssh'],
+                     allbounds={'Temp': [-0.5, 0.5],
+                                'Salt': [-0.1, 0.1],
+                                'ave_ssh': [-0.1, 0.1]},
+                     colormap='RdBu',
                      comout=os.path.join(comout, 'vrfy', 'incr'))
+OcnIncPlotter = StatePlotter(config)
```
guillaumevernieres: if you are going to use camel case style of naming stuff, variables should start with a lower case.

guillaumevernieres (commented):
Thanks for doing this @AndrewEichmann-NOAA . Looks good. Just one comment related to variable naming convention.

AndrewEichmann-NOAA:
Now python coding style is failing on `ush/ufsda/genYAML.py`, which I don't touch

guillaumevernieres:
> Now python coding style is failing on `ush/ufsda/genYAML.py`, which I don't touch

It looks like it's a simple fix @AndrewEichmann-NOAA . Maybe `pycodestyle` was recently updated. Can you try what's suggested, replace `==` with `is` [L34 of genYAML.py](https://github.com/NOAA-EMC/GDASApp/blob/48c9ad8a48ba1698b164d42caf48264d7363e758/ush/ufsda/genYAML.py#L34)

AndrewEichmann-NOAA:
@guillaumevernieres This currently passes soca ctests on Hera

guillaumevernieres (approved):
:+1:

PR #544, opened 2023-07-26
rename yaml file for bump aero DA

CoryMartin-NOAA:
as the title says, this is to make things easier in the workflow


andytangborn (approved):
Looks good.

PR #551, opened 2023-08-07
removed duplicate variable specification

AndrewEichmann-NOAA:
Tweak to recent refactor of vrfy task - made it so that config for plotter classes had variable specified only once, instead of twice when setting the colorbar bounds


guillaumevernieres (approved):
:+1:

PR #585, opened 2023-08-25
Hotfix for bugfix

CoryMartin-NOAA:
Missing one thing that is needed to actually fix issue in PR#584


RussTreadon-NOAA (approved):
Correct, need to specify the number of ensembles.

Approve.

PR #588, opened 2023-08-28
Add in basic C48 aerosol analysis tests that use global-workflow

CoryMartin-NOAA:
As the title says, this PR adds in 3 new j-job ctests that use the global-workflow to run a C48 JEDI aerosol 3DVar analysis. Note this will fail until the workflow is updated to require just 6 PEs for the C48 aerosol analysis case. Also note, for now, the **B** is just identity, which isn't scientifically valid, nor does it exercise the full 'plumbing' but is a good start to a test.


guillaumevernieres (approved):
I'm not a big fan of all that shell scripting, but if it works!

andytangborn (approved):
Good to know that it works.

PR #601, opened 2023-09-05
replace run_bufr2ioda.sh with run_bufr2ioda.py

### LINKED ISSUE #599: Refactor run_bufr2ioda.sh as run_bufr2ioda.py
g-w PR [#1826](https://github.com/NOAA-EMC/global-workflow/pull/1826) states that [all new scripts added to g-w need to be in python](https://github.com/NOAA-EMC/global-workflow/pull/1826#discussion_r1311665469).   

GDASApp `ush/ioda/bufr2ioda/run_bufr2ioda.sh` is a bash script executed from g-w `jobs/JGLOBAL_PREP_ATM_IODA_OBS`.  As such, g-w requires this script to python, not bash.

This issue is opened to document the refactoring of `run_bufr2ioda.sh` to `run_bufr2ioda.py`.

RussTreadon-NOAA:
g-w PR [#1826](https://github.com/NOAA-EMC/global-workflow/pull/1826) states that `ush/ioda/bufr2ioda/run_bufr2iobs.sh` must be in python (see [comment here](https://github.com/NOAA-EMC/global-workflow/pull/1826#discussion_r1311665469)).  

Given this, GDASApp branch [`feature/bufr2iodapy`](https://github.com/NOAA-EMC/GDASApp/tree/feature/bufr2iodapy) was created to replace `run_bufr2ioda.sh` with `run_bufr2ioda.py`.

To fully integrate `run_bufr2ioda.py` into g-w the decision was made to replace variable `DUMP` in `parm/ioda/bufr2ioda/bufr2ioda_satwind_amv_goes.json` with `CDUMP`, the current cycle.   With this change tests reported in g-w [#1826](https://github.com/NOAA-EMC/global-workflow/pull/1826) verify that `run_bufr2ioda.py` generates output identical to that from `run_bufr2ioda.sh`.

This PR is opened to replace `run_bufr2ioda.sh` with `run_bufr2ioda.py` and update `parm/ioda/bufr2ioda/bufr2ioda_satwind_amv_goes.json` as described above.

Fixes #599


parm/ioda/bufr2ioda/bufr2ioda_satwind_amv_goes.json
```diff
@@ -1,7 +1,7 @@
 {
   "data_format"      : "bufr_d",
   "data_type"        : "satwnd",
-  "cycle_type"       : "{{ DUMP }}",
```
CoryMartin-NOAA, line 4: I think I remember @aerorahul saying we should go to `RUN` eventually instead of `DUMP` but perhaps I'm misremembering.

ush/ioda/bufr2ioda/run_bufr2ioda.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env python3
+import os
+import subprocess
+
+
+def main():
+    # Check command line arguments
+    if len(sys.argv) != 6:
```
CoryMartin-NOAA: we might want to use `argparse` here now instead rather than all the `sys.argv` values

ush/ioda/bufr2ioda/run_bufr2ioda.py
```diff
@@ -0,0 +1,60 @@
...
+
+    # Specify observation types to be processed by a script
+    BUFR_py = ["satwind_amv_goes"]
+
+    for obtype in BUFR_py:
+        print(f"Processing {obtype}...")
+        json_output_file = os.path.join(COM_OBS, f"{obtype}_{PDY}{cyc}.json")
+
+        # Generate a JSON from the template
+        subprocess.run([BUFRJSONGEN, "-t", f"{config_template_dir}/bufr2ioda_{obtype}.json", "-o", json_output_file], check=True)
```
CoryMartin-NOAA: If this is all python, we can call `gen_bufr_json` from this script and store the JSON config in memory and then pass it to the subsequent function in the other python script.

ush/ioda/bufr2ioda/run_bufr2ioda.py
```diff
@@ -0,0 +1,60 @@
...
+
+        # Use the converter script for the ob type
+        subprocess.run(["python3", f"{os.path.join(USH_IODA, 'bufr2ioda_' + obtype + '.py')}", "-c", json_output_file], check=True)
```
CoryMartin-NOAA: we can `import` each script in this loop and then call the function from within it that actually does the work, `bufr_to_ioda(config, logger)`

ush/ioda/bufr2ioda/run_bufr2ioda.py
```diff
@@ -0,0 +1,60 @@
...
+
+        # Check if the converter was successful
+        if os.path.exists(json_output_file):
+            os.remove(json_output_file)
+        else:
+            print(f"Problem running converter script for {obtype}")
+
+
+if __name__ == "__main__":
+    import sys
```
CoryMartin-NOAA: If using `sys`, this should probably be at the top

CoryMartin-NOAA (commented):
Nothing wrong with what is here, but this is mostly a direct translation of bash to python, if we are using python, we should take advantage of that fact.

parm/ioda/bufr2ioda/bufr2ioda_satwind_amv_goes.json
RussTreadon-NOAA: Replace `DUMP` with `RUN`.   Done at e9515e0.

g-w sets `RUN` to `gfs` and `gdas` when running gfs and gdas cycle, respectively.  `RUN` is `enkfgfs` or `enkfgdas` when running enkf jobs.  Since g-w `prepatmiodaobs` job is associated with `gfs` or `gdas`, we're OK.

ush/ioda/bufr2ioda/run_bufr2ioda.py
RussTreadon-NOAA: Replace `sys.argv` with `argparse`.  Done at 75e5b9d.

ush/ioda/bufr2ioda/run_bufr2ioda.py
RussTreadon-NOAA: Replace `BUFRJSONGEN` with `gen_bufr_json`.  Done at 75e5b9d.   Still retain `json_output_file` since I am not yet directly invoking `bufr_to_ioda` (see below).

ush/ioda/bufr2ioda/run_bufr2ioda.py
RussTreadon-NOAA: Google searches suggest that it is not good practice to import python functions inside a loop.  If true, how do we handle the fact that `bufr_to_ioda` changes depending on the `obtype` being processed?

CoryMartin-NOAA: Good question. I might need to think about this for a bit. I think this is a case where a 'factory' would be appropriate, but I'm not sure. Basically, we 'register' each of the available options and import them all together and then the loop decides which to call. But not 100% sure how to best implement.

ush/ioda/bufr2ioda/run_bufr2ioda.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def bufr2ioda(CDATE, RUN, DMPDIR, config_template_dir, COM_OBS):
+    logger.info(f"Process {CDATE} {RUN} from {DMPDIR} to {COM_OBS} using {config_template_dir}")
+
+    # Derived parameters
+    PDY = CDATE[:8]
+    cyc = CDATE[8:10]
+
+    config = cast_strdict_as_dtypedict(os.environ)
```
aerorahul: What is the point of passing arguments to this method, if os.environ is going to be invoked here.

ush/ioda/bufr2ioda/run_bufr2ioda.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env python3
+import argparse
+import os
+import subprocess
```
aerorahul: Please use Executable from wxflow.

ush/ioda/bufr2ioda/run_bufr2ioda.py
RussTreadon-NOAA: No point apart from my limited python skills leading me to copy what I see in other scripts.  Remove `os.environ`.  Done at [`2510b32`](https://github.com/NOAA-EMC/GDASApp/commit/2510b32a33a7b5060a227a83cebe002e797acdd6).

ush/ioda/bufr2ioda/run_bufr2ioda.py
RussTreadon-NOAA: `sys` removed

ush/ioda/bufr2ioda/run_bufr2ioda.py
RussTreadon-NOAA: `subprocess` removed in favor of `Executable`.  Done at [`2510b32`](https://github.com/NOAA-EMC/GDASApp/commit/2510b32a33a7b5060a227a83cebe002e797acdd6)

RussTreadon-NOAA:
Test GDASApp branch `feature/bufr2iodapy` along with g-w branch `RussTreadon-NOAA:feature/iodaobs` on Hera and Orion for 20210814 18 gdas.   Job `prepatmiodaobs` ran to completion on both machines.  IODA dump files were created for GOES-16 and GOES-17 ABI from the bufr satwnd dump file.

Does this PR require any additional revisions?

CoryMartin-NOAA (dismissed):
Good for me for now, we can revisit later if we find a 'clever' way to do this in python

ush/ioda/bufr2ioda/run_bufr2ioda.py
```diff
@@ -0,0 +1,67 @@
...
+        template = config_template_dir + '/bufr2ioda_' + obtype + '.json'
+        gen_bufr_json(config, template, json_output_file)
+
+        # Use the converter script for the ob type
+        bufr2iodapy = USH_IODA + '/bufr2ioda_' + obtype + ".py"
+        cmd = Executable(bufr2iodapy)
+        cmd.add_default_arg('-c')
+        cmd.add_default_arg(json_output_file)
+        logger.info(f"Executing {cmd}")
+        cmd()
```
aerorahul, line 48: If this is a python script, then why do we need to step out of the interpreter into a subprocess?

ush/ioda/bufr2ioda/run_bufr2ioda.py
```diff
@@ -0,0 +1,67 @@
...
+import os
+from gen_bufr2ioda_json import gen_bufr_json
+from wxflow import Logger, Executable, cast_as_dtype
+
+# Initialize root logger
+logger = Logger('run_bufr2ioda.py', level='INFO', colored_log=True)
+
+
+def bufr2ioda(CDATE, RUN, DMPDIR, config_template_dir, COM_OBS):
+    logger.info(f"Process {CDATE} {RUN} from {DMPDIR} to {COM_OBS} using {config_template_dir}")
```
aerorahul:
```suggestion
@logit(logger)
def bufr2ioda(CDATE, RUN, DMPDIR, config_template_dir, COM_OBS):
    logger.info(f"Process {CDATE} {RUN} from {DMPDIR} to {COM_OBS} using {config_template_dir}")
```

ush/ioda/bufr2ioda/run_bufr2ioda.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python3
+import argparse
+import os
+from gen_bufr2ioda_json import gen_bufr_json
+from wxflow import Logger, Executable, cast_as_dtype
```
aerorahul:
```suggestion
from wxflow import Logger, Executable, cast_as_dtype, logit
```

ush/ioda/bufr2ioda/run_bufr2ioda.py
```diff
@@ -0,0 +1,67 @@
...
+        # Check if the converter was successful
+        if os.path.exists(json_output_file):
+            os.remove(json_output_file)
+        else:
+            print(f"Problem running converter script for {obtype}")
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Convert bufr dump file to ioda format')
+    parser.add_argument('CDATE', type=str, help='YYYYMMDDHH date to process')
```
aerorahul:
```suggestion
    parser.add_argument('current_cycle', help='current cycle to process',
                            required=True, type=lambda dd: to_datetime(dd))
```
There is no need to use upper case for things like `CDATE`.  That is a convention used in shell scripts to define (environment) variables.
I would also suggest standardizing against `current_cycle`

ush/ioda/bufr2ioda/run_bufr2ioda.py
```diff
@@ -0,0 +1,67 @@
...
+    parser.add_argument('RUN', type=str, help='dump to process, either gdas or gdas')
+    parser.add_argument('DMPDIR', type=str, help='path to bufr dump files')
+    parser.add_argument('config_template_dir', type=str, help='path to templates')
+    parser.add_argument('COM_OBS', type=str, help='path to output ioda format dump files')
+    args = parser.parse_args()
+    bufr2ioda(args.CDATE, args.RUN, args.DMPDIR, args.config_template_dir, args.COM_OBS)
```
aerorahul:
```suggestion
    bufr2ioda(args.CDATE, args.RUN, args.DMPDIR, args.config_template_dir, args.COM_OBS)
```

ush/ioda/bufr2ioda/run_bufr2ioda.py
```diff
@@ -0,0 +1,67 @@
...
+            os.remove(json_output_file)
+        else:
+            print(f"Problem running converter script for {obtype}")
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Convert bufr dump file to ioda format')
+    parser.add_argument('CDATE', type=str, help='YYYYMMDDHH date to process')
+    parser.add_argument('RUN', type=str, help='dump to process, either gdas or gdas')
+    parser.add_argument('DMPDIR', type=str, help='path to bufr dump files')
```
aerorahul: This is a path, not a `str`.  So we should use that information to provide accurate type definition.
```suggestion
    parser.add_argument('DMPDIR', type=pathlib.Path, help='path to bufr dump files')
```
same with `COM_OBS` and `config_template_dir`

ush/ioda/bufr2ioda/run_bufr2ioda.py
```diff
@@ -0,0 +1,67 @@
...
+        cmd.add_default_arg('-c')
+        cmd.add_default_arg(json_output_file)
+        logger.info(f"Executing {cmd}")
+        cmd()
+
+        # Check if the converter was successful
+        if os.path.exists(json_output_file):
+            os.remove(json_output_file)
+        else:
+            print(f"Problem running converter script for {obtype}")
```
aerorahul: We are using a logger. So use it and raise exceptions when there is a problem

ush/ioda/bufr2ioda/run_bufr2ioda.py
```diff
@@ -0,0 +1,67 @@
...
+
+    # Get gdasapp root directory
+    DIR_ROOT = os.path.realpath(os.path.join(os.path.dirname(os.path.realpath(__file__)), "../../.."))
+    USH_IODA = os.path.join(DIR_ROOT, "ush", "ioda", "bufr2ioda")
+
+    # Create output directory if it doesn't exist
+    os.makedirs(COM_OBS, exist_ok=True)
+
+    # Specify observation types to be processed by a script
+    BUFR_py = ["satwind_amv_goes"]
```
aerorahul, line 33: Why can't this be an input argument?  Seems like this is the only hard-wired variable in a mostly generic script

ush/ioda/bufr2ioda/run_bufr2ioda.py
```diff
@@ -0,0 +1,67 @@
...
+
+    for obtype in BUFR_py:
+        logger.info(f"Convert {obtype}...")
+        json_output_file = os.path.join(COM_OBS, f"{obtype}_{PDY}{cyc}.json")
+        template = config_template_dir + '/bufr2ioda_' + obtype + '.json'
```
aerorahul: Use `os.path.join` instead of string joins

aerorahul (changes requested):

ush/ioda/bufr2ioda/run_bufr2ioda.py
RussTreadon-NOAA: Agreed.  

In principal we should not & do not want to step out of the python script via a subprocess to invoke another python script.  However, at present each observation type has specific scripting to generate an ioda format dump file the source bufr dump file.   A python [factory](https://github.com/NOAA-EMC/GDASApp/pull/601#discussion_r1316262132) may be the best way to tackle this.  This noted, a python converter may not be the long term solution.  We may opt to create ioda format dumps via c++ code.  This c++ code might eventually move into JEDI as a bufr to ioda interface. 

The _python scripts_ approach introduced by this PR along with g-w PR [#1826](https://github.com/NOAA-EMC/global-workflow/pull/1826) are needed for JEDI ATM prototyping.   We want to move away from the current approach of (a) run `gsi.x` in observer mode to create netcdf diagnostic files followed by (b) convert `gsi.x` netcdf diagnostic files into ioda format observation files.

aerorahul: You may want to consider looking into the `importlib` module.  It provides the same functionality as `import` does at the top of the script, but is designed for this particular type of situation you are in where you do not want to import everything at the top.

ush/ioda/bufr2ioda/run_bufr2ioda.py
RussTreadon-NOAA: Suggestion accepted at cff2ce5.  Question:  why do we add `@logit(logger)`?

ush/ioda/bufr2ioda/run_bufr2ioda.py
RussTreadon-NOAA: Suggestion accepted at cff2ce5.

ush/ioda/bufr2ioda/run_bufr2ioda.py
RussTreadon-NOAA: Suggestion accepted at cff2ce5  with one exception.   Inclusion of `required=True` generated the run time error

`TypeError: 'required' is an invalid argument for positionals`

Remove `required=True` and script executes.  

Added `to_datetime` to `from wxflow import` line at top of `run_bufr2ioda.py`.

ush/ioda/bufr2ioda/run_bufr2ioda.py
RussTreadon-NOAA: Not sure what is being suggested here.   Is the suggestion to replace `args.CDATE` with `args.current_cycle`?   If so, suggestion accepted at cff2ce5.

ush/ioda/bufr2ioda/run_bufr2ioda.py
RussTreadon-NOAA: Suggestion accepted at cff2ce5 with following tweaks
1. add `from pathlib import Path` to top of `run_bufr2ioda.py`
2. specify type as `Path`, not `pathlib.Path`

ush/ioda/bufr2ioda/run_bufr2ioda.py
RussTreadon-NOAA: Suggestion accepted at cff2ce5.

ush/ioda/bufr2ioda/run_bufr2ioda.py
RussTreadon-NOAA: Valid point.   At present we only have one bufr2ioda conversion script, one for GOES satwind observations.   The `BUFR_py` list will grow as more bufr2ioda conversion scripts are created.   I defer to @CoryMartin-NOAA and @emilyhcliu as to how they would like to handle this.

ush/ioda/bufr2ioda/run_bufr2ioda.py
RussTreadon-NOAA: Suggestion accepted at cff2ce5 with one exception.   `bufr2ioda_{obtype}.json` is a filename.   Refactor above as
```
        filename = 'bufr2ioda_' + obtype + '.json'
        template = os.path.join(config_template_dir, filename)
```

ush/ioda/bufr2ioda/run_bufr2ioda.py
aerorahul: `logit` is a decorator.  Its use is to log the input and output of the function it decorates.
Adding that one liner, will yield to the user, what are the inputs being sent to the function and what the outputs are being returned from the function.  Additionally, it also logs the entry and exit times that can be useful for profiling.
If any additional logging or profiling stats (e.g. memory) are desired, they can be implemented in `logit` and the entire project can reap the benefits.

RussTreadon-NOAA: Thank you @aerorahul for explaining what `@logit` brings to the table.

ush/ioda/bufr2ioda/run_bufr2ioda.py
aerorahul: Instead of this `if-else`, you should use `wxflow.rm_p`.
It not only traps any OSerror it encounters, but will also check if the file is present.
I am just alerting of the tools we have to handle some of these basic functions without having to write too much individual code.

RussTreadon-NOAA: Suggestion accepted at c0c0910.

ush/ioda/bufr2ioda/run_bufr2ioda.py
RussTreadon-NOAA: Thank you @aerorahul for this suggestion.  Something to consider in future revisions.

aerorahul (approved):

PR #606, opened 2023-09-08
Unfinished example of a RADS to ioda converter

### LINKED ISSUE #607: Create template and example based on RADS altimetry to ioda

guillaumevernieres:
## Description
This PR adds a prototype IODA converter that will be used as a template for the other netdcf observation files that will need to be converted. 
I added a base class for the conversion in ```NetCDFToIodaConverter.h```, it probably could do more but for now the only thing useful it does is saving the ioda file.

### RADS to IODA implementation
- The derived class from the above that actually does something useful is called Rads2Ioda and it implements the reading of the RADS netcdf files, which is a virtual function of ```NetCDFToIodaConverter```
- The application code is under obsproc/applications

### Testing
- I added a sub-sampled cdl RADS file that is converted back to netcdf in a ctest
- The ctest reads the RADS netcdf file and creates a IODA file. It's not testing anything more than successful completion
- The converter is not complete, it's missing most of the MetaData and a few other things ... 

### Issues addressed
- fixes #607


utils/obsproc/NetCDFToIodaConverter.h
```diff
@@ -0,0 +1,61 @@
+#ifndef GDAS_UTILS_OBSPROC_NETCDFTOIODACONVERTER_H_  //NOLINT
```
CoryMartin-NOAA:
```suggestion
#pragma once
```

utils/obsproc/Rads2Ioda.h
```diff
@@ -0,0 +1,89 @@
...
+#include "NetCDFToIodaConverter.h"
+
+#include "oops/util/missingValues.h"
+
+namespace gdasapp {
+  class Rads2Ioda : public NetCDFToIodaConverter {
+   public:
+    explicit Rads2Ioda(const eckit::Configuration & fullConfig)
+    : NetCDFToIodaConverter(fullConfig) {
+    oops::Log::info() << this->windowBegin_ << std::endl;
```
CoryMartin-NOAA: How about some strings that say 'window begin:', etc.

utils/obsproc/NetCDFToIodaConverter.h
CoryMartin-NOAA: Just pragma once and remove all the ifndef stuff

utils/obsproc/Rads2Ioda.h
guillaumevernieres: ho ... in the log info @CoryMartin-NOAA ? Yes, good point.

utils/obsproc/NetCDFToIodaConverter.h
guillaumevernieres: Ha! yes ... dummy me. Are we sure it's doing the same as the ifndef stuff?

CoryMartin-NOAA: Indeed, ask ChatGPT to confirm though :-)

guillaumevernieres: Just did, the only issue is portability. It's not part of the c++ standard but it's widely supported. Works for me!

CoryMartin-NOAA (approved):
image

guillaumevernieres:
I think the test will fail, there's a bug in the oops develop that was probably cloned by the ci ... The culprit pr was just reverted ~30mn ago. Zut!

ShastriPaturi (approved):

apchoiCMD (approved):

AndrewEichmann-NOAA (approved):
Looks ok - what's a cdl file?

CoryMartin-NOAA:
@AndrewEichmann-NOAA it's a text file in the same format as what ncdump dumps out that you can use to create a netCDF file.

apchoiCMD (approved):

PR #623, opened 2023-09-13
Fixed typo in satwind_goes yaml files

DavidNew-NOAA:
Removed an extra '$' from the satwind_goes-*.yaml obs files. With the extra '$', the obs files will look something like: gdas.tHHz.satwind.abi_goes-16.$YYYYMMDDHH.nc4  rather than the correct gdas.tHHz.satwind.abi_goes-16.YYYYMMDDHH.nc4


CoryMartin-NOAA (approved):
thanks @DavidNew-NOAA !

PR #631, opened 2023-09-17
Update test reference for both satwind and scatwind.

emilyhcliu:
Minor changes....

Modifications made for satwind and scatwind with the following:

- Update benchmark 
   More number of observations passed QC with the current UFO develop
   6 more points for scatwind and 25056 points more for satwind.
   (Looks a lot more passed for satwind, probably because we do not do data thinning for satwind.)

- Change BlackList to RejectList for scatwind
- Update ObsFunction SatWindsSPDBCheck to WindsSPDBCheck for scatwind (satwind had already been fixed in previous PR)
- Change `land_Type_index_NPOESS` with `water_area_fraction` for scatwind
- Add tests without QC

### DIFF
--- a/parm/atm/obs/testing/satwind.yaml
+++ b/parm/atm/obs/testing/satwind.yaml
@@ -989,7 +989,8 @@ obs post filters:
 linear obs operator:
   name: Identity
 
-passedBenchmark: 1025814 # 2 variables (u,v), both passing 512907 obs, including 512907 GSI/UFO agreements, and:
+#passedBenchmark: 1025814# 2 variables (u,v), both passing 512907 obs, including 512907 GSI/UFO agreements, and:
+passedBenchmark: 1150870 # 2 variables (u,v), both passing 575435 obs, including 512907 GSI/UFO agreements, and:
                          # 16 GOES AMVs (6 u-obs, 6 v-obs) that are rejected by the SatWindsErrnormCheck in UFO
                          # but are retained in GSI's equivalent experr_norm check. All 6 of these disagreements
                          # have error norm values at almost exactly 0.9, indicating a float precision/handling
--- a/parm/atm/obs/testing/satwind_noqc.yaml
+++ b/parm/atm/obs/testing/satwind_noqc.yaml
@@ -1,0 +1,18 @@
+obs space:
+  name: satwind
+  obsdatain:
+    engine:
+      type: H5File
+      obsfile: !ENV satwind_obs_${CDATE}.nc4
+  obsdataout:
+    engine:
+      type: H5File
+      obsfile: !ENV satwind_diag_${CDATE}.nc4
+  simulated variables: [windEastward, windNorthward]
+geovals:
+  filename: !ENV satwind_geoval_${CDATE}.nc4
+obs operator:
+  name: VertInterp
+
+vector ref: GsiHofXBc
+tolerance: 1.e-5
--- a/parm/atm/obs/testing/scatwind.yaml
+++ b/parm/atm/obs/testing/scatwind.yaml
@@ -60,8 +60,8 @@ obs post filters:
     where:
     - variable: ObsType/windEastward
       is_in: 290
-    - variable: GeoVaLs/land_type_index_NPOESS
-      minvalue: 1. 
+    - variable: GeoVaLs/water_area_fraction
+      maxvalue: 0.001 
     action:
       name: reject
   # Reject ASCAT (Type 290) when observed component deviates from background by more than 5.0 m/s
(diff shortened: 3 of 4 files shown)


BrettHoover-NOAA:
Hi @emilyhcliu - for the sake of my understanding, what change to UFO is causing the satwind benchmark to increase the number of passed winds in UFO by 25056?

emilyhcliu:
@BrettHoover-NOAA
I have satwind goes-17 end-to-end test (using IODA obs file and fv3 background file)
The end-to-end test is running with GDASApp, which has the UFO develop from JCSDA.

Here is the result:  There are six time-frame data from satwind in a cycle.  I am showing the first timeframe of data  for U-wind (UFO vs. GSI).  The grey spots are data points before QC.  The UFO has more input points because are using bufr converted ioda obs file.  The GSI has fewer input points (grey) because the input is from gsi nc-diagnostic file converted IODA obs file.  Some data already cycled out in the read-routine (e.g. obs from the edge of the scan etc...).  In the end-to-end test, the data passed QC from GSI and UFO very close, and the statistics are similar. 

ufo_abi_goes-17_omf_windEastward_qc_time1

gsi_abi_goes-17_omf_windEastward_qc_time1

Obs Errors are good as well:
ufo_abi_goes-17_obserr_windEastward_qc_time1
gsi_abi_goes-17_obserr_windEastward_qc_time1

Not sure why the satwind ctest has much more data passed the QC then GSI with the latest UFO.
Let me double-check the test data set to ensure I did not make a mistake in generating the dataset.
I did not follow GMAO's change closely, I will take a look.  
When you have time, could you also run the satwind ctest and see which checks allow more data to pass QC?

BrettHoover-NOAA:
@emilyhcliu I appreciate the explanation! I am confused why UFO is capable of passing more winds if the difference in counts is caused by UFO pulling from the entire BUFR file while GSI is only using observations that were not hard-cycled during prep-QC and survived processing all the way through to the diagnostic file. The UFO may be able to see these observations, but should be rejecting them in its own pre-QC checks, unless UFO is missing some pre-QC checks that didn't need to be accounted-for when it was reading from the diagnostic --> IODA converted files?

Is there a way for me to run the filtering test with these changes from your files, or get access to the diagnostic file you are producing from UFO? I'd like to try running the output through my simulated satwind QC code in python to see if obs are getting through UFO QC that shouldn't be.

emilyhcliu:
> @emilyhcliu I appreciate the explanation! I am confused why UFO is capable of passing more winds if the difference in counts is caused by UFO pulling from the entire BUFR file while GSI is only using observations that were not hard-cycled during prep-QC and survived processing all the way through to the diagnostic file. The UFO may be able to see these observations, but should be rejecting them in its own pre-QC checks, unless UFO is missing some pre-QC checks that didn't need to be accounted-for when it was reading from the diagnostic --> IODA converted files?
> 
> Is there a way for me to run the filtering test with these changes from your files, or get access to the diagnostic file you are producing from UFO? I'd like to try running the output through my simulated satwind QC code in python to see if obs are getting through UFO QC that shouldn't be.

@BrettHoover-NOAA 
**End-to-End test** for GOES-17 AMVs:

UFO output: /work/noaa/da/eliu/JEDI-GDAS/BufrConverterTest/diags/2021080100/satwnd.abi_goes-17_hofx_2021080100.nc4

The UFO output is verified with our ctest obs file
/work2/noaa/da/eliu/UFO_eval/data/gsi_geovals_l127/nofgat_aug2021/20230811/obs/satwind_obs_2021080100.nc4

The GDASApp Ctest for satwind:
Output:/work2/noaa/da/eliu/ufoeval_satwind/2021080100/satwind/satwind_diag_2021080100.nc4

Are the above-listed files what you want? 
Please let me know if I misunderstood or if you want more information.

emilyhcliu:
We are looking for the cause of the change in data passed QC in satwind ctest.
So, I am converting this PR to a draft one.

emilyhcliu:
@BrettHoover-NOAA 
The plots are from satwind ctest (updated benchmark in this PR):
The Hofxdiff  vs. QCdiff (second plot) indicate that the the final QC decisions from GSI and UFO agree with each other.  

gsi_hofx_vs_jedi_hofx_satwind_diag_2021080100_satwind_windEastward
hofxdiff_vs_qcdiff_satwind_diag_2021080100_satwind_windEastward

BrettHoover-NOAA:
@emilyhcliu Thanks for running this down! I ran your UFO diag/geovals files through my simulated QC code and I agree that the UFO and GSI are in nearly complete agreement. There are 8 satwinds that pass GSI but fail UFO QC, with all 8 producing a `SatWindsErrnormCheck` value of almost 0.9, so these are likely float-precision differences just like the 6 satwinds that differed from the prior benchmark. I wanted to make sure I hadn't accidentally missed a pre-QC check with the satwind YAML file that surfaced here. Looks good to me, I'm fine with passing review. Thanks again!

BrettHoover-NOAA (approved):
Looks good, thank you @emilyhcliu!

emilyhcliu:
> Looks good, thank you @emilyhcliu!

@BrettHoover-NOAA Thanks for looking into this.
Just to confirm.  You agree the change in benchmark for satwind?  (Scatwind seems fine, only 6 points more passed)
Can I convert this PR back to active?

BrettHoover-NOAA:
Yeah, the new benchmark for satwind looks good, the 16 GSI/UFO differences
are accounted for. I’m good with going active here.

emilyhcliu:
@BrettHoover-NOAA gave thumbs up. This PR is good to be continued.  The plots of Hofx comparison and the HofxDiff vs qcDiff are [here](https://github.com/NOAA-EMC/GDASApp/pull/631#issuecomment-1724303385).
Converting this draft back to active PR.  
Please continue your review.
Thanks.

CoryMartin-NOAA (approved):

PR #632, opened 2023-09-17
Add IODA2BUFR processing for scatwind

emilyhcliu:
Two updates:

- Add IODA2BUFR processing for scatwind
- Tidy up the satwind python script

### DIFF
--- a/ush/ioda/bufr2ioda/bufr2ioda_satwind_amv_goes.py
+++ b/ush/ioda/bufr2ioda/bufr2ioda_satwind_amv_goes.py
@@ -464,6 +464,7 @@ def bufr_to_ioda(config, logger):
     logger.info("All Done!")
     logger.info(f'Total number of observation processed : {total_ob_processed}')
 
+
 if __name__ == '__main__':
 
     start_time = time.time()
--- a/ush/ioda/bufr2ioda/bufr2ioda_satwind_amv_goes.py
+++ b/ush/ioda/bufr2ioda/bufr2ioda_satwind_amv_goes.py
@@ -98,10 +98,10 @@ def bufr_to_ioda(config, logger):
     sensor_description = '16 channels, balaned visible, near IR, short-wave IR, mid-wave IR, and thermal IR; \
                          central wavelentgh ranges from 470 nm to 13.3 micron'
 
-    logger.info(f"sensor_name = {sensor_name}")
-    logger.info(f"sensor_full_name = {sensor_full_name}")
-    logger.info(f"sensor_id = {sensor_id}")
-    logger.info(f"reference_time = {reference_time}")
+    logger.info(f'sensor_name = {sensor_name}')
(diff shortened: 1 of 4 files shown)


ush/ioda/bufr2ioda/bufr2ioda_satwind_amv_goes.py
```diff
@@ -305,7 +258,7 @@ def bufr_to_ioda(config, logger):
                 satellite_id = satellite_info["satellite_id"]
                 satellite_name = satellite_info["satellite_name"]
                 satinst = sensor_name.lower()+'_'+satellite_name.lower()
-                logger.debug("Split data for {satinst} satid = {sat}")
+                logger.debug('Split data for {satinst} satid = {sat}')
```
CoryMartin-NOAA: this is missing a `f` so it will print literally:
`Split data for {satinst} satid = {sat}`
and not the value in the loop

`logger.debug(f'Split data for {satinst} satid = {sat}')`

ush/ioda/bufr2ioda/bufr2ioda_satwind_scat.py
```diff
@@ -0,0 +1,369 @@
...
+# ==============================================================================================
+# Subset    |  Description                                              |  PrepBUFR Report Type
+# ----------------------------------------------------------------------------------------------
+# NC012122  |  Level-2 ocean surface wind vector retrievals and         |  290
+#           |  derived wind components from the Advanced Scatterometer  |
+#           |  (ASCAT) on MetOp satellites at 50-km sampling resolution |
+# ==============================================================================================
+
+
+def Compute_WindComponents_from_WindDirection_and_WindSpeed(wdir, wspd):
```
CoryMartin-NOAA, line 25: No need to do now, but we should pull this function out somewhere since it is duplicated/shared among wind observations.

CoryMartin-NOAA (commented):
Very minor comment, otherwise looks good.

ush/ioda/bufr2ioda/bufr2ioda_satwind_scat.py
emilyhcliu: @CoryMartin-NOAA Can not agree with you more.  Ron and I had talked about this.  We should have a common Python library for common shared functions.

emilyhcliu: We are not sure where is the best place for this common shared library for Python functions. 
Any thoughts?

CoryMartin-NOAA: Depending on how general this would be, it could be its own python package/stand alone repository. Originally, this was one of the concepts of EMCPy, but now, I'm not sure. Perhaps it still has its place there?

CoryMartin-NOAA (approved):

PR #581, opened 2023-08-24
Remove run_jedi_exe and cleanup legacy YAML tools

CoryMartin-NOAA:
We need to switch to using the `wxflow` jinja-based YAML templating tools everywhere. This PR removes all (that I can see) of the old YAML tool usage, as well as the `run_jedi_exe` script.

PR #633, opened 2023-09-18
Fix YAML generation

CoryMartin-NOAA:
@ADCollard pointed out that whatever I put in #581 doesn't really work. This hopefully at least lets people use the UFO ctest script again.


ADCollard (approved):

ADCollard:
@CoryMartin-NOAA  Thank you!

PR #634, opened 2023-09-20
Cleanup of the of the ioda converter 

guillaumevernieres:
I simplified what the sprinter will have to do next week. There should only be one function to implement now.


utils/obsproc/NetCDFToIodaConverter.h
```diff
@@ -14,46 +14,106 @@
...
 
 namespace gdasapp {
+  // A simple data structure to organize the info to provide to the ioda
+  // writter
+  struct IodaVars {
+    int location;
+    int nVars;
+    Eigen::ArrayXf* obsVal;
+    Eigen::ArrayXf* obsError;
+    Eigen::ArrayXi* preQc;
```
rmclaren: There is no need for these to be pointers. Just make them "by value". Example: `Eigen::ArrayXf obsVal`. 

Just FYI: The usage of this struct (as currently written) will cause a memory leak of obsVal, obsError, preQc as you have not defined a destructor. The better way to handle this, as stated previously, is just to make these by value references. The Eigen frmaework will manage the memory for you. If you ever do get into a situation where you need to manage your own data buffers you should use std::shared_ptr<> so you don't need to worry about manual memory management.

utils/obsproc/NetCDFToIodaConverter.h
```diff
@@ -14,46 +14,106 @@
...
+      // time window info
+      std::string winbegin;
+      std::string winend;
+      std::string obsvar;
+      fullConfig.get("window begin", winbegin);
+      fullConfig.get("window end", winend);
+      fullConfig.get("variable", obsvar);
+      this->windowBegin_ = util::DateTime(winbegin);
+      this->windowEnd_ = util::DateTime(winend);
+      this->variable_ = obsvar;
```
rmclaren: usage of `this->` is not neccessary. Just do `windowBegin_ = util::DateTime(winbegin);`

utils/obsproc/NetCDFToIodaConverter.h
```diff
@@ -14,46 +14,106 @@
...
+      oops::Log::info() << "--- Window begin: " << winbegin << std::endl;
+      oops::Log::info() << "--- Window end: " << winend << std::endl;
+      oops::Log::info() << "--- Variable: " << obsvar << std::endl;
+
+      // get input netcdf files
+      fullConfig.get("input files", this->inputFilenames_);
```
rmclaren: this

utils/obsproc/NetCDFToIodaConverter.h
```diff
@@ -14,46 +14,106 @@
...
+      oops::Log::info() << "--- Input files: " << this->inputFilenames_ << std::endl;
+
+      // ioda output file name
+      this->outputFilename_ = fullConfig.getString("output file");
```
rmclaren: this

utils/obsproc/NetCDFToIodaConverter.h
```diff
@@ -14,46 +14,106 @@
...
       ioda::Group group =
         ioda::Engines::HH::createFile(
                                       this->outputFilename_,
                                       ioda::Engines::BackendCreateModes::Truncate_If_Exists);
-      this->ReadFromNetCDF(group);
+
+      // Extract ioda variables from the provider's files
+      std::string fileName = this->inputFilenames_[0];  // TODO(Guillaume): make it work for a list
+      gdasapp::IodaVars iodaVars;
+      this->ProviderToIodaVars(fileName, iodaVars);
```
rmclaren: The name of the method `ProviderToIodaVars` breaks the naming conventions for a method (camel case starting in lower case letter). Should be `providerToIodaVars`

utils/obsproc/NetCDFToIodaConverter.h
```diff
@@ -14,46 +14,106 @@
...
+      fullConfig.get("input files", this->inputFilenames_);
+      oops::Log::info() << "--- Input files: " << this->inputFilenames_ << std::endl;
+
+      // ioda output file name
+      this->outputFilename_ = fullConfig.getString("output file");
+      oops::Log::info() << "--- Output files: " << this->outputFilename_ << std::endl;
+    }
+
+    // Method to write out a IODA file (writter called in destructor)
     void WriteToIoda() {
```
rmclaren: methods should start in lower case letter(?)

utils/obsproc/NetCDFToIodaConverter.h
```diff
@@ -14,46 +14,106 @@
...
+      adtIodaObsErr.writeWithEigenRegular(*iodaVars.obsError);
     }
 
    private:
-    // Virtual method to read a netcdf file and store the relevant info in a group
-    virtual void ReadFromNetCDF(ioda::Group &) = 0;
-
+    // Virtual method that reads the provider's netcdf file and store the relevant
+    // info in a IodaVars struct
+    virtual void ProviderToIodaVars(std::string fileName, gdasapp::IodaVars & iodaVars) = 0;
```
rmclaren: should method start with lower case letter?

utils/obsproc/NetCDFToIodaConverter.h
```diff
@@ -14,46 +14,106 @@
...
+
+      // Create the IODA variables
+      ioda::Variable adtIodaDatetime =
+        ogrp.vars.createWithScales<float>("Metadata/dateTime",
+                                          {ogrp.vars["Location"]}, float_params);
+      // TODO(Mindo): Get the date info from the netcdf file
+      adtIodaDatetime.atts.add<std::string>("units", {"seconds since 9999-04-15T12:00:00Z"}, {1});
+
+      ioda::Variable adtIodaObsVal =
+        ogrp.vars.createWithScales<float>("ObsValue/"+this->variable_,
```
rmclaren: this

utils/obsproc/NetCDFToIodaConverter.h
```diff
@@ -14,46 +14,106 @@
...
+                                          {ogrp.vars["Location"]}, float_params);
+      ioda::Variable adtIodaObsErr =
+        ogrp.vars.createWithScales<float>("ObsError/"+this->variable_,
```
rmclaren: this

utils/obsproc/Rads2Ioda.h
```diff
@@ -13,75 +13,49 @@
...
-    oops::Log::info() << "Window begin: " << this->windowBegin_ << std::endl;
-    oops::Log::info() << "Window end: " << this->windowEnd_ << std::endl;
-    oops::Log::info() << "IODA output file name: " << this->outputFilename_ << std::endl;
     }
 
-    // Read netcdf file and populate group
-    void ReadFromNetCDF(ioda::Group & group) override {
-      std::string fileName = this->inputFilenames_[0];  // TODO(Guillaume): make it work for a list
+    // Read netcdf file and populate iodaVars
+    void ProviderToIodaVars(std::string fileName, gdasapp::IodaVars & iodaVars) override {
```
rmclaren: could be `final` instead of `override`

rmclaren: Should pass `const std::string& filename` in order to avoid copies.

rmclaren: namoing convention...

utils/obsproc/NetCDFToIodaConverter.h
```diff
@@ -14,46 +14,106 @@
...
+      // Set up the float creation parameters
+      ioda::VariableCreationParameters float_params;
+      float_params.chunk = true;               // allow chunking
+      float_params.compressWithGZIP();         // compress using gzip
+      float missing_value = util::missingValue(missing_value);
+      float_params.setFillValue<float>(missing_value);
+
+      // Create the IODA variables
+      ioda::Variable adtIodaDatetime =
+        ogrp.vars.createWithScales<float>("Metadata/dateTime",
```
apchoiCMD, line 88: Just wonder that is it going to be "float" instead of "int or int64"?

utils/obsproc/Rads2Ioda.h
```diff
@@ -13,75 +13,49 @@
...
       // Get dimensions
-      int location = ncFile.getDim("time").getSize();
-      int nVars = 1;
+      iodaVars.location = ncFile.getDim("time").getSize();
+      oops::Log::debug() << "--- iodaVars.location: " << iodaVars.location << std::endl;
+
+      // Allocate memory
+      iodaVars.obsVal = new Eigen::ArrayXf(iodaVars.location);
+      iodaVars.obsError = new Eigen::ArrayXf(iodaVars.location);
+      iodaVars.preQc = new Eigen::ArrayXi(iodaVars.location);
```
rmclaren: No need for `new` if vars are value refs.

utils/obsproc/Rads2Ioda.h
```diff
@@ -13,75 +13,49 @@
...
       }
-      adtIodaObsVal.writeWithEigenRegular(tmpvar);
 
-      // Write adt obs error to group
-      for (int i = 0; i <= location; i++) {
-        tmpvar[i] = 0.1;
+      // Do something for obs error
+      for (int i = 0; i <= iodaVars.location; i++) {
+        (*iodaVars.obsError)(i) = 0.1;
       }
```
apchoiCMD, line 58: Do we do something for PreQC?

apchoiCMD (dismissed):

guillaumevernieres:
It's still an incomplete example that doesn't include any of your changes yet @apchoiCMD .

apchoiCMD:
> It's still an incomplete example that doesn't include any of your changes yet @apchoiCMD .

Yep, I can imagine, thanks for the reply!

rmclaren (changes requested):

utils/obsproc/Rads2Ioda.h
guillaumevernieres: Thanks for your review @rmclaren . I'll change the code following your suggestions. `providerToIodaVars` has to be implemented in the derived classes, if I use `final`, that would not be possible. Or am I missing something?

apchoiCMD (dismissed):

utils/obsproc/NetCDFToIodaConverter.h
guillaumevernieres: next PR @apchoiCMD

utils/obsproc/Rads2Ioda.h
rmclaren: The `final` keyword goes in your Rads2Ioda subclass (not the base class). This will prevent anyone from creating a subclass of Rads2Ioda and overriding that method. Its probably not very likely that anyone would do this, but anyways... In C++ you usually default to the highest level of protection you can...

guillaumevernieres: Ha, yes, I was clearly missing something.

rmclaren (approved):

rmclaren (approved):

utils/obsproc/Rads2Ioda.h
guillaumevernieres: :+1:

utils/obsproc/Rads2Ioda.h
guillaumevernieres: :+1:

guillaumevernieres: :+1:

utils/obsproc/NetCDFToIodaConverter.h
guillaumevernieres: :+1:

utils/obsproc/NetCDFToIodaConverter.h
guillaumevernieres: :+1:

utils/obsproc/NetCDFToIodaConverter.h
guillaumevernieres: :+1:

apchoiCMD (approved):

utils/obsproc/NetCDFToIodaConverter.h
guillaumevernieres: :+1:

utils/obsproc/NetCDFToIodaConverter.h
guillaumevernieres: :+1:

utils/obsproc/NetCDFToIodaConverter.h
guillaumevernieres: :+1:

PR #638, opened 2023-09-25
Pass a list of files to the marine obs converter

guillaumevernieres:
## Description
This PR adds the possibility to process more than one file. 
```yaml
provider: RADS
window begin: 2018-04-15T06:00:00Z
window end: 2018-04-15T12:00:00Z
variable: absoluteDynamicTopography
output file: rads_adt_j3_2021182.ioda.nc
input files:
- rads_adt_j3_2021182.nc4
- rads_adt_j3_2021182.nc4
```
The 2 RADS obs files in the ```input files``` list are read in parallel, concatenated and written in the ```output file``` ```rads_adt_j3_2021182.ioda.nc```. 
Currently the code requires as many PE's as there are input files, not great, something that I might change later.


apchoiCMD (dismissed):

ShastriPaturi (dismissed):

AndrewEichmann-NOAA (dismissed):
I presume we'll be testing this the coming week

utils/obsproc/NetCDFToIodaConverter.h
```diff
@@ -58,57 +65,108 @@ namespace gdasapp {
...
+      }
+    }
+
+    // Short-cut to create type dependent VariableCreationParameters
+    template <typename T>
+    ioda::VariableCreationParameters createVariableParams() {
+      ioda::VariableCreationParameters params;
+      params.chunk = true;               // allow chunking
+      params.compressWithGZIP();         // compress using gzip
+      T missingValue = util::missingValue(missingValue);
```
rmclaren: shouldn't this be `auto missingValue = util::missingValue<T>();`

utils/obsproc/NetCDFToIodaConverter.h
```diff
@@ -58,57 +65,108 @@ namespace gdasapp {
...
+      params.setFillValue<T>(missingValue);
```
rmclaren: Since yoy only use missingValue here I would just do `params.setFillValue<T>(util:missingValue<T>());`. There is no need for the temporary variable.

utils/obsproc/NetCDFToIodaConverter.h
```diff
@@ -58,57 +65,108 @@ namespace gdasapp {
...
-
-      // Write adt obs value to group
-      adtIodaObsVal.writeWithEigenRegular(iodaVars.obsVal);
-
-      // Write adt obs error to group
-      adtIodaObsErr.writeWithEigenRegular(iodaVars.obsError);
+      int myrank  = comm.rank();
+      int nobs(0);
+      oops::Log::debug() << "ooooooooooooo my rank : " << myrank << comm.size() << std::endl;
+      if (myrank <= inputFilenames_.size()-1) {
```
rmclaren: nitpick: I would add spaces around the `-` operator..

utils/obsproc/NetCDFToIodaConverter.h
```diff
@@ -58,57 +65,108 @@ namespace gdasapp {
...
+
+      // Gather obsVal's
+      gatherObs(comm, iodaVars.obsVal, iodaVarsAll.obsVal);
+      gatherObs(comm, iodaVars.obsError, iodaVarsAll.obsError);
+      gatherObs(comm, iodaVars.preQc, iodaVarsAll.preQc);
+
+      oops::Log::debug() << "--- all nobs: " << iodaVarsAll.obsVal.size() << std::endl;
+      oops::Log::debug() << "--- all obsVal: " << iodaVarsAll.obsVal << std::endl;
+      oops::Log::debug() << "--- all obsError: " << iodaVarsAll.obsError << std::endl;
+      oops::Log::debug() << "--- all preQc: " << iodaVarsAll.preQc << std::endl;
```
rmclaren, line 98: Sorry I'm a little confused here as the gathered values are only used in these debug logs, and not used afterwords... (iodaVars are passed as const and don't change). Is this in purpose?

rmclaren: If you mean for the iodaVars to reflect the gathered data could you just do `oops::mpi::allGatherv(comm, iodaVars.obsVal);` etc...

utils/obsproc/NetCDFToIodaConverter.h
```diff
@@ -58,57 +65,108 @@ namespace gdasapp {
...
+        ioda::Variable adtIodaObsErr =
+          ogrp.vars.createWithScales<float>("ObsError/"+variable_,
+                                            {ogrp.vars["Location"]}, float_params);
+
+        ioda::Variable adtIodaPreQc =
+          ogrp.vars.createWithScales<int>("PreQC/"+variable_,
+                                            {ogrp.vars["Location"]}, int_params);
+
+        // Write adt obs info to group
+        adtIodaObsVal.writeWithEigenRegular(iodaVars.obsVal);
```
rmclaren: Did you mean to write the gathered values?

utils/obsproc/NetCDFToIodaConverter.h
```diff
@@ -58,57 +65,108 @@ namespace gdasapp {
...
+    template <typename T>
+    void gatherObs(const eckit::mpi::Comm & comm,
+                   const Eigen::Array<T, Eigen::Dynamic, 1> & obsPe,
+                   Eigen::Array<T, Eigen::Dynamic, 1> & obsAllPes) {
+      std::vector<T> tmpVec(obsPe.data(), obsPe.data() + obsPe.size());
+      oops::mpi::allGatherv(comm, tmpVec);
+      for (int i = 0; i < tmpVec.size(); ++i) {
+        obsAllPes(i) = tmpVec[i];
+      }
+    }
```
rmclaren, line 156: Is this method needed? See previous comments...

utils/obsproc/NetCDFToIodaConverter.h
guillaumevernieres: oops ... yes!

utils/obsproc/NetCDFToIodaConverter.h
guillaumevernieres: > Sorry I'm a little confused here as the gathered values are only used in these debug logs, and not used afterwords... (iodaVars are passed as const and don't change). Is this in purpose?

That's a bug, obs are gathered in iodavarsall and that's what should have been used to save.

rmclaren, line 98: Could this code go inside the `if` statement. Do you really need to gather in all the MPI tasks or just the one that writes the file (should this only be done in the MPI rank 0 task)?

guillaumevernieres: > If you mean for the iodaVars to reflect the gathered data could you just do `oops::mpi::allGatherv(comm, iodaVars.obsVal);` etc...

I did mean to not do the gather in place,but I don't remember why ...

utils/obsproc/NetCDFToIodaConverter.h
guillaumevernieres: oops::mpi::allGatherv isn't overloaded for input Eigen::Array. This is my short cut to not repeat that chunk of code, short of adding an option in oops::mpi::allGatherv. Is there a better way to do this @rmclaren ?

utils/obsproc/NetCDFToIodaConverter.h
guillaumevernieres: We're going to introduce a thinning step before the writing to ioda. This is probably not the most memory efficient way to do this, but I'm planning to have iodaVarAll available from all pe's ... this might change though ...

utils/obsproc/NetCDFToIodaConverter.h
rmclaren: My MPI is a little rusty (its been a little while), so please forgive me if I say something stupid (I'm not used to thinking this way). It's coming back to me slowly so please bare with me :)

Having said that I don't think this code will work the way you think it does. For one thing the tmpVec is too small if there are more than one MPI tasks reading files (should be size of sum of nobs you computed in `reduce` i think)... I think there is also a lot of unecessary copying, and since the offset always starts at 0 (`int i = 0; `) you are overwriting data (only ending up with the data associated with the very last MPI tasks file). I don't think this method is really necessary at all...

Here is my suggestion (but please keep in mind that my MPI is rusty...).

```cxx
      // Get the total number of obs across pe's
      comm.allReduce(nobs, nobs, eckit::mpi::sum());
      oops::Log::debug() << " my rank : " << myrank
                         << " Num pe's: " << comm.size()
                         << " nobs: " << nobs << std::endl;

      gdasapp::IodaVars iodaVarsAll(nobs);

       // Gather obsVal's
      oops::mpi::allGatherv(comm, iodaVar.obsVal, iodaVarsAll.obsVal);
      oops::mpi::allGatherv(comm, iodaVars.obsVal, iodaVarsAll.obsError);
      oops::mpi::allGatherv(comm, iodaVarsAll.obsVal, iodaVarsAll.obs.preQc);

      oops::Log::debug() << "--- all nobs: " << iodaVarsAll.obsVal.size() << std::endl;
      oops::Log::debug() << "--- all obsVal: " << iodaVarsAll.obsVal << std::endl;
      oops::Log::debug() << "--- all obsError: " << iodaVarsAll.obsError << std::endl;
      oops::Log::debug() << "--- all preQc: " << iodaVarsAll.preQc << std::endl;
   
      // Create empty group backed by HDF file
      if (oops::mpi::world().rank() == 0)
      {
        ioda::Group group =
          ioda::Engines::HH::createFile(
                                        outputFilename_,
                                        ioda::Engines::BackendCreateModes::Truncate_If_Exists);

// ***** end of suggestion ******//
```

apchoiCMD (approved):

utils/obsproc/NetCDFToIodaConverter.h
rmclaren: hold on a second

guillaumevernieres: I think what I did is correct @rmclaren , but maybe not amazingly efficient. It does properly concatenate the input files.
The oops mpi call is here:
https://github.com/JCSDA-internal/oops/blob/c03558814bb8cb81de71265b98f7e54ef0d953f9/src/oops/mpi/mpi.h#L219

The size of x is changed in that function.

What I might change in the future is to all gather "in place" instead of creating a new ```iodaVarsAll```.

guillaumevernieres: the size of ```tmpVec``` changes in line 152 @rmclaren .

rmclaren: Ok, I looked at the oops implementation for allGatherv in oops and its actually swapping out the data buffer under the covers via a std::move...  I think what you had originally works as written, but could be improved.

I think you might not need `iodaVarsAll` at all. This means you don't need to do the `allReduce`. So you could just do the gather and then write the iodaVar values.


```cxx

       // Gather obsVal's
      oops::mpi::allGatherv(comm, iodaVar.obsVal);
      oops::mpi::allGatherv(comm, iodaVars.obsError);
      oops::mpi::allGatherv(comm, iodaVarsAll.preQc);

      oops::Log::debug() << "--- all nobs: " << iodaVars.obsVal.size() << std::endl;
      oops::Log::debug() << "--- all obsVal: " << iodaVars.obsVal << std::endl;
      oops::Log::debug() << "--- all obsError: " << iodaVars.obsError << std::endl;
      oops::Log::debug() << "--- all preQc: " << iodaVars.preQc << std::endl;
   
      // Create empty group backed by HDF file
      if (oops::mpi::world().rank() == 0)
      {
        ioda::Group group =
          ioda::Engines::HH::createFile(
                                        outputFilename_,
                                        ioda::Engines::BackendCreateModes::Truncate_If_Exists);

// ***** a bunch of code *****

        // Write adt obs info to group
        adtIodaObsVal.writeWithEigenRegular(iodaVars.obsVal);
        adtIodaObsErr.writeWithEigenRegular(iodaVars.obsError);
        adtIodaPreQc.writeWithEigenRegular(iodaVars.preQc);
```

rmclaren: Just a side note, instead of manually copying the data in `gatherObs` (in the for loop) you could have just done a std::move and "steal" the pointer.

Sorry for my confusions...

utils/obsproc/NetCDFToIodaConverter.h
rmclaren: I see...

rmclaren (approved):
Suggested changes are up to you.

utils/obsproc/NetCDFToIodaConverter.h
guillaumevernieres: ~Ha! ok, will do.~
I apparently can't do this with and Eigen Array @rmclaren

PR #647, opened 2023-09-27
Marine iodaconv: Add metadata to the ioda writer

guillaumevernieres:
## Description
This feature branch adds the option to pass integer and float metadata to the ioda writer. 

### Issues not addressed
- @AndrewEichmann-NOAA : The cdl to nc4 for sss doesn't seem to be working. I commented it out n this PR. 
- I have yet to implement an "allgather" for the optional metadata, so the concatenation is currently not correct
- While testing on my laptop yesterday, I noticed that the application freezes when trying to concatenate more than 4 files. No idea if this is a hardware or code issue. A problem for today.

### DIFF
--- a/utils/obsproc/Rads2Ioda.h
+++ b/utils/obsproc/Rads2Ioda.h
@@ -41,43 +41,68 @@ namespace gdasapp {
       gdasapp::IodaVars iodaVars(nobs, floatMetadataNames, intMetadataNames);
 
       // Read non-optional metadata: datetime, longitude and latitude
-      netCDF::NcVar latNcVar = ncFile.getVar("lat");
-      int latVal[iodaVars.location];  // NOLINT (can't pass vector to getVar below)
-      latNcVar.getVar(latVal);
+      int lat[iodaVars.location];  // NOLINT
+      ncFile.getVar("lat").getVar(lat);
 
-      netCDF::NcVar lonNcVar = ncFile.getVar("lon");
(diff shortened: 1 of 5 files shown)


apchoiCMD (approved):

utils/obsproc/Rads2Ioda.h
```diff
@@ -18,44 +20,93 @@ namespace gdasapp {
...
+        iodaVars.obsError(i) = 0.1;  // Do something for obs error
+        iodaVars.preQc(i) = 0;
+        // Save MDT in optional floatMetadata
+        iodaVars.floatMetadata(i, 0) = iodaVars.obsVal(i) - static_cast<float>(sla[i])*scaleFactor;
       }
+      return iodaVars;
     };
+    int altimeterMissions(std::string missionName) {
+      std::map<std::string, int> altimeterMap;
+      // TODO(All): This is incomplete, add missions to the list below
```
ShastriPaturi, line 100: Missing Sentinel 6A. We can add it here since these data are available in the data tanks.
mission_name = "SNTNL-6A"

ShastriPaturi (commented):
You can incorporate the changes I suggested.

utils/obsproc/Rads2Ioda.h
guillaumevernieres: a problem for an other day @ShastriPaturi .

apchoiCMD (approved):

AndrewEichmann-NOAA (approved):
looks reasonable enough, will find out soon enough if it's not

Issue #580, opened 2023-08-23
Netcdf to ioda


guillaumevernieres:
## Description
The python ioda converters are too slow: Replace with a c++ implementation instead. 

### Converters to implement
- [x] #607
- [x] #608 (@guillaumevernieres @apchoiCMD)
- [x] #609 (@ShastriPaturi )
- [x] #610 (@apchoiCMD)
- [x] #613 (@AndrewEichmann-NOAA )
- [x] #614 (@AndrewEichmann-NOAA )

### Check the conversion
- [x] #665 (@apchoiCMD)
- [x] #677 (@apchoiCMD)

ShastriPaturi:
Input files for the "Marine Obsprep code sprint" are copied to orion here:
`/work2/noaa/stmp/spaturi/forAndrew`

guillaumevernieres:
I think I over-engineered the concatenation ... Oops. I just implemented the serial version of what we now do with mpi gather and it's a lot faster on my laptop, and it doesn't randomly hang. 
I'm testing with the RADS files, which are a lot smaller than anything else that we have to convert:
Reading in ~50 files with a total obs count of ~3M and writing the ioda file took 2.5s ...

guillaumevernieres:
I'll wait until @ShastriPaturi is done with his converter before removing the gather. We have 36 files to process per 6 hour window, each file has 162M obs ...

ShastriPaturi:
@guillaumevernieres,
`There's more, @ShastriPaturi , can you fill in the rest please?`
I do not think there are anymore netcdf files that we will consider, for now.

PR #653, opened 2023-10-02
Add SMOS SSS to IODA converter with test data

### LINKED ISSUE #613: SMOS

AndrewEichmann-NOAA:
Partially addresses https://github.com/NOAA-EMC/GDASApp/issues/580 by resolving https://github.com/NOAA-EMC/GDASApp/issues/613

Subsampled SMOS SSS data in cdl files had to be modified by converting (with vi) ushort to short to permit conversion to binary NetCDF file - otherwise ushort is not recognized in certain NetCDF environments. The difference seems to be ignored in the conversion to IODA

fixes https://github.com/NOAA-EMC/GDASApp/issues/613


utils/obsproc/Smos2Ioda.h
```diff
@@ -0,0 +1,83 @@
...
+      // unix epoch at Jan 01 2000 00:00:00 GMT+0000
+      const int mjd2000 = 946684800;
+
+      for (int i = 0; i < iodaVars.location; i++) {
+        iodaVars.longitude(i) = lon[i];
+        iodaVars.latitude(i) = lat[i];
+        iodaVars.obsVal(i) = sss[i];
+        iodaVars.obsError(i) = sss_error[i];  // Do something for obs error
+        iodaVars.preQc(i) = sss_qc[i];
+        iodaVars.datetime(i) =  static_cast<int64_t>(datetime[i]*86400.0f) + mjd2000;
```
guillaumevernieres, line 82: looks like you're missing the reference date

utils/test/testinput/gdas_smos2ioda.yaml
```diff
@@ -0,0 +1,8 @@
+provider: SMOS
+window begin: 2018-04-15T06:00:00Z
+window end: 2018-04-15T12:00:00Z
+variable: Salinity
```
guillaumevernieres: variable is not needed anymore

guillaumevernieres (changes requested):
Just a few minor comments @AndrewEichmann-NOAA .
Could you point to the correct issue in your PR description, we should have one for smos.

utils/obsproc/Smos2Ioda.h
AndrewEichmann-NOAA: Not sure what you mean here

utils/test/testinput/gdas_smos2ioda.yaml
AndrewEichmann-NOAA: removed

utils/obsproc/Smos2Ioda.h
AndrewEichmann-NOAA: The datetime here is the unix epoch - I understand that to be the IODA convention, does it need to be explicitly stated?

apchoiCMD: I think you need to add something like this 
"iodaVars.referenceDate = "seconds since 1858-11-17T00:00:00Z";"

AndrewEichmann-NOAA: Oooooh got it

utils/obsproc/Smos2Ioda.h
```diff
@@ -0,0 +1,83 @@
...
+      ncFile.getVar("Longitude").getVar(lon);
+
+      float sss[iodaVars.location];  // NOLINT
+      ncFile.getVar("SSS_corr").getVar(sss);
+
+      float sss_error[iodaVars.location];  // NOLINT
+      ncFile.getVar("Sigma_SSS_corr").getVar(sss_error);
+
+      unsigned short sss_qc[iodaVars.location];  // NOLINT
+      ncFile.getVar("Dg_quality_SSS_corr").getVar(sss_qc);
```
rmclaren: Be careful not to blow the **stack**.... Though the exact of **stack** space reserved for a thread can vary depending on compiler configuration and OS, this number can be quiet low and its not very hard to exceed (for example MS Windows only reserves like 1MB of stack space per thread by default). To avoid this its usual to do larger allocations on the **heap**.

Note: if you use standard c++ containers like std::vector, this will happend automatically...

utils/obsproc/Smos2Ioda.h
rmclaren: Could you avoid some of the element by element copying of values by using the Eigen::Map function (https://eigen.tuxfamily.org/dox/group__TutorialMapClass.html)?

utils/obsproc/Smos2Ioda.h
AndrewEichmann-NOAA: @guillaumevernieres Is there some reason not to do this?

guillaumevernieres: You can't pass an std::vector to getVar @AndrewEichmann-NOAA . Maybe we should have been using ```netcdfcpp.h```? 
Wouldn't ```ulimit -s unlimited``` fix our possible stack size issue @rmclaren ?

utils/obsproc/Smos2Ioda.h
```diff
@@ -0,0 +1,86 @@
...
+      int nobs = ncFile.getDim("n_grid_points").getSize();
+
+      // Set the int metadata names
+      // TODO(AFE): add other metadata in form of
+      // std::vector<std::string> intMetadataNames = {"pass", "cycle", "mission"};
+      std::vector<std::string> intMetadataNames = {};
+
+      // Set the float metadata name
+      // TODO(AFE): add other metadata in form of
+      // std::vector<std::string> floatMetadataNames = {"mdt"};
```
ShastriPaturi, line 43: Remove commented lines

utils/obsproc/Smos2Ioda.h
```diff
@@ -0,0 +1,86 @@
...
+
+      // Open the NetCDF file in read-only mode
+      netCDF::NcFile ncFile(fileName, netCDF::NcFile::read);
+
+      // Get number of obs
+      int nobs = ncFile.getDim("n_grid_points").getSize();
+
+      // Set the int metadata names
+      // TODO(AFE): add other metadata in form of
+      // std::vector<std::string> intMetadataNames = {"pass", "cycle", "mission"};
```
ShastriPaturi, line 38: same as below

ShastriPaturi (changes requested):

utils/obsproc/Smos2Ioda.h
guillaumevernieres: both are part  of the TODO's @ShastriPaturi , seems fine to me as a comment, more so since we know it's work in progress and the next feature will be the addition of metadata.

ShastriPaturi (dismissed):

apchoiCMD (dismissed):

utils/obsproc/Smos2Ioda.h
rmclaren: @guillaumevernieres That will make the program run, yes. It does remove some safety's (for example early detection of infinite recursion). Like I said its generally better to avoid huge stack allocations in the first place, by allocating memory on the heap instead (thats what its for)... 

I believe you could use`std::vector` as follows:

```
      std::vector<float> lat(iodaVars.location);
      ncFile.getVar("Latitude").getVar(lat.data());
```

std::vector<>::data() gives you the pointer to the internal buffer.

guillaumevernieres: Thanks for the pointer (no pun intended :) ) @rmclaren .

AndrewEichmann-NOAA: Done

utils/obsproc/Smos2Ioda.h
AndrewEichmann-NOAA: The Eigen Maps clash a bit with the NetCDF, so I made a todo out of it

apchoiCMD (approved):

guillaumevernieres (approved):
Thanks @AndrewEichmann-NOAA !

PR #655, opened 2023-10-03
Using already created communicator ... and more

guillaumevernieres:
### A few bux fixes:

- It looks like redefining the communicator as `oops::mpi::world()` was an issues. The comm is now using the one defined in `Application` and was added as a data member of `NetCDFToIodaConverter`. 
- Pointing to eckit's gather instead of oops'
- A bit of code tidying

This worked with the 5 RADS files that I have, which is a change from "not working", but the obs count is quite low at ~200,000. We'll declare victory when we successfully test with the other observations.


apchoiCMD (dismissed):

AndrewEichmann-NOAA:
@guillaumevernieres So this means we should be able to test with multiple input files?

guillaumevernieres:
> @guillaumevernieres So this means we should be able to test with multiple input files?

@AndrewEichmann-NOAA , yes, but my tests started hanging in the gather again last night ... Same code. I added a few barriers and it started going through again, but that doesn't make any sense. I thought I found the bug, but looks like not! 
I'll push my last commit but my next step is to workout a more robust serial implementation next.

apchoiCMD (approved):

PR #658, opened 2023-10-04
Addition of an append method

### LINKED ISSUE #659: Add the possibility to concatenate on 1 PE
The converter can currently only concatenate using MPI gather. We need to be able to concatenate even if running on 1 PE only.

guillaumevernieres:
### Description 
This new feature will allow concatenating multiple files on 1 pe only.
- fixes #659 

#### Work done 
- Addition of an `append` method in the `IodaVars` structure

#### How was this tested
- I commented out the 2nd smos file in @AndrewEichmann-NOAA 's SMOS ctest to cover the new code. 
- Tested with ~50 SMOS files, concatenating a total of ~6M obs in a few seconds

#### Note
I'm tempted to :bomb: the mpi implementation but I'll wait for @ShastriPaturi 's `GHRSST` converter before I do this.

### DIFF
--- a/utils/obsproc/NetCDFToIodaConverter.h
+++ b/utils/obsproc/NetCDFToIodaConverter.h
@@ -128,7 +128,7 @@ namespace gdasapp {
       int myrank  = comm_.rank();
       int nobs(0);
 
-      // Currently need 1 PE per file, abort if not the case
+      // Currently need the PE count to be less than the number of files
       ASSERT(comm_.size() <= inputFilenames_.size());
 
       // Read the provider's netcdf file
--- a/utils/obsproc/NetCDFToIodaConverter.h
+++ b/utils/obsproc/NetCDFToIodaConverter.h
@@ -43,6 +43,7 @@ namespace gdasapp {
     Eigen::ArrayXXf intMetadata;                  // Optional array of integer metadata
     std::vector<std::string> intMetadataName;    // String descriptor of the integer metadata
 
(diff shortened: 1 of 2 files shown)


apchoiCMD (approved):

apchoiCMD (approved):

utils/obsproc/NetCDFToIodaConverter.h
```diff
@@ -55,7 +56,43 @@ namespace gdasapp {
...
+    }
+
+    // Append an other instance of IodaVars
+    void append(const IodaVars& other) {
+      // Check if the two instances can be concatenated
+      ASSERT(nVars == other.nVars);
+      ASSERT(nfMetadata == other.nfMetadata);
+      ASSERT(niMetadata == other.niMetadata);
+      ASSERT(floatMetadataName == floatMetadataName);
+      ASSERT(intMetadataName == intMetadataName);
```
AndrewEichmann-NOAA, line 70: What's going on here? It looks like you're making sure the two files are identical - is it just checking for identical structures?

guillaumevernieres: @AndrewEichmann-NOAA I'm making sure the objects can be merged. The files can be different but the number and types of metadata need to be the same.

AndrewEichmann-NOAA (approved):

ShastriPaturi (approved):

PR #662, opened 2023-10-05
Updated GHRSST SST test files

ShastriPaturi:
As the title says


guillaumevernieres (approved):
Thanks @ShastriPaturi !

Issue #667, opened 2023-10-06
Implement masking method in IODA converter


AndrewEichmann-NOAA:
This should probably be in here: https://github.com/NOAA-EMC/GDASApp/issues/580

PR #668, opened 2023-10-06
Implements mask method for IodaVars

AndrewEichmann-NOAA:
Mask method for IodaVars, takes boolean Eigen array mask and applies it to variables in current IodaVars instance. 

Tested on ctest data and real data files with the following mask, before method returns iodaVars:
```
      Eigen::Array<bool, Eigen::Dynamic, 1> mask = (iodaVars.obsVal > 0.0);
      iodaVars.trim(mask);

```
Not tested with metadata variables, so better do that.

Addresses https://github.com/NOAA-EMC/GDASApp/issues/667

### DIFF
--- a/utils/obsproc/NetCDFToIodaConverter.h
+++ b/utils/obsproc/NetCDFToIodaConverter.h
@@ -96,47 +96,39 @@ namespace gdasapp {
     void trim(const Eigen::Array<bool, Eigen::Dynamic, 1>& mask ) {
       int newlocation = mask.count();
 
-      Eigen::ArrayXf longitudeMasked(newlocation);
-      Eigen::ArrayXf latitudeMasked(newlocation);
-      Eigen::Array<int64_t, Eigen::Dynamic, 1> datetimeMasked(newlocation);
-      Eigen::ArrayXf obsValMasked(newlocation);
-      Eigen::ArrayXf obsErrorMasked(newlocation);
-      Eigen::ArrayXi preQcMasked(newlocation);
-
-      Eigen::ArrayXXf floatMetadataMasked(newlocation, nfMetadata);
-      Eigen::ArrayXXf intMetadataMasked(newlocation, niMetadata);
+      IodaVars iodaVarsMasked(newlocation,  this->floatMetadataName, this->intMetadataName);
 
       // this feels downright crude, but apparently numpy-type masks are on the todo list for Eigen
       int j = 0;
-      for (int i = 0; i < location; i++) {
+      for (int i = 0; i < this->location; i++) {
         if (mask(i)) {
-          longitudeMasked(j) = longitude(i);
-          latitudeMasked(j) = latitude(i);
-          obsValMasked(j) = obsVal(i);
-          obsErrorMasked(j) = obsError(i);
-          preQcMasked(j) = preQc(i);
-          datetimeMasked(j) = datetime(i);
-          for (int k = 0; k < nfMetadata; k++) {
-            intMetadataMasked(j, k) = intMetadata(i, k);
+          iodaVarsMasked.longitude(j) = this->longitude(i);
+          iodaVarsMasked.latitude(j) = this->latitude(i);
+          iodaVarsMasked.obsVal(j) = this->obsVal(i);
+          iodaVarsMasked.obsError(j) = this->obsError(i);
+          iodaVarsMasked.preQc(j) = this->preQc(i);
+          iodaVarsMasked.datetime(j) = this->datetime(i);
+          for (int k = 0; k < this->nfMetadata; k++) {
+            iodaVarsMasked.intMetadata(j, k) = this->intMetadata(i, k);
             }
-          for (int k = 0; k < niMetadata; k++) {
-            intMetadataMasked(j, k) = intMetadata(i, k);
+          for (int k = 0; k < this->niMetadata; k++) {
+            iodaVarsMasked.intMetadata(j, k) = this->intMetadata(i, k);
             }
         j++;
         }  // end if (mask(i))
       }
 
-      longitude = longitudeMasked;
-      latitude = latitudeMasked;
-      datetime = datetimeMasked;
-      obsVal = obsValMasked;
-      obsError = obsErrorMasked;
-      preQc = preQcMasked;
-      floatMetadata = floatMetadataMasked;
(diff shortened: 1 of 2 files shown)


rmclaren (dismissed):

utils/obsproc/NetCDFToIodaConverter.h
```diff
@@ -93,6 +93,52 @@ namespace gdasapp {
...
+            intMetadataMasked(j, k) = intMetadata(i, k);
+            }
+          for (int k = 0; k < niMetadata; k++) {
+            intMetadataMasked(j, k) = intMetadata(i, k);
+            }
+        j++;
+        }  // end if (mask(i))
+      }
+
+      longitude = longitudeMasked;
```
rmclaren: **nitpick**: should class variables end in underscores to indicate their status (ex: `longitude_`). Do we have a convention for this?

ShastriPaturi (dismissed):
Awesome!!!

apchoiCMD (dismissed):

utils/obsproc/NetCDFToIodaConverter.h
guillaumevernieres: Yes @rmclaren ... my bad. I started doing this for the other class but I forgot to do it for IodaVars ...

utils/obsproc/NetCDFToIodaConverter.h
```diff
@@ -93,6 +93,52 @@ namespace gdasapp {
       location += other.location;
       oops::Log::trace() << "IodaVars::IodaVars done appending." << std::endl;
     }
+    void trim(const Eigen::Array<bool, Eigen::Dynamic, 1>& mask ) {
+      int newlocation = mask.count();
+
+      Eigen::ArrayXf longitudeMasked(newlocation);
```
guillaumevernieres: just use the constructor from above to declare a new instance of IodaVars, something like:
```c++
IodaVars iodaVarsMasked(newlocation,  this->floatMetadataName, this->intMetadataName);
```

guillaumevernieres (changes requested):
Thanks @AndrewEichmann-NOAA for the quick turn around!
Looks good, beside:

1. The declaration of the eigen arrays that can be done in a one-liner. See my comment in the code.
2. You need to try to cover the new code with one of the ctests
~~3. I think the ioda writer will actually write something that has the original size, not the masked size~~

utils/obsproc/NetCDFToIodaConverter.h
```diff
@@ -93,6 +93,52 @@ namespace gdasapp {
...
+
+      longitude = longitudeMasked;
+      latitude = latitudeMasked;
+      datetime = datetimeMasked;
+      obsVal = obsValMasked;
+      obsError = obsErrorMasked;
+      preQc = preQcMasked;
+      floatMetadata = floatMetadataMasked;
+      intMetadata = intMetadataMasked;
+
```
guillaumevernieres, line 129: I'm actually not convinced this is going to work @AndrewEichmann-NOAA . Does the ```=``` above properly resizes the eigen arrays? Did you check the size of the iodafile that was created?

rmclaren: @guillaumevernieres I believe this will work fine. `=` will replace the existing arrays with the new ones which are of the masked size. I'm not sure it would make sense to explicitly do a std::move... I'm guessing the compiler will do that for you anyways when you do an optimized build...

utils/obsproc/NetCDFToIodaConverter.h
AndrewEichmann-NOAA: So I presume this needs to be changed in IodaVars and any currently merged converters, all at once? I.e., not in this PR?

utils/obsproc/NetCDFToIodaConverter.h
AndrewEichmann-NOAA: My understanding is that, as @rmclaren suggests, the existing arrays get replaced and then cleaned up when they go out of scope. This seemed more efficient than how I sorta think std::move seems to work, since the new arrays are likely to be considerably smaller than the old ones. Is that reasoning sound?

utils/obsproc/NetCDFToIodaConverter.h
AndrewEichmann-NOAA: Is the idea to return the new IodaVars instance?

utils/obsproc/NetCDFToIodaConverter.h
AndrewEichmann-NOAA: @guillaumevernieres To address your question about checking, I tried masking masking salinity values less than zero (which includes the fill values) and looking at the resulting IODA files and got what I expected, also when I accidentally did more or less the reverse.

utils/obsproc/NetCDFToIodaConverter.h
guillaumevernieres: correct @AndrewEichmann-NOAA , not this pr.

utils/obsproc/NetCDFToIodaConverter.h
AndrewEichmann-NOAA: (I responded this morning, but for some reason github is treating my comments like a code review that needs to be submitted??)

rmclaren: @AndrewEichmann-NOAA All std::move does is steal the pointer...  You can avoid unnecessary copies if you use it correctly...

utils/obsproc/NetCDFToIodaConverter.h
AndrewEichmann-NOAA: Done

utils/obsproc/NetCDFToIodaConverter.h
AndrewEichmann-NOAA: Now that I'm using a whole class instance, I wonder if I can do it to `this` - feels a little Indian rope trick-like

guillaumevernieres: sounds good to me @AndrewEichmann-NOAA , we'll switch to `_` in a later PR!

guillaumevernieres (approved):
Thanks @AndrewEichmann-NOAA

apchoiCMD (approved):

PR #671, opened 2023-10-10
GHRSST and Superobing

guillaumevernieres:
## Description
This feature PR adds the following:

- The `GHRSST` netcdf reader, this work was mostly done by @ShastriPaturi
- A simple binning function that assumes the original obs are organized in a structured grid and nearest neighbors can be inferred from the indices. The binning parameters are passed through the yaml file which now looks like this:
```yaml
provider: GHRSST
window begin: 2018-04-15T06:00:00Z
window end: 2018-04-15T12:00:00Z
variable: seaSurfaceTemperature
binning:
  stride: 2
  min number of obs: 5
output file: ghrsst_sst_mb_20210701.ioda.nc
input files:
- ghrsst_sst_mb_202107010000.nc4
- ghrsst_sst_mb_202107010100.nc4
```
### Testing
A ctest was added, which should cover the new code.

**Realistic case:**
36 files for a grand total of **~5.8B obs** , which I think is representative of what we will be doing for 1 instrument.

| Number of PE's | Runtime          | Total Memory               | Memory per Task          |
|----------------------|------------------|----------------------------|--------------------------|
| Serial | 95.77 sec| total: 6.47 Gb            | per task: min = 6.47 Gb, max = 6.47 Gb |
| 2 PEs | 52.82 sec | total: 12.93 Gb           | per task: min = 6.47 Gb, max = 6.47 Gb |
| 4 PEs | 40.24 sec | total: 25.84 Gb           | per task: min = 6.46 Gb, max = 6.46 Gb |

I can't test with more PE's, my laptop is limited to 32GB of RAM.

Sooooo ... 40sec vs the 30mn we had to deal with before :) . Two weeks well spent I would say!

### What's left to do
A lot, here's an incomplete list:
- no mpi gather of the optional metadata yet
- add things like surface wind, rfi, std dev of obs value, etc ... to the optional metadata
- global attributes
- ...


guillaumevernieres:
The build is failing in `gdas_meanioda.h`. It's not finding the ```ObsSpace``` constructor.
See recent `ioda` [PR#1121](https://github.com/JCSDA-internal/ioda/pull/1121) . We need to now pass 
```c++
const util::TimeWindow
```
instead of a 
```c++
const util::DateTime
```

apchoiCMD (approved):

guillaumevernieres:
I'm not waiting for Orion!

utils/obsproc/Ghrsst2Ioda.h
```diff
@@ -0,0 +1,195 @@
...
+          iodaVars.preQc(loc)     = 0;
+          iodaVars.datetime(loc)  = seconds_s[i][j];
+          loc += 1;
+        }
+      }
+
+      // Remove
+      Eigen::Array<bool, Eigen::Dynamic, 1> boundsCheck =
+        (iodaVars.obsVal > -3.0 && iodaVars.obsVal < 50.0);
+      iodaVars.trim(boundsCheck);
```
rmclaren, line 190: Did you mean to remove this (looking at the comment)?

utils/obsproc/superob.h
```diff
@@ -0,0 +1,62 @@
...
+      int stride;
+      int minNumObs;
+      fullConfig.get("binning.stride", stride);
+      fullConfig.get("binning.min number of obs", minNumObs);
+
+      // Calculate the dimensions of the subsampled array
+      int numRows = inputArray.size();
+      int numCols = inputArray[0].size();
+      int subsampledRows = (numRows + stride - 1) / stride;
+      int subsampledCols = (numCols + stride - 1) / stride;
```
rmclaren, line 26: Why not `int subsampledRows = std::ceil(numRows/stride)` etc?

rmclaren (approved):

utils/obsproc/Ghrsst2Ioda.h
guillaumevernieres: > Did you mean to remove this (looking at the comment)?

Ha! ... Nope, I probably meant to write "Remove out of bound obs" or something.

PR #679, opened 2023-10-16
Add in 3DVar FGAT (with time interpolation) for aerosol DA 

### LINKED ISSUE #568: Add FGAT for aerosol DA
Need to explore the use of FGAT for aerosol DA for GDASv17 and see if it is viable. Given that we will still be using Gaussian history for GSI, it may not be feasible and may have to wait until GDASv18. We should explore the benefit of FGAT vs pure 3DVar in the AOD departures, and also see about interpolation and or multiple RESTART writes as a potential option.

CoryMartin-NOAA:
Closes #568 

This PR adds a 3DVar FGAT YAML for aerosol variational DA. Once this is merged, a global-workflow corresponding PR will be made.

### DIFF
--- a/parm/aero/obs/config/viirs_n20_aod.yaml
+++ b/parm/aero/obs/config/viirs_n20_aod.yaml
@@ -14,6 +14,7 @@ obs space:
   channels: 4
 get values:
   interpolation method: barycentric
+  time interpolation: linear
 obs operator:
   name: AodCRTM
   Absorbers: [H2O,O3]
--- a/parm/aero/obs/config/viirs_npp_aod.yaml
+++ b/parm/aero/obs/config/viirs_npp_aod.yaml
@@ -14,6 +14,7 @@ obs space:
   channels: 4
 get values:
   interpolation method: barycentric
+  time interpolation: linear
 obs operator:
   name: AodCRTM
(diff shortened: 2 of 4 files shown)


CoryMartin-NOAA:
A quick test does show differences in H(x) between this and the existing YAML template.

andytangborn (dismissed):
It looks fine. The important line (from what I understand) is  tstep: PT3H. I'm assuming everything else is pretty much the same, though I see a few differences from 3dvar_gfs_aero.yaml

parm/aero/obs/config/viirs_n20_aod.yaml
```diff
@@ -14,6 +14,7 @@ obs space:
   channels: 4
 get values:
   interpolation method: barycentric
```
guillaumevernieres, line 16: Ha! I wasn't aware this was working @CoryMartin-NOAA. Nice!

CoryMartin-NOAA: Well it wasn't, but there is an OOPS bugfix PR so I figured might as well.

guillaumevernieres (approved):
:tada:

PR #681, opened 2023-10-17
Fix for aerosol staticb yaml

andytangborn:
This PR renames staticb_bump.yaml to staticb_bump_aero.yaml so that it is consistent with the workflow. It also adjusts the indentation on the stddev part of the yaml so that the standard deviation files are read propler.y


CoryMartin-NOAA:
Let's double check that this renaming is needed. Also, a 'git mv' would be preferred as it would track changes better.

CoryMartin-NOAA:
https://github.com/NOAA-EMC/global-workflow/blob/3de102c28daf9eb50ce65e0afe911f5a416ce861/ush/python/pygfs/task/aero_analysis.py#L88 suggests it should not be bump_aero?

andytangborn:
> https://github.com/NOAA-EMC/global-workflow/blob/3de102c28daf9eb50ce65e0afe911f5a416ce861/ush/python/pygfs/task/aero_analysis.py#L88 suggests it should not be bump_aero?

I've needed to manually change bump to bump_aero to make it run. But I'm not sure what is causing this. Do you mean that this line in aero_analysis.py implies that it needs to be staticb_bump.yaml?

if self.task_config.get('STATICB_TYPE', 'identity') in ['bump']:

CoryMartin-NOAA:
@andytangborn correct, what is your line in `config.aeroanl`? Is it `bump_aero` there when it should just be `bump`?

andytangborn:
@CoryMartin-NOAA  You know, it's possible that I've been changing it when I set up my experiments. I always do xxdiff on the new config files with previous experiments, and I may have changed it at that point without realizing it. Let me set up a new experiment and see if the default is just staticb_bump.yaml.

andytangborn:
@CoryMartin-NOAA   Yes, that was the reason. The default is staticb_bump.yaml   I'll change it back.

andytangborn:
Now the question is, if I  use 'mv staticb_bump_aero.yaml staticb_bump.yaml', is github going to consider that to be 2 moves, or will it be unchanged?

CoryMartin-NOAA:
if you do that and add/commit, the PR here should just show your changes and nothing more.

parm/aero/berror/staticb_bump.yaml
```diff
@@ -5,19 +5,21 @@ saber central block:
...
+saber outer blocks:
+- saber block name: StdDev
+  read: 
+    model file: 
+      datetime: '2021-02-27T00:00:00Z'
+      set datetime on read: true
+      filetype: fms restart
+      psinfile: true
+      datapath: $(DATA)/berror
+      filename_core: 20210227.000000.stddev.fv_core.res.nc
```
CoryMartin-NOAA, line 22: We need to figure out how to make these dates generic and filled in at runtime based on the config file(s). But that can be a separate PR.

CoryMartin-NOAA (approved):

parm/aero/berror/staticb_bump.yaml
andytangborn: Yes, I was thinking about that. There will be several different dates for staticB files available.

PR #688, opened 2023-10-20
Fix issues on hera with test data

CoryMartin-NOAA:
JCSDA changed the logic by which test data is downloaded.

First, new top level CMake flags have to be added, else it will try to download tarballs rather than use the cloned repos.
Second, CRTM has no choice, if they are not already available, the code will try to download a tarball.
These tarballs are currently unavailable from Hera, as the host server is blocked. The logic should work fine on other machines though.

This PR fixes this on Hera but including the CMake flags for all platforms, and staging/linking the CRTM coefficients from a place I have them already installed on Hera.


RussTreadon-NOAA (approved):
Clone `feature/testdata` on Hera.  Execute `/build.sh -v`.    All bundles cloned.  Build still running ... up to `ufo`.

Approve.

guillaumevernieres (approved):
Thanks @CoryMartin-NOAA

RussTreadon-NOAA:
FYI, the Hera build successfully completed.

ShastriPaturi:
I could also build on Hera.

PR #690, opened 2023-10-23
Only use the gdas backgrounds

guillaumevernieres:
When running the `gfs` cycle, the marine DA prep was trying to link to the gfs background, which is not correct.

### DIFF
--- a/scripts/exgdas_global_marine_analysis_prep.py
+++ b/scripts/exgdas_global_marine_analysis_prep.py
@@ -129,7 +129,7 @@ def gen_bkg_list(bkg_path, out_path, window_begin=' ', yaml_name='bkg.yaml', ice
     fcst_hrs = list(range(3, 10, dt_pseudo))
     files = []
     for fcst_hr in fcst_hrs:
-        files.append(os.path.join(bkg_path, f'{RUN}.t'+gcyc+'z.ocnf'+str(fcst_hr).zfill(3)+'.nc'))
+        files.append(os.path.join(bkg_path, 'gdas.t'+gcyc+'z.ocnf'+str(fcst_hr).zfill(3)+'.nc'))
 
     # Identify the ocean background that will be used for the  vertical coordinate remapping
     ocn_filename_ic = os.path.splitext(os.path.basename(files[0]))[0]+'.nc'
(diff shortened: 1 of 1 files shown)


CoryMartin-NOAA (dismissed):
Fine, but FYI, the current atmospheric convention is to define:
```
GDUMP=${GDUMP:-"gdas"}
```
and then use `${GDUMP}.t${gcyc}z`...

guillaumevernieres:
> Fine, but FYI, the current atmospheric convention is to define:
> 
> ```
> GDUMP=${GDUMP:-"gdas"}
> ```
> 
> and then use `${GDUMP}.t${gcyc}z`...

OK, thanks for the feed back @CoryMartin-NOAA . I'll make the changes.

ShastriPaturi (dismissed):

scripts/exgdas_global_marine_analysis_prep.py
```diff
@@ -507,12 +507,12 @@ def find_clim_ens(input_date):
 # Copy initial condition
 ics_list = []
 # ocean IC's
-mom_ic_src = glob.glob(os.path.join(bkg_dir, f'{RUN}.*.ocnf003.nc'))[0]
```
CoryMartin-NOAA, line 510: are you defining `$GDUMP` anywhere?

guillaumevernieres: Yes @CoryMartin-NOAA , in the jjob ... Luck, or unfortunate copy/paste that became a blessing?

CoryMartin-NOAA: ok sure, but then don't you need something like line 126?
https://github.com/NOAA-EMC/GDASApp/blob/61aa906e8e7bf68dd3db54a1e2d5f39a8b6f0224/scripts/exgdas_global_marine_analysis_prep.py#L126

guillaumevernieres: Ha!!! I clearly didn't test these changes! Thanks @CoryMartin-NOAA .

CoryMartin-NOAA (approved):

PR #691, opened 2023-10-24
Comment out aero tests

CoryMartin-NOAA:
The aerosol tests are failing, so until we have a fix in OOPS, will comment them out. This will allow feature/stable-nightly to update properly.


guillaumevernieres (approved):
:cry:

Issue #88, opened 2022-07-08
Ensure R2D2 is archiving diags properly


CoryMartin-NOAA:
Looks like currently the `post` jobs are archiving diags to R2D2 but treating them as input observations (with a provider, etc.). Need to make sure the post jobs and any associated ush/ufsda scripts are configured correctly to archive the diags in the right place.

Issue #72, opened 2022-06-29
Add initial YAML files for AMSU-A


CoryMartin-NOAA:
- [ ] amsua_aqua
- [ ] amsua_metop-a
- [ ] amsua_metop-b
- [ ] amsua_metop-c
- [ ] amsua_n15
- [ ] amsua_n18

PR #696, opened 2023-10-28
Update satwind yamls for testing and end-to-end

emilyhcliu:
Update satellite wind YAML files for the following:

- Update test data set:
     - error inflation for duplicate obs and VarQC turned off.
     - save the original 10-meter wind reduction factor, not the one modified by GSI.  The modification factor is implemented in the UFO `VertInterp`

- Remove minvalue/maxvalue thresholds in error assignments.  The use of the threshold is not correct.  The updated UFO detects the mistake. The removal does not affect results.

- Update observation operator options and parameters (due to changes from GMAO in UFO)

One YAML for UFO evaluation for HofX only with GSI obs/geovals
One YAML for UFO evaluation for HofX + QC with GSI obs/geovals
Two YAMLS (goes-16 and goes-17) for End-to-End test

100% replication of GSI results (details documented in [Issue #88](https://github.com/NOAA-EMC/JEDI-T2O/issues/88))
The end-to-end test results are closer to GSI after fixing the water area fraction threshold (detail documented in [Issue #72](https://github.com/NOAA-EMC/JEDI-T2O/issues/72))

Note: After merging this PR, we should see the green light for the satellite wind test.


ush/ufoeval/run_ufo_hofx_test.sh
```diff
@@ -100,12 +100,7 @@ else
    yamlpath=$GDASApp/parm/atm/obs/testing/${obtype}.yaml
 fi
 
-#exename=test_ObsFilters.x
```
CoryMartin-NOAA, line 103: Why get rid of this here?

emilyhcliu: For each data type we have two yaml files. One for HofX only (no data filtering) and one for HofX + QC.  
Original the test_ObsOperator.x is for the hofx only, and test_OBsFilters.x is for HofX+QC.  

However, I found that it is better include the error assignment part of the filter to the HofX only test for the following two reasons:
(1) the default EVA configuration will not failed due to missing observation errors
(2) we can validate the original (input) observation error with the GsiInput observation error stored in the obs file.  

Therefore, the test_ObsOperator.x was replaced with test_ObsOperator.x

ADCollard:
I can replicate Emily's results in [Issue 88](https://github.com/NOAA-EMC/JEDI-T2O/issues/88).  Changes look good approving.

ADCollard (dismissed):
All looks good.

CoryMartin-NOAA (dismissed):

nicholasesposito (dismissed):

CoryMartin-NOAA (approved):

PR #709, opened 2023-11-03
Add error compare plot

ADCollard:
Just an extra plot added to the template for conventional obs.   It compares obs errors as a function of temperature (filtering out QC'd obs).


CoryMartin-NOAA (approved):
Excellent (except for those with red-green colorblindness 🥇  )

PR #713, opened 2023-11-09
Change yamls for goes amvs

CoryMartin-NOAA:
This PR makes the filenames that JEDI is looking for for GOES AMVs consistent with the files produced by the BUFR to IODA converter (hopefully)

### DIFF
--- a/parm/atm/obs/config/satwind_goes-17.yaml
+++ b/parm/atm/obs/config/satwind_goes-17.yaml
@@ -3,7 +3,7 @@ obs space:
   obsdatain:
     engine:
       type: H5File
-      obsfile: $(DATA)/obs/$(OPREFIX)satwind.abi_goes-17.{{ current_cycle | to_YMDH }}.nc4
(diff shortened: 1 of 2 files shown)


RussTreadon-NOAA (approved):
Changes consistent with what I used in GDAS-validation tests.

parm/atm/obs/config/satwind_goes-17.yaml
```diff
@@ -3,7 +3,7 @@ obs space:
   obsdatain:
     engine:
       type: H5File
-      obsfile: $(DATA)/obs/$(OPREFIX)satwnd.abi_goes-16.tm00.nc
+      obsfile: $(DATA)/obs/$(OPREFIX)satwnd.abi_goes-17.tm00.nc
```
RussTreadon-NOAA, line 6: Good catch!

CoryMartin-NOAA: that's what I get for a quick copy/paste :-)

PR #717, opened 2023-11-09
GHRSST bug fix ... an more

guillaumevernieres:
Quite a few bug fixes, but the `sstdtime` is still not correct.

I added a feature in the testing so our ctest for GHRSST does more than just checking if things run without crashing! The GHRSST application outputs a few basic stats from the ioda variables:

```
Reading ghrsst_sst_mb_202107010000.nc4
obsVal:
    Min: 276.708
    Max: 276.9
    Sum: 4982.63
obsError:
    Min: 0.32
    Max: 0.32
    Sum: 5.76
...
```
See file below.
Still not enough but better than doing nothing!


rmclaren (dismissed):

apchoiCMD:
Thanks @guillaumevernieres I will test it on hpc

utils/obsproc/Ghrsst2Ioda.h
```diff
@@ -28,10 +28,21 @@ namespace gdasapp {
...
       oops::Log::info() << "Processing files provided by GHRSST" << std::endl;
 
+      // Get the sst bounds from the configuration
+      std::string sstUnits;
+      fullConfig_.get("bounds.units", sstUnits);
+      float sstMin;
+      fullConfig_.get("bounds.min", sstMin);
+      float sstMax;
+      fullConfig_.get("bounds.max", sstMax);
+      if ( sstUnits == "C" ) {
```
AndrewEichmann-NOAA, line 39: If the units are not definitively known in advance, how safe is the assumption that not-"C" equals Kelvin? And are the increments in the lines below in the right direction?

utils/obsproc/Ghrsst2Ioda.h
```diff
@@ -57,64 +68,51 @@ namespace gdasapp {
...
       std::string refDate;
       ncFile.getVar("time").getAtt("units").getValues(refDate);
 
       // Read sst_dtime to add to the reference time
-      int sstdTime[dimTime][dimLat][dimLon];  // NOLINT
-      ncFile.getVar("sst_dtime").getVar(sstdTime);
-      float dtimeOffSet;
-      ncFile.getVar("sst_dtime").getAtt("add_offset").getValues(&dtimeOffSet);
+      // TODO(AMG): What's below does not read the field the same way python does
+      std::vector<int32_t> sstdTime(dimTime*dimLat*dimLon);
```
AndrewEichmann-NOAA, line 91: Minor style thing - is `dimTime*dimLat*dimLon` calculated elsewhere as `numloc` or some such?

utils/obsproc/Ghrsst2Ioda.h
apchoiCMD: Probably lines below makes sst "C " @AndrewEichmann-NOAA 
`
sst[i][j] = static_cast<float>(sstObsVal[index]) * sstScaleFactor + sstOffSet
            - static_cast<float>(sstObsBias[index]) * biasScaleFactor;
`

guillaumevernieres: No @apchoiCMD , it doesn't. The output is in K.

apchoiCMD: @guillaumevernieres Yes, thanks for pointing me out.

guillaumevernieres: @AndrewEichmann-NOAA , maybe the variable name is confusing. `sstUnits` is meant to be the unit of the bounds read from the yaml.

utils/obsproc/Ghrsst2Ioda.h
guillaumevernieres: nope.

guillaumevernieres: nope.

utils/obsproc/Ghrsst2Ioda.h
```diff
@@ -124,51 +122,57 @@ namespace gdasapp {
...
 
       // number of obs after subsampling
       int nobs = sst_s.size() * sst_s[0].size();
 
       // Create instance of iodaVars object
       gdasapp::IodaVars iodaVars(nobs, {}, {});
 
-      // unix epoch at Jan 01 1981 00:00:00 GMT+0000
+      // Reference time is Jan 01 1981 00:00:00 GMT+0000
       iodaVars.referenceDate_ = refDate;
```
ShastriPaturi, line 188: The refDate from the input file is: "seconds since 1981-01-01 00:00:00".
The refDate for the  string dateTime:units should be "seconds since 1981-01-01T00:00:00Z", otherwise the UFO fails.
I have a fix that can be added...
      // Reference time is Jan 01 1981 00:00:00 GMT+0000
      // The string dateTime:units = "seconds since 1981-01-01T0:00:00Z"
      // Currentlythe refDate is "seconds since 1981-01-01 00:00:00"
      std:: string refDaten;
      refDaten.append(refDate.substr(0,24));
      refDaten.append("T");
      refDaten.append(refDate.substr(25,8));
      refDaten.append("Z");
      oops::Log::info() << refDaten << std::endl;
      iodaVars.referenceDate_ = refDaten;

ShastriPaturi (changes requested):

utils/obsproc/Ghrsst2Ioda.h
apchoiCMD: @guillaumevernieres @ShastriPaturi @AndrewEichmann-NOAA Hi all, I am still confused that ALL old python converters has ""seconds since 1970-01-01 00:00:00" as string dateTime:units,,, so do we need to match reference dTime as old python has had?

@ShastriPaturi I am not sure you mention reference Date "seconds since 1980-01-01 00:00:00". Am I thinking wrong?

ShastriPaturi: @apchoiCMD I found this issue when I ran the 3dvar_godas ctest.

guillaumevernieres: We're using the reference date from the file @apchoiCMD , which is 1980-01-01 . So yes, the datetime will be different.

apchoiCMD (approved):

utils/obsproc/Ghrsst2Ioda.h
guillaumevernieres: I fixed your bug @ShastriPaturi ...

utils/obsproc/Ghrsst2Ioda.h
guillaumevernieres: Tentatively assuming you are OK with the explanation above @AndrewEichmann-NOAA and marking the conversation as "resolved ".

ShastriPaturi (approved):
ctest passes

utils/obsproc/Ghrsst2Ioda.h
ShastriPaturi: Thanks @guillaumevernieres

guillaumevernieres:
> Thanks @guillaumevernieres I will test it on hpc

Can I assume you approved because you tested on hpc @apchoiCMD ?

apchoiCMD:
Yes, it was tested and all are passed. @guillaumevernieres sorry for the late reply!

guillaumevernieres:
I'll merge later today (unless somebody objects) since the failure are unrelated to the work done for this PR. @ShastriPaturi , could you have a look at the failing tests?

PR #732, opened 2023-11-15
Adding ocean basin in the Metadata

guillaumevernieres:
#### What was done
- Reorganization of what's in `NetCDFToIodaConverter.h`: Moved structs and functions to `util.h` as an attempt to make the code slightly cleaner
- Added a simple OceanMask struct that reads in regional ocean masks from [reccap2 ocean mask](https://reccap2-ocean.github.io/regions/). The struct has a method to return the integer mask at a specific location.
- Fixed a bug in `trim`, related to the optional float metadata
- implementation of the new feature in the RADS converter

#### Notes
It's currently not part of the `CI`, something to think about in the future. We could add `RECCAP2_region_masks_all_v20221025.nc` to the tarball or stage it on hpc somewhere ...

### DIFF
--- a/utils/obsproc/Ghrsst2Ioda.h
+++ b/utils/obsproc/Ghrsst2Ioda.h
@@ -26,7 +26,7 @@ namespace gdasapp {
     }
 
     // Read netcdf file and populate iodaVars
-    gdasapp::IodaVars providerToIodaVars(const std::string fileName) final {
+    gdasapp::obsproc::iodavars::IodaVars providerToIodaVars(const std::string fileName) final {
       oops::Log::info() << "Processing files provided by GHRSST" << std::endl;
 
       // Get the sst bounds from the configuration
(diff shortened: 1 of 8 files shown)


ShastriPaturi (approved):
Looks fine to me. Didn't run for RADS to test the ocean basins

guillaumevernieres:
Tested with more obs:
ocean-basins

guillaumevernieres:
Well ... That's a new one
```
1606 - test_gdasapp_soca_incr_handler (Failed)
```

AndrewEichmann-NOAA (approved):
nothing obviously wrong

AndrewEichmann-NOAA:
> Tested with more obs: ocean-basins

What's zero/red? Continental shelf?

guillaumevernieres:
> > Tested with more obs: ocean-basins
> 
> What's zero/red? Continental shelf?

Yes @AndrewEichmann-NOAA . I wonder if we should make sure we don't flag land ... Humm ...

Issue #742, opened 2023-11-17
Typo in parm/atm/obs/config/satwind_goes-17.yaml


RussTreadon-NOAA:
https://github.com/NOAA-EMC/GDASApp/blob/40e81bb07b2d0496c4e276bc04610155ea01123f/parm/atm/obs/config/satwind_goes-17.yaml#L20

differs from

https://github.com/NOAA-EMC/GDASApp/blob/40e81bb07b2d0496c4e276bc04610155ea01123f/parm/atm/obs/config/satwind_goes-16.yaml#L20

The order in `satwind_goes-17.yaml` is reversed compared to `satwind_goes-16.yaml`.   The order in `satwind_goes-16.yaml` is correct.  

This issue is opened to document the typo and its correction.   The correction can be committed to [`feature/gdas-validation-satwind`](https://github.com/NOAA-EMC/GDASApp/blob/feature/gdas-validation-satwind/modulefiles/GDAS/orion.lua)

RussTreadon-NOAA:
@emilyhcliu , a test of the fix is [described](https://github.com/NOAA-EMC/JEDI-T2O/issues/98#issuecomment-1815632229) in JEDI-T2O issue [#98](https://github.com/NOAA-EMC/JEDI-T2O/issues/98).

Would you like me to make the fix in [`feature/gdas-validation-satwind`](https://github.com/NOAA-EMC/GDASApp/blob/feature/gdas-validation-satwind/modulefiles/GDAS/orion.lua), open a PR, and list you as a reviewer?

RussTreadon-NOAA:
The following has been done in a working copy of [feature/gdas-validation-satwind](https://github.com/NOAA-EMC/GDASApp/blob/feature/gdas-validation-satwind/modulefiles/GDAS/orion.lua) on Orion
1. merge `feature/gdas-validation` into `feature/gdas-validation-satwind`
2. correct typo in `parm/atm/obs/config/satwind_goes-17.yaml`

Given your green light, @emilyhcliu , I'll push the updated branch to NOAA-EMC/GDASApp, open a PR, and ask you to review.

CoryMartin-NOAA:
@RussTreadon-NOAA sorry I saw you also created a branch, whoops! I didn't intend to reproduce work (but it was a quick fix) I think this can be closed now.

PR #98, opened 2022-07-25
Submit the marine/soca prep and run step as a batch job

guillaumevernieres:
## Description
Leverage the `run_jedi_exe.py` application to allow submission of a batch script that will execute:
```
 exgdas_global_marine_analysis_prep.py
```
and
```
exgdas_global_marine_analysis_run.sh
```

### Features added and bug fixes
- removed hard-coded date ... Oops
- generate MOM6 /SOCA Fortran namelist from configuration
- pass DA variables from config ... not passed down to the yamls yet 
- I got tiered of `chmoding +x`  the `submit_job.sh` bash script generated by `run_jedi_exe.py`. It is now made executable when written by `run_jedi_exe.py`

### Known problems
- I made assumptions on what the `global-workflow` (GW) environment is, it's probably mostly wrong but it will be fixed once we have access to the UFS-mom6/cic6/data atm model as a cycling option of the GW ... Patience!
- The added feature is currently not tested as part of the `ci`. To be revisited in the future and possibly added as part of a suite or **regression tests**.
- I have not tested the atmospheric application to make sure I did not break that part of the code ... TODO before I un-draft this PR
- It definitely will not work as is for the atmospheric `prep` and `run`. That need to be sorted out once I understand better the GW environement
- I need to add the R2D2 config to the experiment yaml, it currently can only point to a tiny R2D2 test database 

### How to test this PR on Orion
I have only prepared static files for `Orion` so far. Here's what needs to be done to test:

```console
$ git clone -b feature/submit-prep-run https://github.com/NOAA-EMC/GDASApp.git
$ cd GDASApp
$ module use $PWD/modulefiles
$ module load GDAS/orion
$ ./build.sh
```
it takes a while to build the entire bundle ... So better forget for an hour and go do something else.

There are many ways one could prepare an experiment, here's one:

 #### Create an experiment directory somewhere
```console
$ mkdir test-soca-gdasapp     # Create an experiment directory somewhere
$ cd test-soca-gdasapp
$ ln -s <pat to GDASApp>/ush/run_jedi_exe.py . # link of convenience
$ cp <pat to GDASApp>/ush/examples/run_scripts/soca5deg.yaml .  # make a copy of an example yaml config
```
#### Edit `soca5deg.yaml` 
```yaml
working directory: /work2/noaa/da/gvernier/runs/test-gdasapp/test-5deg # Needs editing
# replace the above by working directory: <path to test-soca-gdasapp>/test-5deg
[...]

PR #743, opened 2023-11-17
Fix YAML so minimization works for satwind G17

CoryMartin-NOAA:
@RussTreadon-NOAA noted that `obs linear operator` should be `linear obs operator` in the GOES-17 satwind YAML. This fixes that.


RussTreadon-NOAA (approved):
Tested revised `parm/atm/obs/config/satwind_goes-17.yaml` in working copy of gdas-validation.  It works as reported in GDASApp issue #742 and JEDI-T2O issue [#98](https://github.com/NOAA-EMC/JEDI-T2O/issues/98)

PR #771, opened 2023-11-27
Add an option to point to the "enkfgdas" backgrounds instead of the pre-generated 30 member ensemble

guillaumevernieres:
#### Work done 
Mainly what the title says. The default ctest is now assuming that `DOHYBVAR`=`YES`, I don't really think it's worth it to cover the `DOHYBVAR`=`NO` in the ctest. I may change my mind, but in both cases, `soca` is configured to use the hybrid envar solver, the only difference is the origin of the ensemble members.

#### Work left to be done
We need to define an ensemble "COM_OCEAN/ICE_HISTORY_PREV" environment variable in the prep jjob (in the g-w). Flagged as a `TODO` in the code.


AndrewEichmann-NOAA (approved):
no complaints

ShastriPaturi (approved):
admissible

PR #786, opened 2023-12-01
Fix test reference mismatch

### LINKED ISSUE #783: Test reference mismatch
The issue is a bug in the handling of the datetime with a dependency of a reference date that may or may not be gmt based. Oops!

              Looks like a reference mismatch @apchoiCMD :
```console
The following tests FAILED:
        1314 - test_gdasapp_util_smap2ioda (Subprocess aborted)
        1316 - test_gdasapp_util_icecamsr2ioda (Subprocess aborted)
Errors while running CTest
```

_Originally posted by @guillaumevernieres in https://github.com/NOAA-EMC/GDASApp/issues/719#issuecomment-1832844732_

apchoiCMD:
#### What this PR fixed
- Two new ioda converters <del>that work with `mktime` in a C library function</del> that need to ensure consistent behavior across different machines with different time zones (e.g., Orion (CST) and Hera (UTC))
- <del>Set the time zone explicitly using the `tzset` function with UTC time zone</del>
- Used ```oops/utils``` to make dateime to Julian date so now code is a lot simpler than before

Fixed https://github.com/NOAA-EMC/GDASApp/issues/783


utils/obsproc/IcecAmsr2Ioda.h
```diff
@@ -65,6 +66,10 @@ namespace gdasapp {
       ncFile.getVar("Scan_Time").getVar(oneTmpdateTimeVal.data());
       iodaVars.referenceDate_ = "seconds since 1970-01-01T00:00:00Z";
 
+      // Set the time zone (UTC)
+      putenv("TZ=UTC");
```
guillaumevernieres: Changing the environment variable is sketchy. We need to figure out a better way to do this. 
Did you dig into the `ioda` code? a little bit of grepping showed a few possibly useful tools here `gdas.cd/ioda/src/core/IodaUtils.h`

guillaumevernieres (commented):
I don't think changing the environment is a safe thing to do, you're assuming `TZ` isn't used anywhere else.

utils/obsproc/IcecAmsr2Ioda.h
apchoiCMD: @guillaumevernieres Yes, I did try to use other ways but it took some time. Agree with your comment, I will think more efficiently. Thanks!

guillaumevernieres: dig into `ioda` @apchoiCMD , julian dates are handled somewhere in the jedi code.

guillaumevernieres (commented):
Have a look here @apchoiCMD :
```https://github.com/JCSDA/ioda/blob/develop/src/core/IodaUtils.h```

File: utils/obsproc/IcecAmsr2Ioda.h [hunk 2]
```diff
@@ -66,18 +69,31 @@ namespace gdasapp {
...
-        time_t epochtime = std::mktime(&timeinfo);
-        iodaVars.datetime_(index) = static_cast<int64_t>(epochtime);
+        int year = oneTmpdateTimeVal[i];
+        int month = oneTmpdateTimeVal[i+1];
+        int day = oneTmpdateTimeVal[i+2];
+        int hour = oneTmpdateTimeVal[i+3];
+        int minute =  oneTmpdateTimeVal[i+4];
+        int second = static_cast<int>(oneTmpdateTimeVal[i+5]);
+
+        // Bypassing validYYYYMMDD function within dateToJulian
```
guillaumevernieres: I'm not sure the comment fits what's being done below.

utils/obsproc/IcecAmsr2Ioda.h
```diff
@@ -66,18 +69,31 @@ namespace gdasapp {
...
+        if (year == -9999 && month == -9999 && day == -9999 &&
```
guillaumevernieres: shouldn't it be or?

utils/obsproc/IcecAmsr2Ioda.h
```diff
@@ -66,18 +69,31 @@ namespace gdasapp {
...
+             hour == -9999 && minute == -9999 && second == -9999) {
+          year = 0, month = 0, day = 0;
+          hour = 0, minute = 0, second = 0;
+        }
+
+        uint64_t julianDate = util::datefunctions::dateToJulian(year, month, day);
+
+        // Convert a date to Julian date
+        // Subtract Julian date for January 1, 1970 (epoch)
```
guillaumevernieres: check English in the comment ... My frenglish tells me it's wrong.

utils/obsproc/Smap2Ioda.h
```diff
@@ -70,14 +74,23 @@ namespace gdasapp {
...
+      }
+
+      uint64_t julianDate = util::datefunctions::dateToJulian(year, month, day);
+
+      // Convert a date to Julian date
+      // Subtract Julian date for January 1, 1970 (epoch)
+      int daysSinceEpoch = julianDate - 2440588;
+
+      // Calculate seconds
+      int unixStartDay = (daysSinceEpoch + startDay) * 86400;
```
guillaumevernieres: I thought we were going away from the system's clock? If it's just a leftover variable name from the previous implementation, can we rename it to something without "unix".

guillaumevernieres (commented):
Thanks @apchoiCMD . Looks good, just a few minor comments. Did you check that the datetime is still correct?

utils/obsproc/Smap2Ioda.h
apchoiCMD: > I thought we were going away from the system's clock? If it's just a leftover variable name from the previous implementation, can we rename it to something without "unix".

OK, I just kept it what a previous left over

utils/obsproc/IcecAmsr2Ioda.h
apchoiCMD: > check English in the comment ... My frenglish tells me it's wrong.

OMG, yes, I will correct it.

utils/obsproc/IcecAmsr2Ioda.h
apchoiCMD: > I'm not sure the comment fits what's being done below.

Will write down more clearly-

apchoiCMD:
> Thanks @apchoiCMD . Looks good, just a few minor comments. Did you check that the datetime is still correct?

Yes, I have checked min & max output by using online tools (UTC) compared to original netcdf data, too! Everything was ok!

utils/obsproc/IcecAmsr2Ioda.h
ShastriPaturi: Change it to:
"Subtract Julian day from January 1, 1970 (convert to epoch)"

AndrewEichmann-NOAA (dismissed):
Aside from the filtering of obs with bad dates, looks good

guillaumevernieres (approved):
Looks good. Thanks @apchoiCMD .

ShastriPaturi (approved):
Approving based on others comments

Issue #785, opened 2023-12-01
Updates to ADPSFC, ACFT_PROFILES and conventional ps bufr2ioda python scripts


nicholasesposito:
Some updates were made to the following bufr2ioda python sripts.

There were a number of changes needed to some of the converter scripts. The changes are outlined in issue #785 .

From that issue:
Some updates were made to the following bufr2ioda python sripts.

1. acft_profiles prepbufr script
      - added units to instantaneousAltitudeRate (should be ms-1)
      - added ObsError for all variables
      - added windEastward to QM
      - dateTime code fix to make it run faster. Missing is now _, 

2. ADPSFC prepbufr script - changes to the SFCSHP script that came up during the code sprint needed to be done to ADPSFC as well
      - I added ObsError for stationPressure. 
      - I edited the height variable down from heightOfStation. 
      - I added stationElevation to MetaData.
      - datetime fix

3. the conventional ps script.
      - Added ObsSubType ([GSI code](https://github.com/NOAA-EMC/GSI/blob/44a8f59ef5cb6def27ba27fcc6a35c074a30332b/src/gsi/read_prepbufr.f90#L693C1-L729C14)) 
      - Added ObsError
      - Added stationElevation to MetaData
      - heightOfStation -> height
      - fixed types
      - ObsError and QM ==0 are changed to "Missing"
      - dateTime codefix. Runs slightly faster. Missing is no longer 0, it is _,

emilyhcliu:
@nicholasesposito These stationPressures are from your conventional_ps converter.  
They look good.  

![sfc_ps](https://github.com/NOAA-EMC/GDASApp/assets/36091766/2208a6a4-8560-48f8-9846-44decf82fa9a)
![sfcship_ps](https://github.com/NOAA-EMC/GDASApp/assets/36091766/e8eb800e-9a73-4347-a5bc-68ec270644ad)
![sonde_ps](https://github.com/NOAA-EMC/GDASApp/assets/36091766/c14e4d2e-3303-4b12-8bd1-7e0ae9a44aa3)

I check the conventional ps IODA obs file.  The stationPressures are comparable to those from [GSI.]( https://docs.google.com/presentation/d/1OvVMYOutxVyk0GE93Jzoh3hQVoj-dCgrwsVWDDlPT1s/edit#slide=id.g27d3030a211_0_0)

nicholasesposito:
[...]

PR #792, opened 2023-12-04
Feature/ne acftpro and conventional updates

nicholasesposito:
There were a number of changes needed to some of the converter scripts. The changes are outlined in issue #785 .

From that issue:
Some updates were made to the following bufr2ioda python sripts.

1. acft_profiles prepbufr script
      - added units to instantaneousAltitudeRate (should be ms-1)
      - added ObsError for all variables
      - added windEastward to QM
      - dateTime code fix to make it run faster. Missing is now _, 

2. ADPSFC prepbufr script - changes to the SFCSHP script that came up during the code sprint needed to be done to ADPSFC as well
      - I added ObsError for stationPressure. 
      - I edited the height variable down from heightOfStation. 
      - I added stationElevation to MetaData.
      - datetime fix

3. the conventional ps script.
      - Added ObsSubType
      - Added ObsError
      - Added stationElevation to MetaData
      - heightOfStation -> height
      - fixed types
      - ObsError and QM ==0 are changed to "Missing"
      - dateTime codefix. Runs slightly faster. Missing is no longer 0, it is _,


PraveenKumar-NOAA (dismissed):
Looks good.

ush/ioda/bufr2ioda/bufr2ioda_adpsfc_prepbufr.py
```diff
@@ -107,25 +110,30 @@ def bufr_to_ioda(config, logger):
 
     logger.debug(" ... Executing QuerySet: get ObsType ...")
     # ObsType
+    logger.debug(" ... Executing QuerySet: get ObsType ...")
```
CoryMartin-NOAA, line 122: isn't this going to print out twice in a row an identical message?

ush/ioda/bufr2ioda/bufr2ioda_conventional_prepbufr_ps.py
```diff
@@ -39,6 +39,33 @@ def Mask_typ_for_var(typ, var):
...
+def Compute_obsSubType(typ, t29, sid):
+
+    obssubtype = ma.array(np.full(typ.shape[0], 0))
+    for i in range(len(typ)):
+        newtyp = typ[i]
+        if ((typ[i] == 180) or (typ[i] == 280)) and (t29[i] == 562):
+            newval = 0
+            try:
+                newval = int(sid[i][-3:])
+            except Exception:
```
CoryMartin-NOAA: is there a specific type of exception we should be capturing here?

ush/ioda/bufr2ioda/bufr2ioda_adpsfc_prepbufr.py
nicholasesposito: True. Missed that. I'll delete it.

ush/ioda/bufr2ioda/bufr2ioda_conventional_prepbufr_ps.py
nicholasesposito: Probably a valueError. Let me test real quick

nicholasesposito: Yes. Updating now.

CoryMartin-NOAA (dismissed):

nicholasesposito:
Need to make ObsSubType its own group. Doing that now!

nicholasesposito:
Ok changes were tested and updated. Please re-check the small changes @CoryMartin-NOAA @PraveenKumar-NOAA

CoryMartin-NOAA (dismissed):

emilyhcliu (approved):

emilyhcliu (changes requested):
Wait.  Please do not merge yet.  I have one question.  
Will follow up with a comment.

ush/ioda/bufr2ioda/bufr2ioda_conventional_prepbufr_ps.py
```diff
@@ -39,6 +39,33 @@ def Mask_typ_for_var(typ, var):
...
+            try:
+                newval = int(sid[i][-3:])
+            except ValueError:
+                newval = 0
+
+            if newval > 500:
+                newtyp == typ[i] + 19
+
+        if ((typ[i] == 180) and ((t29[i] == 522) or (t29[i] == 523))):
+            newtyp == typ[i] + 18
```
emilyhcliu: @nicholasesposito for conventional stationPressure application, we do not need to create new types for bouy and ship.  This is specific to the NSST application (where the id_drifter and id_ship are set to true).   

This is my bad.  I did not read the GSI code carefully.  
This is the code section specific for NSST application.  It identifies buoy and ship data from prepbufr file. 
```
        if (id_drifter .and. (kx==180 .or. kx==280) .and. nint(hdr(3),r_double)==562) then
           rstation_id=hdr(4)
           read(c_station_id,*,iostat=ios) iwmo
           if (ios == 0 .and. iwmo > 0) then
              if(mod(iwmo,1000) >=500) then
                 kx = kx + 19
              end if
           end if
        end if

        if (id_ship .and. (kx==180) .and. (nint(hdr(3),r_double)==522 .or. nint(hdr(3),r_double)==523)) then
           rstation_id=hdr(4)
           kx = kx + 18
        end if
```

The code you spotted in GSI and already added to this PR is what we want.  
```
! temporary specify iobsub until put in bufr file
        iobsub = 0
        if(kx == 280 .or. kx == 180 ) iobsub=hdr(3)
        if(kx == 280 .or. kx ==180) then
          if ( hdr(3) >555.0_r_kind .and. hdr(3) <565.0_r_kind ) then
            iobsub=00
          else
            iobsub=01
          endif
        endif

```

emilyhcliu (changes requested):

ush/ioda/bufr2ioda/bufr2ioda_conventional_prepbufr_ps.py
nicholasesposito: @emilyhcliu The way I see it, we still need the initial part of the code to "filter out" other typ's/kx's and add 18/19 to them where typ/kx has those specific conditions because then the second part works on everything that remains.

nicholasesposito: At some point, it seems like it reverts back to 180/280 but I can't tell where. There is no 198/199/299 in the outputs. I can show you what I mean tomorrow.

emilyhcliu: @nicholasesposito the 199, 198, and 299 are for NSST only when user set id_drifter and id_ship to true in the namelist. For Non SST data set, the id_drifter and id_ship are set to false

nicholasesposito: @emilyhcliu Oh I think I see what you're saying. Let me comment them out and see what the new results would be.

nicholasesposito: Yep, you're right. I made the changes and checked them in

nicholasesposito: Yep, you're right. I made the changes and checked them in

ush/ioda/bufr2ioda/bufr2ioda_conventional_prepbufr_ps.py
```diff
@@ -658,6 +704,24 @@ def bufr_to_ioda(config, logger):
         .write_attr('long_name', 'Observation Type') \
         .write_data(typ_ps)
 
+    # ObsSubType - airTemperature
+    obsspace.create_var('obsSubType/airTemperature', dtype=obsSubType.dtype,
```
emilyhcliu: Shall we change `obsSubType` to `ObsSubType` since we have `ObsType`?

emilyhcliu (changes requested):

ush/ioda/bufr2ioda/bufr2ioda_conventional_prepbufr_ps.py
emilyhcliu: I checked the output and found that the variable type is int64.  We should use int32.  
```
group: ObsSubType {
  variables:
        int64 airTemperature(Location) ;
                airTemperature:_FillValue = 999999LL ;
                string airTemperature:long_name = "Observation SubType" ;
        int64 stationPressure(Location) ;
                stationPressure:_FillValue = 999999LL ;
                string stationPressure:long_name = "Observation SubType" ;
        int64 virtualTemperature(Location) ;
                virtualTemperature:_FillValue = 999999LL ;
                string virtualTemperature:long_name = "Observation SubType" ;
```

nicholasesposito: I'm guessing int64 should only be for dateTime?

nicholasesposito: Regardless, the code has been rewritten and the ObsSubType is now an int

nicholasesposito:
I'm reconverting to draft as there's still some more I want to do. I will re-publish once I'm content with it.

Still to do:
- Validate ObsError from conventional_ps
- Add ObsError to acft_profiles and validate.

Apologies for originally publishing too soon.

ush/ioda/bufr2ioda/bufr2ioda_conventional_prepbufr_ps.py
emilyhcliu: I tested the updated `conventional_ps` IODA file in the end-to-end test.  The IODA input file worked great!  
The number that passed QC comparison between GSI and JEDI is 85378 and 85387.   This is close enough.  👍

emilyhcliu (dismissed):
Changes look good!

ush/ioda/bufr2ioda/bufr2ioda_adpsfc_prepbufr.py
```diff
@@ -80,13 +80,16 @@ def bufr_to_ioda(config, logger):
     q.add('latitude', '*/YOB')
     q.add('longitude', '*/XOB')
     q.add('obsTimeMinusCycleTime', '*/DHR')
-    q.add('heightOfStation', '*/Z___INFO/Z__EVENT{1}/ZOB')
+    q.add('height', '*/Z___INFO/Z__EVENT{1}/ZOB')
     q.add('pressure', '*/P___INFO/P__EVENT{1}/POB')
 
-#   # Quality Infomation (Quality Indicator)
+    # Quality Marker
     q.add('qualityMarkerStationPressure', '*/P___INFO/P__EVENT{1}/PQM')
```
PraveenKumar-NOAA, line 97: @nicholasesposito any reason for changing QM variable names, like from pressureQM to qualityMarkerPressure?

ush/ioda/bufr2ioda/bufr2ioda_adpsfc_prepbufr.py
```diff
@@ -80,13 +80,16 @@ def bufr_to_ioda(config, logger):
     q.add('latitude', '*/YOB')
     q.add('longitude', '*/XOB')
     q.add('obsTimeMinusCycleTime', '*/DHR')
-    q.add('heightOfStation', '*/Z___INFO/Z__EVENT{1}/ZOB')
+    q.add('height', '*/Z___INFO/Z__EVENT{1}/ZOB')
     q.add('pressure', '*/P___INFO/P__EVENT{1}/POB')
 
-#   # Quality Infomation (Quality Indicator)
+    # Quality Marker
```
PraveenKumar-NOAA, line 96: @nicholasesposito same here.

ush/ioda/bufr2ioda/bufr2ioda_adpsfc_prepbufr.py
```diff
@@ -80,13 +80,16 @@ def bufr_to_ioda(config, logger):
...
     q.add('qualityMarkerStationPressure', '*/P___INFO/P__EVENT{1}/PQM')
     q.add('qualityMarkerStationElevation', '*/Z___INFO/Z__EVENT{1}/ZQM')
 
+    # ObsError
+    q.add('obsErrorStationPressure', '*/P___INFO/P__BACKG{1}/POE')
```
PraveenKumar-NOAA, line 101: @nicholasesposito for curiosity, we need to add ObsError to the converter?

nicholasesposito:
Quick update: it's taking a little longer than expected because some values are coming up as 0 when it should be "missing"

emilyhcliu:
The `obsTimeMinusCycleTime` should be read out as `float` 

https://github.com/NOAA-EMC/GDASApp/blob/32d4c9b01c11d920545475181b61556f63833250/ush/ioda/bufr2ioda/bufr2ioda_conventional_prepbufr_ps.py#L226

The [transformation of observation `dateTime` to  `obsTimeMinusCycleTime`](https://github.com/NOAA-EMC/GDASApp/blob/32d4c9b01c11d920545475181b61556f63833250/ush/ioda/bufr2ioda/bufr2ioda_conventional_prepbufr_ps.py#L24C1-L30C1) should use `float` and then convert it to int64. 

```
def Compute_dateTime(cycleTimeSinceEpoch, dhr):
    dhr = np.int64(dhr*3600)
    dateTime = dhr + cycleTimeSinceEpoch
    return dateTime
```

The `dateTime` values lose accuracy when they are converted from `obsTimeMinuscycleTime` with int64.  

Suggestions:

```
    dhrorig1 = t.get('obsTimeMinusCycleTime', type='float') 
     :
    dhrorig2 = u.get('obsTimeMinusCycleTime', type='float')
     :   
    dhr1 = np.array([], dtype=np.float32)
     :
    dhr2 = np.array([], dtype=np.float32)
     :
    dhr3 = v.get('timeOffset', 'prepbufrDataLevelCategory', type='float') 
     :
    dhr = ma.concatenate((dhr1, dhr2, dhr3), axis=0).astype(np.int64)
    dhr = ma.masked_values(dhr, np.iinfo(np.int64).max)
```


```
def Compute_dateTime(cycleTimeSinceEpoch, dhr):
    dhr = dhr * 3600.
    dateTime = np.int64(dhr) + cycleTimeSinceEpoch
    return dateTime
```

emilyhcliu:
The accuracy in dateTime is necessary.  Some QC procedures use dateTime difference between observations.  

Here is the example of the error inflation factors calculation using the dateTime difference between observations with **less dateTime precision**:

dupfact_hist_noqc
dupfact_vs_ps_scatter_noqc

Error inflation factor calculation with **accurate dateTime precision**
GSI (red) vs. JEDI (blue) ----> They are on top of each other (identical) so the histogram color bars become purple.
dupfact_hist_qc
dupfact_vs_ps_scatter_qc

nicholasesposito:
I had major issues with dateTime running too long and Orion cancelling the job, but I figured out a faster way to run it. Orion no longer has issues. 

That being said, I think all the changes we all wanted are in the code. Please look over. I'll be moving it from draft to in review shortly.

CoryMartin-NOAA:
@emilyhcliu @nicholasesposito is this ready to review?

nicholasesposito:
@CoryMartin-NOAA  for me it is

CoryMartin-NOAA (approved):

emilyhcliu:
> @emilyhcliu @nicholasesposito is this ready to review?

I reviewed the changes.  I am running the conventional ps test again with the updated converter from this PR.  The `atmanlrun` is sitting in the queue.  I will approve this PR as soon as I confirm the end-to-end result is good.

PraveenKumar-NOAA (approved):
Looks good.

ush/ioda/bufr2ioda/bufr2ioda_adpsfc_prepbufr.py
nicholasesposito: Partially for consistency. But also in the prep code there's certain variables they call quality Information that eventually get converted to qualityMarker's so I"m trying to make sure I don't get confused in the future

ush/ioda/bufr2ioda/bufr2ioda_adpsfc_prepbufr.py
nicholasesposito: Same as above

ush/ioda/bufr2ioda/bufr2ioda_adpsfc_prepbufr.py
nicholasesposito: No, most of the time it gets overwritten. I need it for my ADPSFC/SFCSHP where it will eventually get overwritten but it was needed to run some of the end-to-end tests so I was told to put it in and if it gets overwritten then so  be it.

emilyhcliu:
@CoryMartin-NOAA and @RussTreadon-NOAA I re-ran conventional Ps assimilation using the conventional_ps IODA generated from this PR.  The[ results](https://github.com/NOAA-EMC/GDASApp/pull/803#issuecomment-1851387894) are identical to my previous test with my own fix to the [dateTime issue.](https://github.com/NOAA-EMC/GDASApp/pull/792#issuecomment-1852442578)  
So, this PR is good to merge.  We will test the aircraft when the valid observation filter is ready.

emilyhcliu (approved):
Tested and it is good.
This PR is ready to merge!

nicholasesposito:
Perfect. Thanks everyone.

PR #797, opened 2023-12-06
Add some error handling to marine obs prep task

AndrewEichmann-NOAA:
Some error handling to the obs prep task:
- fails if the `OBS_YAML` file is incorrect
- prints a warning and skips to the next observation source if the yaml for that particular source is wrong, or there are no files in `DMPDIR` to copy and convert

`print` statements used instead of `logging` due to `logging` not working for some reason


apchoiCMD (dismissed):
Look good to me

ShastriPaturi (dismissed):
Looks edible.

guillaumevernieres:
Since you're starting to look into the "cosmetics" @AndrewEichmann-NOAA , I figured it might be time to get some feedback from @aerorahul and @CoryMartin-NOAA .

scripts/exglobal_prep_ocean_obs.py
```diff
@@ -24,30 +25,51 @@
...
+try:
+    # For each of the observation sources (observers) specificed in the OBS_YAML...
+    for observer in obsConfig['observers']:
 
-    obsSpaceName = observer['obs space']['name']
-    print(f"obsSpaceName: {obsSpaceName}")
+        try:
+            obsSpaceName = observer['obs space']['name']
+        except KeyError:
+            print(f"observer: {observer}")
```
CoryMartin-NOAA, line 44: Why was logging not working? It looks like it gets imported just fine?

scripts/exglobal_prep_ocean_obs.py
```diff
@@ -24,30 +25,51 @@
...
+            print("WARNING: Ill-formed observer yaml file, skipping")
+            continue  # to next observer
 
-    for observation in obsprocConfig['observations']:
+        print(f"obsSpaceName: {obsSpaceName}")
```
CoryMartin-NOAA: Shouldn't this be inside the try? The exception won't define this variable and then it will fail...

scripts/exglobal_prep_ocean_obs.py
```diff
@@ -24,30 +25,51 @@
...
+# ...look through the observations in OBSPROC_YAML...
+        for observation in obsprocConfig['observations']:
```
CoryMartin-NOAA, line 49: Isn't there a better way than looping through a second list and finding a match? Like a key, value pair dict?

scripts/exglobal_prep_ocean_obs.py
```diff
@@ -24,30 +25,51 @@
...
+                obsprocSpace['input files'] = matchingFiles
+                obsprocSpace['window begin'] = windowBegin
+                obsprocSpace['window end'] = windowEnd
+
+                iodaYamlFilename = obsprocSpaceName + '2ioda.yaml'
+                save_as_yaml(obsprocSpace, iodaYamlFilename)
+
+                subprocess.run([OCNOBS2IODAEXEC, iodaYamlFilename], check=True)
+except TypeError:
+    sys.exit("CRITICAL: Ill-formed OBS_YAML file, exiting")
```
CoryMartin-NOAA: raise an exception instead of `sys.exit`?

ush/soca/prep_marine_obs.py
```diff
@@ -11,7 +11,7 @@
 CDUMP = os.getenv('CDUMP')
 COMIN_OBS = os.getenv('COMIN_OBS')
 
-cycdir = os.path.join(DMPDIR, CDUMP + '.' + str(PDY), str(cyc))
+cycDir = os.path.join(DMPDIR, CDUMP + '.' + str(PDY), str(cyc))
```
CoryMartin-NOAA: `RUN` not `CDUMP` right?

scripts/exglobal_prep_ocean_obs.py
AndrewEichmann-NOAA: it imports (I should take that out), but when you try to use it nothing shows up in the output

scripts/exglobal_prep_ocean_obs.py
AndrewEichmann-NOAA: This gets the `observer` from `OBS_YAML`, which looks like this:
```
observers:
- !INC ${OBS_YAML_DIR}/adt_all.yaml
- !INC ${OBS_YAML_DIR}/adt_j3.yaml
```
This `try` fails when the individual observer yaml file can't provide `observer['obs space']['name']`, and then moves on to the next one

CoryMartin-NOAA: what if `obsSpaceName` is not yet defined?

scripts/exglobal_prep_ocean_obs.py
```diff
@@ -6,8 +6,10 @@
 import os
 import prep_marine_obs
 import subprocess
+import sys
```
CoryMartin-NOAA: no more `sys` needed

CoryMartin-NOAA (approved):

guillaumevernieres:
is this still a draft @AndrewEichmann-NOAA ?

guillaumevernieres (approved):
:+1:

PR #800, opened 2023-12-07
Fix example GDAS OOPS utility

CoryMartin-NOAA:
Time Window issues strike again!

I haven't tested yet because I think fv3-jedi also needs merged now...

### DIFF
--- a/test/land/letkfoi_land.yaml
+++ b/test/land/letkfoi_land.yaml
@@ -164,5 +164,6 @@ observations:
 output increment:
   filename_sfcd: xainc.sfc_data.nc
   filetype: fms restart
-window begin: '2021-03-23T15:00:00Z'
-window length: PT6H
+time window:
+  begin: '2021-03-23T15:00:00Z'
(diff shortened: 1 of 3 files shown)


CoryMartin-NOAA:
FV3-JEDI was merged 5 mins ago, we should try to build now.

RussTreadon-NOAA (dismissed):
Looks right.  Caveat:  I say this as an inexperienced c++ programmer.  

Approve.

AndrewEichmann-NOAA (dismissed):
Let's give it a whirl

ShastriPaturi (dismissed):

ShastriPaturi (approved):

RussTreadon-NOAA:
@CoryMartin-NOAA , with this PR now in `develop` is it time to update `feature/gdas-validation` with the current head of `develop`?

CoryMartin-NOAA:
@RussTreadon-NOAA probably yes, but maybe double check with others since that would require a recompile of code

PR #802, opened 2023-12-11
Fix UFO nightly testing script for time window change

CoryMartin-NOAA:
As the title says. Darth Vader - NOAA Affiliate was not happy over the weekend, this should fix that.


RussTreadon-NOAA (approved):
Haven't tested the proposed change to `ush/ufoeval/run_ufo_hofx_test.sh` but it looks correct.

Approve.

PR #816, opened 2023-12-19
make ctest dmpdir more like real life

AndrewEichmann-NOAA:
This makes the ctest DMPDIR for observation files more like the DMPDIR used in model runs, in order to provide for closer-to-life ctests, and in anticipation of the R2D2-ectomy.


apchoiCMD (approved):
Looks more real and convenient

guillaumevernieres (approved):
Good enough for now, but at some point we'll need to use the correct obs filenames.

ShastriPaturi (approved):
Fine with me.

PR #819, opened 2023-12-19
Copy prepped marine obs to ROTDIR

AndrewEichmann-NOAA:
The obs prepped and converted to IODA are copied to `ROTDIR` under obs. Also, the order of `JGDAS_GLOBAL_OCEAN_ANALYSIS_PREP` and `JGLOBAL_PREP_OCEAN_OBS` tasks are swapped , as `JGDAS_GLOBAL_OCEAN_ANALYSIS_PREP` creates the target directory.


AndrewEichmann-NOAA:
The CI tests failed with the vrfy, but also because I changed a script from the ioda converter ctests that I repurposed for the obs prep ctest. The latter should work now

apchoiCMD:
> The CI tests failed with the vrfy, but also because I changed a script from the ioda converter ctests that I repurposed for the obs prep ctest. The latter should work now

Yes, the log shows that ctest's inputs are not properly prepared. thanks @AndrewEichmann-NOAA

guillaumevernieres:
Why do we have 2 `prepdata.sh` @AndrewEichmann-NOAA ?

AndrewEichmann-NOAA:
> Why do we have 2 `prepdata.sh` @AndrewEichmann-NOAA ?

@guillaumevernieres One works for the gdas-utils IODA converter ctests, the other for the `JGLOBAL_PREP_OCEAN_OBS` ctests. I originally had the g-w ctests use the same script as the gdas-utils tests. That the two are mostly identical is, for the moment, partially a matter of laziness, but when I get into having `JGLOBAL_PREP_OCEAN_OBS` work on obs that are supposed to work for the rest of the global-workflow task ctests, it will work on a different set of observation files.

guillaumevernieres:
Is this ready to review @AndrewEichmann-NOAA ?

AndrewEichmann-NOAA:
@guillaumevernieres submitted!

ShastriPaturi (approved):
Thanks

guillaumevernieres (approved):
:+1:

PR #822, opened 2023-12-21
Bugfix for CI and linking loop

CoryMartin-NOAA:
This PR adds in changes to clone recursively in CI, plus a fix to not create infinite loops in symlinks (still not clear how it happens but a `rm` should do the trick)

Note the stable driver script will be heavily modified as part of the switch to submodules.


build.sh
```diff
@@ -103,6 +103,7 @@ CMAKE_OPTS+=" -DWORKFLOW_TESTS=${WORKFLOW_BUILD}"
 
 # JCSDA changed test data things, need to make a dummy CRTM directory
 if [[ $BUILD_TARGET == 'hera' ]]; then
+  if [ -d "$dir_root/test-data-release/" ]; then rm -rf $dir_root/test-data-release/; fi
```
CoryMartin-NOAA, line 106: This is the offending section with the crtm link, I don't understand why the ln points to itself

RussTreadon-NOAA (approved):
Changes look correct to me.

Approve.

PR #803, opened 2023-12-12
Update observation YAMLs for conventional surface pressure

emilyhcliu:
This PR includes the following updates for conventional surface pressure
1. Add observation YAML under configuration
2. Update observation YAML under testing
3. Add conv_ps to 3dvar obs list
4. Set window shift to true (time boundary is inclusive)
5. Validate 3dvar for surface (station) pressure (end-to-end)
     - [QC validation](https://github.com/NOAA-EMC/GDASApp/pull/803#issuecomment-1851387894)
     - [Increments](https://github.com/NOAA-EMC/GDASApp/pull/803#issuecomment-1852496937)

See [PR #792 ](https://github.com/NOAA-EMC/GDASApp/pull/792) for the BUFR converter work to convert the following surface pressure observation into IODA:
- SFC - 181, 187
- SFCSHIP - 180, 183
- SONDE - 120

The related UFO evaluation  (using GSI obs and geovals) results for surface pressure are documented [here](https://docs.google.com/presentation/d/1OvVMYOutxVyk0GE93Jzoh3hQVoj-dCgrwsVWDDlPT1s/edit?usp=sharing).

PR #824, opened 2023-12-21
Update observation YAMLs for conventional surface pressure

CoryMartin-NOAA:
Same as #803  but into develop!


RussTreadon-NOAA (approved):
Since the changes were approved for merger into `feature/gdas-validation`, they should be ready to go into `develop`.  I haven't tested the conventional ps yamls myself but trust the work of others.

Approve.

PR #834, opened 2024-01-04
Fix missing datetime "unit" and Ghrsst2ioda output "K" to "C"

### LINKED ISSUE #830: The Ghrsst2Ioda converter should output C instead of K

### LINKED ISSUE #831: Missing datetime "units"
A residual bug from an mpi implementation ... Ask @AndrewEichmann-NOAA for the fix

apchoiCMD:
This PR fixes two bugs:

1. Missing datetime "units"
2. Ghrsst2ioda converter outputs sea surface temperature with unit "C"

All outputs are placed at each issue at https://github.com/NOAA-EMC/GDASApp/issues/831 and https://github.com/NOAA-EMC/GDASApp/issues/830

Close https://github.com/NOAA-EMC/GDASApp/issues/831 , Close https://github.com/NOAA-EMC/GDASApp/issues/830

### DIFF
--- a/utils/obsproc/Ghrsst2Ioda.h
+++ b/utils/obsproc/Ghrsst2Ioda.h
@@ -36,10 +36,6 @@ namespace gdasapp {
       fullConfig_.get("bounds.min", sstMin);
       float sstMax;
       fullConfig_.get("bounds.max", sstMax);
-      if ( sstUnits == "C" ) {
-        sstMin += 0.0;
-        sstMax += 0.0;
-      }
 
       // Open the NetCDF file in read-only mode
       netCDF::NcFile ncFile(fileName, netCDF::NcFile::read);
--- a/utils/obsproc/Ghrsst2Ioda.h
+++ b/utils/obsproc/Ghrsst2Ioda.h
@@ -37,8 +37,8 @@ namespace gdasapp {
       float sstMax;
       fullConfig_.get("bounds.max", sstMax);
       if ( sstUnits == "C" ) {
-        sstMin += 273.15;
-        sstMax += 273.15;
+        sstMin += 0.0;
+        sstMax += 0.0;
       }
 
       // Open the NetCDF file in read-only mode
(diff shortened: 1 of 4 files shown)


utils/obsproc/Ghrsst2Ioda.h
```diff
@@ -37,8 +37,8 @@ namespace gdasapp {
       float sstMax;
       fullConfig_.get("bounds.max", sstMax);
       if ( sstUnits == "C" ) {
```
guillaumevernieres: Keep the condition below as it was originally but replace "C" with "K". No point doing `+=0` ...
Or just remove the option and just assume we only deal with C.

ShastriPaturi (dismissed):
Looks good to me.

AndrewEichmann-NOAA (dismissed):
Looks good

guillaumevernieres (commented):
We should probably entirely remove the option of outputting K and just assume everything is in C.

utils/obsproc/Ghrsst2Ioda.h
apchoiCMD: > Keep the condition below as it was originally but replace "C" with "K". No point doing `+=0` ... Or just remove the option and just assume we only deal with C.

@guillaumevernieres I will push again with your comment.

guillaumevernieres (commented):
[that line](https://github.com/NOAA-EMC/GDASApp/blob/a34773ad8f80d2074e0528caae97d70d2f23afd7/utils/test/testinput/gdas_ghrsst2ioda.yaml#L8)  is not needed anymore.

apchoiCMD:
> [that line](https://github.com/NOAA-EMC/GDASApp/blob/a34773ad8f80d2074e0528caae97d70d2f23afd7/utils/test/testinput/gdas_ghrsst2ioda.yaml#L8) is not needed anymore.

ha! thanks for heading me up! Pushing again...

guillaumevernieres (approved):
Thanks @apchoiCMD

PR #836, opened 2024-01-04
Remove tests with R2D2 and anything referencing it

AndrewEichmann-NOAA:
Removes soca ctests related to R2D2, associated data, and any lines in scripts, yamls, or other config files referencing it. SOCA ctests now run on obs prepped by the prep obs task.

Almost takes care of https://github.com/NOAA-EMC/GDASApp/issues/815 as references are still present in global-workflow task config files.

`logging.info` has been replaced by `print` because logging hasn't been working


scripts/exgdas_global_marine_analysis_prep.py
```diff
@@ -90,7 +90,7 @@ def cice_hist2fms(input_filename, output_filename):
     ds = xr.open_dataset(input_filename)
 
     if 'aicen' in ds.variables and 'hicen' in ds.variables and 'hsnon' in ds.variables:
-        logging.info(f"*** Already reformatted, skipping.")
+        print(f"*** Already reformatted, skipping.")
```
CoryMartin-NOAA: why are all these logging replaced with print?

guillaumevernieres: see PR description @CoryMartin-NOAA .

CoryMartin-NOAA: Ok but why? as it presumably was working before? I'd prefer to find the reason.

AndrewEichmann-NOAA: @CoryMartin-NOAA IIRC there was speculation that it was something coming down from global-workflow - we haven't figured it out and it's easy to change back to logging when we do

guillaumevernieres: I think we all agree on that @CoryMartin-NOAA ! @AndrewEichmann-NOAA , can you revert your `prints` back to `loggings` and submit an issue? I thought we had an issue describing this problem already, but I can't find it.

guillaumevernieres:
FYI, here's the `ci` result summary:
```
The following tests FAILED:                                                                                                                                       
        1688 - test_gdasapp_soca_JGDAS_GLOBAL_OCEAN_ANALYSIS_VRFY (Failed)                                                                                        
        1699 - test_gdasapp_store_gsi_satbias (Not Run)                                                                                                           
Errors while running CTest   
```

The verify job fails in `eva`, which was pointed out by @RussTreadon-NOAA and @CoryMartin-NOAA a few days ago. I'm not sure what the `satbias` job does, is that a left over from the r2d2 days @RussTreadon-NOAA and @CoryMartin-NOAA ?

test/atm/test_store_gsi_satbias.py
CoryMartin-NOAA, line 1: The cmake needs updated to remove this test now that the script is removed

guillaumevernieres:
@AndrewEichmann-NOAA : @CoryMartin-NOAA says `test_gdasapp_store_gsi_satbias` can be :bomb:

test/atm/test_store_gsi_satbias.py
AndrewEichmann-NOAA: yes, I neglected to do that

AndrewEichmann-NOAA: I've removed test_gdasapp_store_gsi_satbias but others such as  test_gdasapp_land_letkfoi_snowda are failing - is that a concern? @CoryMartin-NOAA @guillaumevernieres

AndrewEichmann-NOAA:
@guillaumevernieres BTW the analysis is still rejecting all the `sst_metopb_l3u_so025` obs

scripts/exgdas_global_marine_analysis_prep.py
AndrewEichmann-NOAA: done

guillaumevernieres (dismissed):
Thanks @AndrewEichmann-NOAA

test/atm/test_store_gsi_satbias.py
guillaumevernieres: it looks like it ran fine in the `CI` @AndrewEichmann-NOAA , so all good!

CoryMartin-NOAA:
Can we remove it here: https://github.com/NOAA-EMC/GDASApp/blob/bd5701221379c6d55a5197b633289e7280c47fa1/.github/workflows/unittests.yaml#L18

AndrewEichmann-NOAA:
> Can we remove it here:
> 
> https://github.com/NOAA-EMC/GDASApp/blob/bd5701221379c6d55a5197b633289e7280c47fa1/.github/workflows/unittests.yaml#L18

@CoryMartin-NOAA The whole `- name:` section, or just the three lines with "r2d2"?

CoryMartin-NOAA:
Do we have any need for solo ? I don't think so, right?

ShastriPaturi (dismissed):
Will test with my in situ obs.
Approving it for now.

guillaumevernieres:
> Do we have any need for solo ? I don't think so, right?

no need for solo anymore @CoryMartin-NOAA .

guillaumevernieres:
> > Can we remove it here:
> > https://github.com/NOAA-EMC/GDASApp/blob/bd5701221379c6d55a5197b633289e7280c47fa1/.github/workflows/unittests.yaml#L18
> 
> @CoryMartin-NOAA The whole `- name:` section, or just the three lines with "r2d2"?

The entire section @AndrewEichmann-NOAA ... I'm 99.3% sure we don't use solo outside of the r2d2 implementation ...

AndrewEichmann-NOAA:
@CoryMartin-NOAA @guillaumevernieres ok, removed, let's see what happens

AndrewEichmann-NOAA:
`test_gdasapp_soca_JGDAS_GLOBAL_OCEAN_ANALYSIS_VRFY` fails, `test_gdasapp_atm_jjob_ens_run` times out, and `test_gdasapp_atm_jjob_ens_final` fails - the vrfy is a known issue, what of the other two?

guillaumevernieres:
> `test_gdasapp_soca_JGDAS_GLOBAL_OCEAN_ANALYSIS_VRFY` fails, `test_gdasapp_atm_jjob_ens_run` times out, and `test_gdasapp_atm_jjob_ens_final` fails - the vrfy is a known issue, what of the other two?

I suspect the `test_gdasapp_atm_jjob_ens_final`  fails because of `test_gdasapp_atm_jjob_ens_run` and is probably unrelated to your changes ... So maybe we can ignore? @RussTreadon-NOAA , @CoryMartin-NOAA ?

guillaumevernieres (commented):
Looks good @AndrewEichmann-NOAA , do we have the dependencies sorted out properly in the `g-w`? If not it needs to be done asap, since merging this would break the cycling.

RussTreadon-NOAA:
Yes, @guillaumevernieres, you are correct.  `test_gdasapp_atm_jjob_ens_final` processes output from `test_gdasapp_atm_jjob_ens_run`.  If the ens or var `run` job fails, the corresponding `final` job will fail.  The `ens_run` job executes `fv3jedi_letkf.x` with 36 tasks.  Depending on the system load, the queue wait time can exceed the ctest timeout limit.

AndrewEichmann-NOAA:
@guillaumevernieres The changes to g-w depend on this PR and I don't think vice versa, so once this is in I can change g-w. I think it's just a matter of removing assignments from config files.

guillaumevernieres (approved):
Approving but lets **not merge** just quite yet until we have an understanding of how close we are with the corresponding `g-w` updates.
Thanks for doing this @AndrewEichmann-NOAA .

PR #863, opened 2024-01-18
Fix the logging

### LINKED ISSUE #853: Logging for the marine tasks isn't working
Time to look at this issue. It's making it impossible to debug.

AndrewEichmann-NOAA:
The use of the standard Python logging has, for some reason, has not been working: nothing apparently gets sent to stdout. This PR replaces the Python logging with wxflow Logger. Some configuration previously used is perhaps lost, but this is a 90% solution. 

Resolves https://github.com/NOAA-EMC/GDASApp/issues/853

### DIFF
--- a/scripts/exgdas_global_marine_analysis_post.py
+++ b/scripts/exgdas_global_marine_analysis_post.py
@@ -21,7 +21,6 @@
 import os
 import glob
 import shutil
-import logging
 from datetime import datetime, timedelta
 from wxflow import FileHandler, Logger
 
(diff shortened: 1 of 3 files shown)


apchoiCMD (dismissed):

ShastriPaturi (dismissed):

scripts/exgdas_global_marine_analysis_post.py
```diff
@@ -23,7 +23,9 @@
 import shutil
 import logging
```
CoryMartin-NOAA: do we need `logging` here still? Or can it be removed like the others?

CoryMartin-NOAA (commented):
One minor comment, otherwise looks good

scripts/exgdas_global_marine_analysis_post.py
AndrewEichmann-NOAA: yes, good catch

CoryMartin-NOAA (approved):

PR #875, opened 2024-01-25
Changes to get snow DA test working

CoryMartin-NOAA:
This is to get the first snow DA global-workflow test working. Subsequent PRs will be needed when the test is updated but this gets a test started.


jiaruidong2017 (approved):
Thanks for your efforts.

PR #913, opened 2024-02-08
Changing 'land' to 'snow' to better reflect analysis jobs

CoryMartin-NOAA:
This PR changes 'land' everywhere to 'snow' to differentiate that this is snow DA and not land DA. There will need to be a corresponding G-W PR to do this in the workflow.


ush/ufsda/yamltools.py
YoulongXia-NOAA, line 1: snow -> SNOW?

RussTreadon-NOAA (dismissed):
Looks reasonable to me.  

Approve pending success from CI tests

ush/ufsda/yamltools.py
```diff
@@ -206,8 +206,8 @@ def calc_time_vars(config):
     bkg_time_obj = valid_time_obj
     config['BKG_YYYYmmddHHMMSS'] = bkg_time_obj.strftime('%Y%m%d.%H%M%S')
     config['BKG_ISOTIME'] = bkg_time_obj.strftime('%Y-%m-%dT%H:%M:%SZ')
-    config['LAND_BKG_YYYYmmddHHMMSS'] = bkg_time_obj.strftime('%Y%m%d.%H%M%S')
-    config['LAND_BKG_ISOTIME'] = bkg_time_obj.strftime('%Y-%m-%dT%H:%M:%SZ')
+    config['snow_BKG_YYYYmmddHHMMSS'] = bkg_time_obj.strftime('%Y%m%d.%H%M%S')
+    config['snow_BKG_ISOTIME'] = bkg_time_obj.strftime('%Y-%m-%dT%H:%M:%SZ')
```
YoulongXia-NOAA: snow_BKG_YYYYmmddHHMMSS -> SNOW_BKG_YYYYmmddHHMMSS, as LAND is capital, I am not sure. The same case is for snow_BKG_ISOTIME -> SNOW_BKG_ISOTIME.

YoulongXia-NOAA: Others looks fine with me.

YoulongXia-NOAA (dismissed):
If you think lower case snow is fine, it looks good to me.

ush/ufsda/yamltools.py
CoryMartin-NOAA: Thanks @YoulongXia-NOAA for catching my mistake!

jiaruidong2017:
If we consider the soil DA is added in the future, I would suggest to keep land directories at: `${PROJECT_SOURCE_DIR}/test/land/` and `${PROJECT_SOURCE_DIR}/ush/land/`.

parm/snow/letkfoi/letkfoi.yaml
```diff
@@ -54,7 +54,7 @@ local ensemble DA:
 
 output increment:
   datapath: $(DATA)/anl
-  prefix: landinc
+  prefix: snowinc
```
yuanxue2870, line 57: I assume we will introduce "soilinc" later, correct?

CoryMartin-NOAA: My logic here, and it is up for discussion, is that soil DA is strongly coupled with the atmosphere, so it would need a more holistic solution when that is ready in JEDI. This snow DA is a separate analysis (for now), hence why everything is defined here as 'snow' specifically

jiaruidong2017:
@CoryMartin-NOAA Need to update `sorc/land-jediincr/sorc/apply_incr_noahmp_snow.f90` by changing from `landinc` to `snowinc`.

parm/snow/letkfoi/letkfoi.yaml
yuanxue2870: Got it. Thanks.

CoryMartin-NOAA:
Thanks @jiaruidong2017 , I was trying to debug but that is the problem.

CoryMartin-NOAA:
Before making the changes to the Fortran increment code, I want to make sure @ClaraDraper-NOAA @barlage and others are okay with the name change from `land` to `snow`

jiaruidong2017:
I am fine with the name change from `landinc` to `snowinc`, because it is the snow specific in `apply_incr_noahmp_snow.f90`.

barlage:
Fine with me, I'm always in favor of more descriptive naming.

ClaraDraper-NOAA (dismissed):
looks good to me @CoryMartin-NOAA. It's kinda annoying to have to do, so thanks for addressing this now. I think I labelled everything in my global_workflow PR as "gsi_soil", but I will go back and double-check this.

parm/snow/letkfoi/letkfoi.yaml
ClaraDraper-NOAA: I agree with Cory. If all goes to plan, the soil DA from JEDI will be strongly coupled, so there will not be a separate increment file. Until we have a single land DA update, which is 5-10 years away, having clearly named files will be really helpful.

CoryMartin-NOAA:
Thanks all, let me make the change in the fortran code, issue a PR there, then update the hash as part of this PR once that is merged.

ClaraDraper-NOAA (approved):

YoulongXia-NOAA (approved):
I looked at the change. It is good to me. I approve this PR.

jiaruidong2017 (approved):
Approve pending on success from CI test

yuanxue2870 (approved):

jiaruidong2017:
The four failed tests seem caused by trying to submit batch jobs.

CoryMartin-NOAA:
The failed tests are known to fail outside of this PR, I will merge. Thanks!

PR #920, opened 2024-02-15
Change IMS prep path to point to GDAS fix directory

CoryMartin-NOAA:
The IMS remapping files should not be in COM_OBS but in FIXgdas, now that they are stored there and maintained by the global-workflow team. This PR has the workflow copy these files from that place now instead. 

Note, this is one in many PRs in the next few weeks to refactor the snow/land DA.


guillaumevernieres (approved):
:+1:

jiaruidong2017 (commented):
@CoryMartin-NOAA do we need to add `gdas_obs_ver` as below?

`- [{{ FIXgdas }}/obs/gdas_obs_ver/ims/IMS_4km_to_C{{ CASE }}.mx{{ OCNRES }}.nc`

jiaruidong2017:
@CoryMartin-NOAA Forget my last comment, because the `[{{ FIXgdas }}/obs` already include the `gdas_obs_ver`.

PR #923, opened 2024-02-16
Update aerosol YAMLs to support increment write updates

CoryMartin-NOAA:
JEDI recently changed the YAML format for the increment write to require a `state component` subsection. Without this, the code will lead to a vague segmentation fault.


andytangborn (approved):
Looks like just one line added, and some indentation.

PR #934, opened 2024-02-22
Add 'bundle' directory and make Hera not download CRTM coeffs

CoryMartin-NOAA:
What the title says. Ecbuild should now link from sorc/ to bundle/ and not complain about unable to link a directory


.gitignore
```diff
@@ -182,4 +182,28 @@ install*/
...
+/bundle/jedicmake
+/bundle/land-imsproc
+/bundle/land-jediincr
+/bundle/mom6
+/bundle/oops
+/bundle/saber
+/bundle/soca
+/bundle/test-data-release/
+/bundle/ufo
+/bundle/vader
```
aerorahul: Is `bundle` not sufficient?

aerorahul (dismissed):
Looks good. 
Thanks for taking care of both.

.gitignore
CoryMartin-NOAA: we want `/bundle/CMakeLists.txt` :-)

aerorahul: :-)

One can add `bundle/.gitignore` and put
```
*
!CMakeLists.txt
```
in it.

or in the top-level `.gitignore`

```
bundle/*
!bundle/CMakeLists.txt
```

CoryMartin-NOAA: I didn't know that ! trick, that's nice! Thanks!

aerorahul (approved):
looks good.

guillaumevernieres (approved):

DavidNew-NOAA (approved):

PR #942, opened 2024-02-28
Fix bug in gdas_meanioda.h

DavidNew-NOAA:
This fixes a bug in gdas_meanioda.h resulting from a change ([#1186](https://github.com/JCSDA-internal/ioda/pull/1186)) to the ioda::ObsSpace constructor which now takes a eckit::LocalConfiguration object as input rather than ioda::ObsTopLevelParameters object for the obs space parameters. It also updates most of the JEDI submodules.


CoryMartin-NOAA (approved):

guillaumevernieres (approved):
:+1:

Issue #912, opened 2024-02-08
Recenter ens. members around the deterministic background.


guillaumevernieres:
The code/application exists (under utils) but is not implemented in the workflow yet. We need to pass recentering increments to the ensemble members.

guillaumevernieres:
Notes:
1 - We only need the initial conditions (MOM.res.nc and cice.res.nc), not need to stage f006 & f009
2 - Create `soca_recenter.yaml` based off of `soca_ensb.yaml`
3 - We need to turn off most of the options of the re-centering ... (@guillaumevernieres 's job)
4 - move the ensemble of increment to the comroot/model_data

AndrewEichmann-NOAA:
in [feature/ocnanalecen](https://github.com/NOAA-EMC/GDASApp/tree/feature/ocnanalecen)

AndrewEichmann-NOAA:
@guillaumevernieres Which of I presume the following should be copied back?
```
ocn.ssh_recentering_error.incr.2018-04-15T09:00:00Z.nc
ice.ssh_recentering_error.incr.2018-04-15T09:00:00Z.nc
ocn.steric_explained_variance.incr.2018-04-15T09:00:00Z.nc
ocn.ssh_unbal_stddev.incr.2018-04-15T09:00:00Z.nc
ocn.ssh_total_stddev.incr.2018-04-15T09:00:00Z.nc
ocn.ssh_steric_stddev.incr.2018-04-15T09:00:00Z.nc
ocn.pert.steric.4.nc
ocn.pert.steric.3.nc
ocn.pert.steric.2.nc
ocn.pert.steric.1.nc
ocn.bkgerr_stddev.incr.2018-04-15T09:00:00Z.nc
ice.trash.incr.2018-04-15T09:00:00Z.nc
ice.bkgerr_stddev.incr.2018-04-15T09:00:00Z.nc
```

guillaumevernieres:
I forgot to change the name of the increments ... and we're not outputting sea-ice.
We should copy the ocn.pert.steric*, but they also should be name differently.

AndrewEichmann-NOAA:
@CoryMartin-NOAA Do you know what the increment files should be called when copied back to COM after recentering? Started with (for example) `enkfgdas.20180415/12/mem001/model_data/ocean/restart/gdas.ocean.t12z.incr.nc`

CoryMartin-NOAA:
[...]

PR #946, opened 2024-02-29
Move bkg renamer to function

AndrewEichmann-NOAA:
Moves bit of code to rename background files so that it can be used for centering task

Works towards addressing https://github.com/NOAA-EMC/GDASApp/issues/912

### DIFF
--- a/ush/soca/bkg_utils.py
+++ b/ush/soca/bkg_utils.py
@@ -168,7 +168,7 @@ def rename_bkg(bkg_dir, anl_dir, RUN):
     mom_ic_src = glob.glob(os.path.join(bkg_dir, f'{RUN}.ocean.*.inst.f003.nc'))[0]
     mom_ic_dst = os.path.join(anl_dir, 'INPUT', 'MOM.res.nc')
     ics_list.append([mom_ic_src, mom_ic_dst])
-   
+
     # seaice IC's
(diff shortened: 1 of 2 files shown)


ush/soca/bkg_utils.py
```diff
@@ -157,3 +158,19 @@ def gen_bkg_list(bkg_path, out_path, window_begin=' ', yaml_name='bkg.yaml', ice
...
     # copy ocean backgrounds to RUNDIR
     FileHandler({'copy': bkg_list_src_dst}).sync()
+
+
+def rename_bkg(bkg_dir, anl_dir, RUN):
+
+    # Copy and rename initial condition
+    ics_list = []
+    # ocean IC's
+    mom_ic_src = glob.glob(os.path.join(bkg_dir, f'{RUN}.ocean.*.inst.f003.nc'))[0]
```
CoryMartin-NOAA: why are we globbing here?

AndrewEichmann-NOAA: @guillaumevernieres ?

CoryMartin-NOAA: Does this file name ever vary? and if so, is it predictable? like is it date/time based? If this is a crutch to not use 't06z' then we need to include that here.

CoryMartin-NOAA (approved):

guillaumevernieres (approved):
:+1:

PR #953, opened 2024-03-06
Fix averaging datetime on geostationary sst's

### LINKED ISSUE #905: Check the binning of the geostationary sst's
Our binning works on the L3 uncollated grid but I have no idea if this is valid for the "collated" product, which is what we use for the geostationary instruments. I presume it's fine, but we should probably check!

Just plotting the input and output binned sst field would be sufficient.

Example here:
```/scratch1/NCEPDEV/stmp2/Guillaume.Vernieres/runs/gw-ci-c384/test/COMROOT/test/gdas.20210702/00/obs```
look for ```gdas.t00z.sst_abi_g16_l3c.2021070200.nc4``` and the corresponding yaml file for the input.

### PR DESCRIPTION
**Pull request by apchoiCMD**
#### Description
- Implemented an averaging datetime outside of superob other than obs

#### Problem detected
- Superobing datetime within ioda converter doesn't makes datetime a sense compared to original datetime

- Resolves #905

### CODE REVIEW

File: utils/obsproc/Ghrsst2Ioda.h [hunk 1]
```diff
@@ -157,21 +157,33 @@ namespace gdasapp {
...
         sst_s = sst;
         lon2d_s = lon2d;
         lat2d_s = lat2d;
         obserror_s = obserror;
-        seconds_s = seconds;
       }
 
+      // Non-Superobing part only applied to datetime
+      // TODO(Mindo): Think superob datetime efficiently
+      int64_t sum = 0;
```
guillaumevernieres, line 176: Can't you use ```std::accumulate``` to compute the mean if you can't use ```.sum()```? I think it's in ```numeric``` (#include <numeric>).

utils/obsproc/Ghrsst2Ioda.h
```diff
@@ -157,21 +157,33 @@ namespace gdasapp {
...
-        seconds_s = gdasapp::superobutils::subsample2D(seconds, mask, fullConfig_);
       } else {
         sst_s = sst;
         lon2d_s = lon2d;
         lat2d_s = lat2d;
         obserror_s = obserror;
-        seconds_s = seconds;
       }
 
+      // Non-Superobing part only applied to datetime
```
guillaumevernieres, line 172: It's not a problem of efficiency, it 's a bug that we haven't been able to figure out yet. Re-write your comment and todo to reflect this. We WILL forget what we did a few month down the road when we decide this needs to be fixed properly!

guillaumevernieres:
is this ready to review @apchoiCMD ?

apchoiCMD:
> is this ready to review @apchoiCMD ?

Yes, please review and as we discussed, `std::accumulate` won't make the code simpler so that what I did first time is remained.

utils/obsproc/Ghrsst2Ioda.h
```diff
@@ -2,6 +2,7 @@
 
 #include <iostream>
 #include <netcdf>    // NOLINT (using C API)
+#include <numeric>
```
guillaumevernieres: Is this needed?

apchoiCMD: > Is this needed?

Nope, I am going to remove it.

guillaumevernieres (commented):
Add a `TODO` where appropriate to remind us that for some reason that still elude us, what's read by our code doesn't exactly match what's in the file for the date ...

guillaumevernieres (approved):
Looks good. Thanks @apchoiCMD

PR #961, opened 2024-03-08
Switch to FileHandler

### LINKED ISSUE #960: Remove copyfile here so that we can get rid of solo dependency
https://github.com/NOAA-EMC/GDASApp/blob/a3c3c10a62ac54dde502598129f8b543197325a6/ush/socaincr2mom6.py#L29

guillaumevernieres:
- fixes #960 
I think?

### DIFF
--- a/ush/socaincr2mom6.py
+++ b/ush/socaincr2mom6.py
@@ -6,7 +6,7 @@ import numpy as np
 from netCDF4 import Dataset
 from scipy.interpolate import griddata
 import ufsda
-from wxflow import (FileHandler, YAMLFile)
+from wxflow import FileHandler, YAMLFile
 
 
 def socaincr2mom6(incr, bkg, grid, incr_out, nsst_yaml=None):
--- a/ush/socaincr2mom6.py
+++ b/ush/socaincr2mom6.py
@@ -6,8 +6,7 @@ import numpy as np
 from netCDF4 import Dataset
 from scipy.interpolate import griddata
 import ufsda
-from wxflow import YAMLFile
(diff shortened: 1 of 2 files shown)


ush/socaincr2mom6.py
```diff
@@ -6,7 +6,7 @@
 from netCDF4 import Dataset
 from scipy.interpolate import griddata
 import ufsda
-from wxflow import YAMLFile
+from wxflow import (FileHandler, YAMLFile)
```
CoryMartin-NOAA: Why in parentheses?

CoryMartin-NOAA (dismissed):
Looks good, thanks for doing this so quickly! Agile development at its finest!

ush/socaincr2mom6.py
guillaumevernieres: I saw @AndrewEichmann-NOAA doing this and I thought it looked cool ... And since I want to be cool too ...

guillaumevernieres:
> Looks good, thanks for doing this so quickly! Agile development at its finest!

damn ... You probably just jinxed the ci test @CoryMartin-NOAA !!!! Your fault if the tests fail.

guillaumevernieres:
Was there a reason for the `()` for the import @AndrewEichmann-NOAA ? I vaguely remember this being mentioned a while back.

CoryMartin-NOAA:
@guillaumevernieres a google suggests it's for line continuation, but it's a short enough line

CoryMartin-NOAA (approved):

PR #970, opened 2024-03-13
Update aerosol DA yamls with jinja2 templating

aerorahul:
This PR:
- replaces `$( )` with `{{ }}` in the aerosol DA yamls
- adds extension `.j2` to the yaml files that contain jinja2 templating
- is required for global-workflow PR https://github.com/NOAA-EMC/global-workflow/pull/2387


CoryMartin-NOAA:
Thanks @aerorahul looks good except for the unit test, can you (or I can push to this branch) modify this file: https://github.com/NOAA-EMC/GDASApp/blob/develop/test/aero/genyaml_3dvar.sh I think it's just 1) the name of the input template and the change to which wxflow class is imported/used.

CoryMartin-NOAA (approved):

PR #975, opened 2024-03-15
Add EVA modulefile to rocky8 branch

CoryMartin-NOAA:
What the title says :-)

This gets rid of conda envs on Hera and allows for EVA to work on Rocky8.

### DIFF
--- a/modulefiles/EVA/hera.lua
+++ b/modulefiles/EVA/hera.lua
@@ -8,11 +8,24 @@ local pkgNameVer = myModuleFullName()
 
 conflict(pkgName)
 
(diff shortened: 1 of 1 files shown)


CoryMartin-NOAA:
Load this module, then run:
`eva_tests application` and things will start to run. There is an issue, but I'm not sure if it is eva related or this env. So we can debug later if users complain about 'plot logo'

CoryMartin-NOAA:
I found the issue, it is eva related, so this environment is tested and makes figures.
Screenshot 2024-03-15 at 4 34 04 PM

RussTreadon-NOAA (approved):
Looks reasonable.  Test plot generated.  Getting one or more developers to test on Hera would be nice but is not required for merger into `feature/rocky8`.

Approve.

PR #983, opened 2024-03-20
JEDI increment write to cubed sphere history

DavidNew-NOAA:
This PR, a companion to Global Workflow PR [#2420](https://github.com/NOAA-EMC/global-workflow/pull/2420) changes the variational YAML for JEDI to write to cubed sphere history rather than the Gaussian grid. With the new changes to Global Workflow, the new gdas_fv3jedi_jediinc2fv3.x OOPS app will read the JEDI increment from the cubed sphere history, compute the FV3 increment, and interpolate/write it the the Gaussian grid. The only meaningful difference is that the internal calculations, namely computation of the hydrostatic layer thickness increment, will be computed on the native grid rather than on the Gaussian grid, before interpolation rather than after. This makes more sense physically. Eventually the FV3 increment will be written and read to/from cubed sphere history anyway.


CoryMartin-NOAA (dismissed):

DavidNew-NOAA:
Need to do a touch more testing before re-opening

DavidNew-NOAA:
OK, re-opening. I thought there was an error, but things looks good.

CoryMartin-NOAA (approved):

CoryMartin-NOAA:
@DavidNew-NOAA before merging, do you have any plots/statistics summarizing your testing that show this produces comparable results?

DavidNew-NOAA:
@CoryMartin-NOAA 
> @DavidNew-NOAA before merging, do you have any plots/statistics summarizing your testing that show this produces comparable results?

I compared the min, max, and std of the delp and delz increment with the new OOPS app and with the old Python script. The relative error of the delp and delz increments was 10^-7 and 10^-2 respectively for each these three statistics.

PR #998, opened 2024-03-27
Prepare observations for snow DA updates to the ensemble members

jiaruidong2017:
This PR updates several configures to prepare observations for snow DA updates to the ensemble members. Because the snow observations are fetched from the COM_OBS in the gdas deterministic analysis and are saved into the COM_OBS in the enkfgdas analysis for the enkfgdas run, we have to define two different COM_OBS directories (COMIN_OBS vs COMOUT_OBS) and two obs basename heads (OPREFIX_IN vs OPREFIX_OUT) to identify the enkf analysis from the gdas deterministic analysis. 

This PR will contribute to https://github.com/NOAA-EMC/global-workflow/pull/2416.


CoryMartin-NOAA (commented):
Do you have a draft, corresponding workflow PR? I am not particularly thrilled about these additional variable names and would prefer to come up with a way that is more consistent with other observation types. 

EDIT: Never mind, I see you have it linked above. I will comment there.

CoryMartin-NOAA (approved):

CoryMartin-NOAA:
@jiaruidong2017 I'm going to revert this PR as I think it is holding up other workflow PRs. We can re-merge it when the workflow is ready...

PR #1004, opened 2024-03-28
Bugfix on Broken GHRSST Ioda Converter

### LINKED ISSUE #995: GHRSST Ioda Converter is broken in CI
- Few of GHRSST observations are not taken care of when in ci with `Numerical Failed', those obs. should be taken care of

```
1/1 Test #5: test_gdasapp_util_ghrsst2ioda ....***Exception: Numerical 25.81 sec
```

apchoiCMD:
#### This PR should fix the ghrsst ioda converter in ci including;
- If there are no datetime available then should be 0
- Add an additional filter to avoid to append unnecessary obs. and datetime
- New ctest sub-sample was added with the case above.
- fixes #995


utils/obsproc/Ghrsst2Ioda.h
```diff
@@ -184,7 +184,8 @@ namespace gdasapp {
         }
       }
       // Calculate the average and store datetime
-      int64_t seconds_s = static_cast<int64_t>(sum / count);
+      // Replace the seconds_s to o when count is zero
```
guillaumevernieres: typo, not `o` but `0`

utils/obsproc/Ghrsst2Ioda.h
```diff
@@ -184,7 +184,8 @@ namespace gdasapp {
...
+      int64_t seconds_s = count != 0 ? static_cast<int64_t>(sum / count) : 0;
```
guillaumevernieres, line 188: sorry but I forgot, was this the cause of the failure when processing the obs in March?

apchoiCMD: > sorry but I forgot, was this the cause of the failure when processing the obs in March?

Yes, it is- 5 out of 164 obs has no datetime, so count was 0-

utils/obsproc/Ghrsst2Ioda.h
```diff
@@ -220,7 +221,7 @@ namespace gdasapp {
 
       // Basic QC
       Eigen::Array<bool, Eigen::Dynamic, 1> boundsCheck =
-        (iodaVars.obsVal_ > sstMin && iodaVars.obsVal_ < sstMax);
+        (iodaVars.obsVal_ > sstMin && iodaVars.obsVal_ < sstMax && iodaVars.datetime_ > 0.0);
```
guillaumevernieres, line 224: what if the reference time is the middle of the window? I might be missing something but I don't think we should remove obs for which `iodaVars.datetime_ <= 0.0`

AndrewEichmann-NOAA (approved):
Looks good, aside from what Guillaume mentioned

utils/obsproc/Ghrsst2Ioda.h
apchoiCMD: > what if the reference time is the middle of the window? I might be missing something but I don't think we should remove obs for which `iodaVars.datetime_ <= 0.0`

I believe `iodaVars.datetime_` should be positive since it has been set from epochtime, and it was also applied to trim function at AMSR2 ioda converter.

https://github.com/NOAA-EMC/GDASApp/blob/ed441fd15cb25036bd6fc1f0ffe10acc03887bc9/utils/obsproc/IcecAmsr2Ioda.h#L117

PR #1012, opened 2024-03-29
Fix inconsistent VIIRS preprocessing test

### LINKED ISSUE #993: test_gdasapp_util_viirsaod2ioda fails on Hera Rocky 8
When GDASApp `develop` at 2198b41 is built and run on Hera Rocky 8 nodes, `test_gdasapp_util_viirsaod2ioda` fails with 
```
1395: terminate called after throwing an instance of 'oops::TestReferenceFloatMismatchError'
1395:   what():  Test reference Float mismatch @ Line:4
1395: Test Val : 2.2290000000000001e-03
1395: Ref  Val : 2.1952299999999998e-03
1395: Delta    : 3.3770000000000241e-05
1395: Relative tolerance: 2.2121149999999997e-09
1395: Absolute tolerance: 0.0000000000000000e+00
1395: Test Line: '    Min: 0.002229'
1395: Ref Line : '    Min: 0.00219523'
1/1 Test #1395: test_gdasapp_util_viirsaod2ioda ...Subprocess aborted***Exception:   2.29 sec
```
The Hera Rocky 8 execution of this test does not generate a result consistent with the test reference.

This issue is opened to document this failure and its resolution.

CoryMartin-NOAA:
Closes #993 

The `preqc` key was missing from the test YAML and eckit just decided to place some integer in there. This PR should ensure tests pass on all platforms instead of just by chance.

### DIFF
--- a/utils/test/testinput/gdas_viirsaod2ioda.yaml
+++ b/utils/test/testinput/gdas_viirsaod2ioda.yaml
@@ -9,6 +9,7 @@ input files:
 thinning:
   threshold: 0
 channel: 4
+preqc: 2
 
 test: 
   reference filename: testref/viirsaod2ioda.test


CoryMartin-NOAA:
@ypwang19 @RussTreadon-NOAA adding this one line causes the test to pass on Hera. Adding this line but making `preqc: 1` causes Orion to fail the same way Hera did. So this change should be the correct fix.

RussTreadon-NOAA (commented):
Simple enough.  Wonder why `preqc` needs to be explicitly set to 2 now and not before.  The test passes on Orion without the `preqc` line.  The same was true on non-Rocky 8 Hera nodes.   Adding `preqc 2` doesn't adversely impact this test on Orion, does it?

CoryMartin-NOAA:
@RussTreadon-NOAA nope, still passes on Orion with `preqc: 2`. I think it was just some sort of coincidence that it passed before. It's not clear to me why it didn't fail when looking for that value and not finding it, but instead just putting in an integer.

RussTreadon-NOAA (approved):
Thanks, @CoryMartin-NOAA , for confirming that the test still passes on Orion.  Nice detective work!

Approve.

PR #1013, opened 2024-03-29
Use _anl rather than _ges dimensions for increments in FV3 increment converter YAML

DavidNew-NOAA:
npx_ges should be npx_anl for the JEDI and FV3 increments. Same for npy and npz.  See https://github.com/NOAA-EMC/global-workflow/pull/2420#issuecomment-2027575028

### DIFF
--- a/parm/atm/utils/fv3jedi_fv3inc.yaml.j2
+++ b/parm/atm/utils/fv3jedi_fv3inc.yaml.j2
@@ -35,9 +35,9 @@ jedi increment:
     layout:
     - {{ layout_x }}
     - {{ layout_y }}
-    npx: {{ npx_ges }}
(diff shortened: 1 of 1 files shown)


CoryMartin-NOAA:
@DavidNew-NOAA where will you account for the interpolation of the increment when it is added to the background?

In current GDAS operations, the analysis resolution (i.e. that which the increment is computed on) is the ensemble resolution which is half the background resolution. That is why the `.ensres` files exist. The benefit of JEDI writing the increment on the input resolution is that you can just do `analysis = guess + increment`, but if you write out the native increment, you have to interpolate it to the guess resolution first. 

Does that make sense?

CoryMartin-NOAA (approved):
Ignore my late Friday ramblings

PR #1016, opened 2024-04-02
make driver consistent with workflow driver

CoryMartin-NOAA:
The non workflow CI driver would not update submodules. This fixes this (hopefully)


guillaumevernieres (approved):
That should do it! :crossed_fingers:

Issue #972, opened 2024-03-14
How do we want to handle the IODA converter failing on individual sensors?


AndrewEichmann-NOAA:
On a cycling tests then GRHSST IODA converter in `ocanobsprep` failed on sensors sst_avhrr_ma_l3u, sst_viirs_npp_l3u, and sst_viirs_n20_l3u. Removing these sensors from the obs list allowed obs prep to complete with the remaining sensors. Do we want the whole task to fail if the IODA converter fails on an individual sensor, or to perform some error handling and carry on?

RussTreadon-NOAA:
Thank you, @AndrewEichmann-NOAA , for opening this issue.  The same question applies to JEDI ATM cycling.

PR #1028, opened 2024-04-06
Catch error when trying to copy missing obs files from DATA to ROTDIR in prepoceanobs

AndrewEichmann-NOAA:
Catches error when `FileHandler` tries to copy a nonexistant obs file (usually due to a failing IODA converter) and prints a warning to the log.

Tested by including sst_avhrr_mb_l3u and then deleting file after conversion.

Partially addresses https://github.com/NOAA-EMC/GDASApp/issues/972

### DIFF
--- a/scripts/exglobal_prep_ocean_obs.py
+++ b/scripts/exglobal_prep_ocean_obs.py
@@ -170,7 +170,16 @@ for obsspace_to_convert in obsspaces_to_convert:
     processes.append(process)
 
 # Wait for all processes to finish
+# TODO(AFE): add return value checking
 for process in processes:
     process.join()
 
-FileHandler({'copy': files_to_save}).sync()
+#TODO(AFE): Find a better way to do the "no file found" exception handling -
+# this way make individual calls to FileHandler for each file, instead of 
+# batching them.
+for file_to_save in files_to_save:
+    try:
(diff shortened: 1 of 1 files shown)


scripts/exglobal_prep_ocean_obs.py
```diff
@@ -170,7 +170,16 @@ def run_netcdf_to_ioda(obsspace_to_convert):
     processes.append(process)
 
 # Wait for all processes to finish
+# TODO(AFE): add return value checking
 for process in processes:
     process.join()
 
-FileHandler({'copy': files_to_save}).sync()
+# TODO(AFE): Find a better way to do the "no file found" exception handling -
```
guillaumevernieres, line 177: Create an issue pointing to this line so we don't forget to disable this once the functionality is in FileHandler.

guillaumevernieres (dismissed):
Looks fine to me as a temporary fix. Thanks @AndrewEichmann-NOAA

apchoiCMD (dismissed):

scripts/exglobal_prep_ocean_obs.py
AndrewEichmann-NOAA: done

guillaumevernieres (approved):

Issue #1032, opened 2024-04-09
Run g-w link script before running workflow tests


AndrewEichmann-NOAA:
The anticipated merging of https://github.com/NOAA-EMC/global-workflow/pull/2459 should break GDASApp jjob ctests if the g-w linking script is not run, this should probably be added to the `build.sh`

PR #1034, opened 2024-04-10
Run g-w linker script before ctest for prepoceanobs task

AndrewEichmann-NOAA:
In anticipation of the merging of https://github.com/NOAA-EMC/global-workflow/pull/2459 , the `test_gdasapp_soca_setup_obsprep` ctest which sets up for `test_gdasapp_soca_JGLOBAL_PREP_OCEAN_OBS` now runs the global-workflow `sorc/link_workflow.sh` script to allow the observation config files to be accessed in a global-workflow-compliant way. 

Effectively addresses https://github.com/NOAA-EMC/GDASApp/issues/1032


test/soca/gw/setup_obsprep.sh
```diff
@@ -13,6 +13,11 @@ project_source_dir="$1"
 testdatadir="${project_source_dir}/test/soca/testdata"
 testdatadir_bufr="${project_source_dir}/build/gdas/test/testdata"
 
+# run the g-w link script so that the config files can be found
+cd ${project_source_dir}/../../..
```
guillaumevernieres: why not just 
```bash
(cd ${project_source_dir}/../../.. && ./link_workflow.sh)
```
?

guillaumevernieres (dismissed):
Seems fine to me.

test/soca/gw/setup_obsprep.sh
CoryMartin-NOAA: Also, can you not do:
`${project_source_dir}/../../../link_workflow.sh` ?

the script should get its path regardless of where it is executed from

apchoiCMD (dismissed):

test/soca/gw/setup_obsprep.sh
AndrewEichmann-NOAA: @guillaumevernieres Your bash-fu is superior to mine, Sir
@CoryMartin-NOAA I tried that and for some reason it didn't like it

apchoiCMD (approved):

guillaumevernieres (approved):

PR #1039, opened 2024-04-12
Fix test output for fv3jedi_fv3inc.h

DavidNew-NOAA:
I foolishly moved the test output print statements in fv3jedi_fv3inc.h around for aesthetic reasons AFTER I had ran test_gdasapp_fv3jedi_fv3inc to confirm that it still passed. It broke the test but I moved the print statements back to where they belong.

The test now passes.


CoryMartin-NOAA (approved):

DavidNew-NOAA:
@CoryMartin-NOAA I was not expecting this test to pass on Orion, but fail on Hera. Something about "double free or corruption (!prev)" I will have to take a look on Monday.

PR #1040, opened 2024-04-12
Save basic stats in csv at each cycle

guillaumevernieres:
### Work done
Computes basic stats (count, bias and rmse of omb's and oma's) for different ocean basins, this assumes the obs were flagged with the ocean mask from `RECCAP2_region_masks_all_v20221025.nc`.
- The stats are computed within the post ocean analysis jjob. Failure of this job will only results in a warning.
- I provided a simple plotting script that will accept multiple experiments as arguments and plot time series. Not sure if we want this in the vrfy job

### Work left to do
We're not flagging the obs currently, so that code will do nothing until we update the prep ocean obs yaml to point to  `RECCAP2_region_masks_all_v20221025.nc` (it's been copied to the soca fix dir on hera and orion). To have the obs flagged in develop will require the ioda converters to not fail if the `RECCAP2_region_masks_all_v20221025` isn't found.


CoryMartin-NOAA (approved):
Looks good, I think things will need cleaned up for "ops" but fine for current purpose.

PR #1050, opened 2024-04-16
New ocean/ice B diag

### LINKED ISSUE #927: Localization for EnVar is too slow
I might have been a bit too quick to swap `BUMP` with `diffusion`. Test/profile/come up with a faster solution, the EnVar is currently turned off because of this issue.

### LINKED ISSUE #939: Test the marine DA with DOHYBVAR=YES AND NO
It was set to yes in the ci, so we failed catching the bug related to the wrong naming of the static members. We need to implement the 2 options as part of the ci.

### LINKED ISSUE #982: Parametric diag of B
Get away from the dependence to an offline ensemble for the static B.

guillaumevernieres:
### What was done
- Variance partitioning of ice and ocean backgrounds
- Switch to using the `StdDev` saber block with the above fields as a proxy of the background error
- Changed the obs filtering
- Removed the hybrid B, will plug it back in later
- Use Travis' script to extract the vertical decorrelation scales from the bkg, using the mld.
- Use of jinja for the new yamls
- setting NMEM_ENS to something larger that 3 will trigger the hybrid envar and use static or real members, depending on the status or `DO_HYBVAR`  

The results are reasonable but some of the parameters of the variance partitioning still need to be optimized further

- fixes #982 
- fixes #927 
- fixes #939 (not exactly but good enough. It tests the hybrid envar with static members)


scripts/exgdas_global_marine_analysis_bmat.sh
```diff
@@ -95,15 +103,22 @@ if [ $err -gt 0  ]; then
...
 fi
 
+################################################################################
+# Initialize the vertical diffusion blocks
+# Compute the MLD dependent vertical scales
+# TODO(G): Probably not the right way to execute this script
+python ${HOMEgfs}/sorc/gdas.cd/sorc/soca/tools/calc_scales.py soca_vtscales.yaml
 clean_yaml soca_parameters_diffusion_vt.yaml
+
+# Vertical fiffusion and normalization
```
apchoiCMD: probably "diffusion"?

apchoiCMD (dismissed):

scripts/exgdas_global_marine_analysis_post.py
```diff
@@ -65,12 +65,12 @@ def list_all_files(dir_in, dir_out, wc='*', fh_list=[]):
 domains = ['ocn', 'ice']
 for domain in domains:
     # Copy of the diagonal of the background error for the cycle
-    post_file_list.append([os.path.join(anl_dir, 'static_ens', f'{domain}.bkgerr_stddev.incr.{bdate}.nc'),
+    post_file_list.append([os.path.join(anl_dir, f'{domain}.bkgerr_stddev.incr.{mdate}.nc'),
```
CoryMartin-NOAA, line 69: what is `mdate`?

CoryMartin-NOAA (dismissed):
One minor comment, that if I understand it properly should probably be instead `${PDY}${cyc}`

scripts/exgdas_global_marine_analysis_post.py
guillaumevernieres: `mdate` is `cdate` in iso format.

guillaumevernieres:
> One minor comment, that if I understand it properly should probably be instead `${PDY}${cyc}`

that wouldn't work, the file written by soca has the iso date.

CoryMartin-NOAA:
@guillaumevernieres the ensemble handler test fails, is that expected?

scripts/exgdas_global_marine_analysis_post.py
CoryMartin-NOAA: got it

guillaumevernieres:
> @guillaumevernieres the ensemble handler test fails, is that expected?

Not expected, I'll look into it.

parm/soca/berror/saber_blocks.yaml
```diff
@@ -1,101 +1,42 @@
...
 
-    linear variable change:
-      input variables: [cicen, hicen, hsnon, socn, tocn, uocn, vocn, ssh]
-      output variables: [cicen, hicen, hsnon, socn, tocn, uocn, vocn, ssh]
-      linear variable changes:
+saber outer blocks:
+- saber block name: StdDev
+  read:
+    model file:
+      date: '{{ATM_WINDOW_MIDDLE}}'
```
AndrewEichmann-NOAA: Should this line and the others in this file have spaces around the variable, inside the curly brackets?

guillaumevernieres: It's not jinja template yet, so no.

apchoiCMD (approved):

parm/soca/berror/soca_hybrid_bmat.yaml
```diff
@@ -0,0 +1,92 @@
+covariance model: hybrid
+components:
+- covariance:
+    covariance model: SABER
+    saber central block:
+      saber block name: EXPLICIT_DIFFUSION
+      active variables: [tocn, socn, ssh, cicen]
+      geometry:
+        mom6_input_nml: mom_input.nml
+        fields metadata: ./fields_metadata.yaml
```
AndrewEichmann-NOAA, line 10: should this be the whole path?

parm/soca/berror/soca_hybrid_bmat.yaml
```diff
@@ -0,0 +1,92 @@
...
+      pattern: '%mem%'
+      nmembers: ${ENS_SIZE}
+    localization:
+      localization method: SABER
+      saber central block:
+        saber block name: EXPLICIT_DIFFUSION
+        active variables: [tocn, socn, ssh]
+        geometry:
+          mom6_input_nml: mom_input.nml
+          fields metadata: ./fields_metadata.yaml
```
AndrewEichmann-NOAA, line 69: same question about path

parm/soca/berror/soca_hybrid_bmat.yaml
```diff
@@ -0,0 +1,92 @@
...
+  weight:
+    value: 1.00
+
+- covariance:
+    covariance model: ensemble
+    members from template:
+      template:
+        read_from_file: 1
+        date: '{{ATM_WINDOW_MIDDLE}}'
+        basename: ./static_ens/
```
AndrewEichmann-NOAA, line 56: same question about path

parm/soca/berror/soca_static_bmat.yaml
```diff
@@ -0,0 +1,42 @@
+covariance model: SABER
+saber central block:
+  saber block name: EXPLICIT_DIFFUSION
+  active variables: [tocn, socn, ssh, cicen]
+  geometry:
+    mom6_input_nml: mom_input.nml
+    fields metadata: ./fields_metadata.yaml
```
AndrewEichmann-NOAA, line 7: path?

parm/soca/berror/soca_hybrid_bmat.yaml
guillaumevernieres: We've defined it this way for all the yamls that have geometry. If we have to have a full path, we'll issue a separate pr for the change.

AndrewEichmann-NOAA (approved):

PR #1058, opened 2024-04-17
Remove R2D2 and SOLO from the GitHub Actions workflow CI

CoryMartin-NOAA:
We removed these from the unit tests in #836 but missed this one. This is failing because of our RZDM issues, but it exposed that these need to be removed.


guillaumevernieres (approved):
Good catch.

PR #1064, opened 2024-04-19
Fix aero_diagb template

CoryMartin-NOAA:
The aero_diagb template was not quite right, this should allow the executable to complete successfully.


guillaumevernieres (approved):
:+1:

PR #1066, opened 2024-04-22
Revert "Prepare observations for snow DA updates to the ensemble members"

CoryMartin-NOAA:
Reverts NOAA-EMC/GDASApp#998


CoryMartin-NOAA:
@aerorahul I will merge this and then we should try the subsequent hash in the workflow to see if this fixes the snow test.

aerorahul (approved):

PR #1067, opened 2024-04-22
Add basin flags to the ocean obs

### LINKED ISSUE #985: Ocean basin is missing in the obsprep yamls
Looks like we forgot to turn that on in the obs. proc. yamls ...

guillaumevernieres:
- fixes #985


AndrewEichmann-NOAA (dismissed):
Looks good to me. Was it crashing before due to the lack of a masking file?

AndrewEichmann-NOAA (approved):

PR #1080, opened 2024-04-29
Bug fix for the datetime in Smap converter

### LINKED ISSUE #1023: SMAP converter produces the wrong datetime
The converted datetime looks off. Discovered while adding smap to the cycling.

apchoiCMD:
#### For Smap_SSS:
- Replace the method to get a correct datetime from epoch time by using boost/date_time/Gregorian
- JulianDate - 2440588 will make half day off- 
- The above method will avoid dealing with decimal precision.
- Resolves #1023


utils/obsproc/Smap2Ioda.h
```diff
@@ -83,21 +85,47 @@ namespace gdasapp {
...
+      while (day > 31) {
+          if (month == 4 || month == 6 || month == 9 || month == 11) {
+              day -= 30;
+          } else if (month == 2) {
+              day -= 28;  // Assuming not a leap year
+          } else {
+              day -= 31;
+          }
+          month++;
+      }
```
guillaumevernieres: That looks wrong (what if it is a leap year) and definitely out of place.

utils/obsproc/Smap2Ioda.h
```diff
@@ -83,21 +85,47 @@ namespace gdasapp {
...
-      int daysSinceEpoch = julianDate - 2440588;
+      // Set epoch time for SMAP_SSS
+      util::DateTime epochDtime("1970-01-01T00:00:00Z");
 
-      // Calculate seconds
-      int secondsSinceEpoch = (daysSinceEpoch + startDay) * 86400;
+      // Convert Obs DateTime objects to epoch time offsets in seconds
+      // 0000-00-00T00:00:00Z will be converterd to negative seconds
+      int64_t timeOffsets
+           = ioda::convertDtimeToTimeOffsets(epochDtime, {dateTime})[0];
```
guillaumevernieres, line 103: we shouldn't have to do this.
Try to make use of boost:
```c++
#include <boost/date_time/gregorian/gregorian.hpp>
```
has a bunch of nifty datetime utilities that probably would have saved us some headache.

utils/obsproc/Smos2Ioda.h
```diff
@@ -80,7 +80,7 @@ namespace gdasapp {
         iodaVars.obsVal_(i) = sss[i];
         iodaVars.obsError_(i) = sss_error[i];
         iodaVars.preQc_(i) = sss_qc[i];
-        iodaVars.datetime_(i) =  static_cast<int64_t>(datetime_[i]*86400.0f) + mjd2000;
+        iodaVars.datetime_(i) =  static_cast<int64_t>((datetime_[i]*86400.0f) + mjd2000);
```
guillaumevernieres: is there a point to the addition of the parenthesis?

guillaumevernieres (changes requested):
We shouldn't have to worry about leap years ... Somebody did before us. I think this can be re-written using ```boost```

AndrewEichmann-NOAA (commented):
Nothing to add - yeah there should be libraries

utils/obsproc/Smos2Ioda.h
apchoiCMD: No, it is correct and I reverse smos.

utils/obsproc/Smap2Ioda.h
```diff
@@ -82,22 +84,24 @@ namespace gdasapp {
...
-        year = 0;
+        // 1400 is given by boost library
+        // final seconds goes negative then trimmed
+        year = 1400;
       }
 
-      // Convert a date to Julian date
-      uint64_t julianDate = util::datefunctions::dateToJulian(year, month, day);
+      // Define a date
+      boost::gregorian::date d(year, 1, 1);
```
guillaumevernieres: could we try to avoid single letter variables ...

utils/obsproc/Smap2Ioda.h
```diff
@@ -82,22 +84,24 @@ namespace gdasapp {
...
+
+      // Add a startDay to the defined date
```
guillaumevernieres: that does not describe what's done below

utils/obsproc/Smap2Ioda.h
```diff
@@ -82,22 +84,24 @@ namespace gdasapp {
...
+      d += boost::gregorian::date_duration(startDay - 1);
```
guillaumevernieres: why do we need to do this?

guillaumevernieres (commented):
Some minor clean up and that should be good to go. Thanks @apchoiCMD .

utils/obsproc/Smap2Ioda.h
```diff
@@ -82,22 +84,24 @@ namespace gdasapp {
...
       if (year == -9999) {
-        year = 0;
+        // 1400 is given by boost library
+        // final seconds goes negative then trimmed
+        year = 1400;
       }
 
-      // Convert a date to Julian date
-      uint64_t julianDate = util::datefunctions::dateToJulian(year, month, day);
+      // Define a date
```
guillaumevernieres: sorry for being picky with comments, but what do you mean by "define a date"? A random date?

AndrewEichmann-NOAA (approved):
This looks much nicer

guillaumevernieres (approved):
Thanks @apchoiCMD . Did you create a github issue for the change in datetime format prior to 2021 (or whatever the transition date is)?

PR #1086, opened 2024-04-30
Time series of csv stats

guillaumevernieres:
Fixed a few issues in the time series plot. Does things like this:
csv-ex


apchoiCMD (approved):
Very interesting and look forward to seeing other basin, too

PR #1089, opened 2024-05-01
Visualize stats in simple html document

guillaumevernieres:
I got tiered of looking for figures and added a basic html file that displays what we need. Ugly and basic, but it does the trick. The page looks like this:

html-stats

_ps: I'd like the same thing for the 100's of figure that used to come out of the verify job._


apchoiCMD (approved):
Very nice! Thanks for doing this!

PR #1090, opened 2024-05-02
Initial commit of marine letkf stubs

AndrewEichmann-NOAA:
Adds stubs and ctest for marine LETKF task. Co-dependent on global-workflow PR https://github.com/NOAA-EMC/global-workflow/pull/2564

Addresses https://github.com/NOAA-EMC/GDASApp/issues/1091

### DIFF
--- a/ush/soca/marine_letkf.py
+++ b/ush/soca/marine_letkf.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python3
+
+from datetime import datetime, timedelta
+import f90nml
+from logging import getLogger
+import os
+from soca import bkg_utils
+from typing import Dict
+import ufsda
+from ufsda.stage import soca_fix
+from wxflow import (AttrDict,
+                    chdir,
+                    Executable,
+                    FileHandler,
+                    logit,
+                    parse_j2yaml,
+                    Task,
+                    Template,
+                    TemplateConstants,
+                    WorkflowException,
+                    YAMLFile)
+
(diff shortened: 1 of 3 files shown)


apchoiCMD (approved):
Looks good-

CoryMartin-NOAA (approved):

ush/soca/marine_letkf.py
```diff
@@ -0,0 +1,90 @@
...
+                    Task,
+                    Template,
+                    TemplateConstants,
+                    WorkflowException,
+                    YAMLFile)
+
+logger = getLogger(__name__.split('.')[-1])
+
+
+class MarineLETKF(Task):
```
aerorahul, line 26: I know `Analysis` task has a few (2) atmosphere specific methods, but the constructor for it sets up the window, etc. Also, there are generic methods to construct the JEDI yaml, copy observations, etc.
This class deviates from the framework that the other analysis tasks are developed around.
If there is a need to update `Analysis` task to be further accommodating of ocean/seaice requirements, it would be better to discuss and develop rather than have to revisit when there is a crunch.

guillaumevernieres: That's why we added you as a reviewer and that's what a PR is for @aerorahul, discuss. 

Can you switch to using `Analysis` as parent class @AndrewEichmann-NOAA ? Also, while re-reading your code, I saw a few places where you are using `recentering` instead of `letkf`.

PR #1107, opened 2024-05-11
Resume the Marine Vrfy Task on ctests

### LINKED ISSUE #1057: The ocean/ice verify task is broken
We commented out the test, but it'd be nice to have it working again! It's failing when reading the grid I think ...
Note: It's possible that `h` was removed from the geometry. We'll need to get the thicknesses from the history files.

apchoiCMD:
#### This PR will resume the Marine Vrfy Task on ctests
- Checked removed `h` thickness variable and make it work again
- Make modules work again
- Correction for history files
- Fixes #1057


guillaumevernieres (dismissed):
The bug fix looks good, but it seems that my previous PR broke the `post` test ... Oops ...
The fix is in review in the `g-w`.

guillaumevernieres (approved):

PR #1104, opened 2024-05-10
Change YAML and add jjob CTest for new "atmensanlfv3inc" job in Global Workflow

DavidNew-NOAA:
This PR is a companion to Global Workflow PR #[2593](https://github.com/NOAA-EMC/global-workflow/pull/2592), which adds the "atmensanlfv3inc" job to Global Workflow. That job converts the JEDI ensemble increments into FV3 readable increments. This JCB-GDAS companion PR is #[3](https://github.com/NOAA-EMC/jcb-gdas/pull/3), which updates the input YAMLs for the FV3 increment converter

This PR does two things:
1. It adds a jjob CTest to GDASApp to test this new job.
2. It modifies the LETKF input YAML to write out cubed-sphere history increments, rather than Gaussian increments, so the FV3 increment converter can read the resulting increment off the native grid and then itself write out a Gaussian increment.

PR #1108, opened 2024-05-13
Update JCB hashes

CoryMartin-NOAA:
Goes with #1104


DavidNew-NOAA (approved):

RussTreadon-NOAA (approved):
Looks right to me, @DavidNew-NOAA is the expert here.

PR #1112, opened 2024-05-16
Fix ice water FV3 increment variable name and add a few more

DavidNew-NOAA:
This PR corrects the name of ice water mixing ratio in the corresponding fields metadata YAML and adds a few more tracers. Many these additional tracers may never have increments, but I might as well add them just in case. (I got these tracers by putting a print statement in the UFS code when I was debugging my cubed-sphere increment reading branch)

### DIFF
--- a/parm/io/fv3jedi_fieldmetadata_fv3inc.yaml
+++ b/parm/io/fv3jedi_fieldmetadata_fv3inc.yaml
@@ -19,11 +19,30 @@ field metadata:
   io name: o3mr_inc
 
 - long name: cloud_liquid_ice
-  io name: icmr_inc
+  io name: ice_wat_inc
 
 - long name: air_pressure_thickness
   io name: delp_inc
 
 - long name: hydrostatic_layer_thickness
   io name: delz_inc
+
+- long name: rain_water
+  io name: rainwat_inc
+
+- long name: snow_water
+  io name: snowwat_inc
+
+- long name: graupel
+  io name: graupel_inc
+
+- long name: cloud_ice_number_concentration
+  io name: ice_nc_inc
+
+- long name: rain_number_concentration
+  io name: rain_nc_inc
+
+- long name: sgs_tke
+  io name: sgs_tke_inc
+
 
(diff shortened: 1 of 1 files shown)


CoryMartin-NOAA (approved):

RussTreadon-NOAA:
@DavidNew-NOAA and @CoryMartin-NOAA .  This PR breaks g-w jobs gdasanalcalc and gfsanalcalc.  

[gsi_utils.fd](https://github.com/NOAA-EMC/GSI-utils) [`src/netcdf_io/interp_inc.fd/driver.F90`](https://github.com/NOAA-EMC/GSI-utils/blob/develop/src/netcdf_io/interp_inc.fd/driver.F90) hardwires the name of the cloud liquid ice water increment as `icmr_inc` on [line 82](https://github.com/NOAA-EMC/GSI-utils/blob/bb03e172e0d0d9c56d6da7788ca033bfb5ef5119/src/netcdf_io/interp_inc.fd/driver.F90#L82) .

gfsanalcalc and gdasanalcalc abort with
```
 - PROCESS RECORD: icmr_inc

 FATAL ERROR: inquiring icmr_inc id for file=siginc.nc.06:
 NetCDF: Variable not found
 STOP.
999
```

While we could replace `icmr_inc` in `driver.F90` with `ice_wat_inc`, doing so will break GSI cycling.  The GSI writes the analysis increment file with `icmr_inc`.   We need to add logic to `driver.F90` to support both names.

`grep -r icmr_inc` in `gsi_utils.fd/src/` caught `src/EnKF/gfs/src/recentersigp.fd/recentersigp.f90` as another code which this PR will likely break.  We execute `recentersigp.x` in the enkfgdasecen* jobs.  It's easier to fix `recentersigp.f90`.  `icmr_inc` appears in a `case` statement.  We can add a new `case` statement for `ice_wat_inc`.

DavidNew-NOAA:
@RussTreadon-NOAA Ah, sorry. I'll take a look at both codes and try to fix it

DavidNew-NOAA:
@RussTreadon-NOAA See https://github.com/NOAA-EMC/GSI-utils/pull/43

DavidNew-NOAA:
I'm perplexed though why GSI is using a different variable name than UFS for the ice increment.

PR #1114, opened 2024-05-17
Replicate the creation of the gw-ci cycling test

guillaumevernieres:
It's useful to me so I suspect it could be useful to others. This PR adds 3 ctests that run the `g-w` `create_experiment.py` script for :
- Aero/Snow JEDI DA
- Atmos JEDI DA
- WCDA (SOCA + GSI)

each ctest will create an exp directory with a `COMROOT` and `EXPDIR`, the user can then cycle the experiment with `rocoto`.

I stopped at the creation of the experiment but this could be pushed further if we wanted too ...


CoryMartin-NOAA (dismissed):
Looks good, thanks!

test/gw-ci/create_exp.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+expyaml="$1"
+export pslot="$2"
+HOMEgfs="$3"
+export RUNTESTS="$4"/${pslot}
+export ICSDIR_ROOT=/scratch1/NCEPDEV/global/glopara/data/ICSDIR
```
CoryMartin-NOAA: This is Hera specific, can we support other platforms too?

test/gw-ci/create_exp.sh
```diff
@@ -0,0 +1,13 @@
...
+export SLURM_ACCOUNT="da-cpu"
```
CoryMartin-NOAA, line 6: What is this used for? This test just runs a script, it doesn't need slurm does it?

test/gw-ci/create_exp.sh
guillaumevernieres: arf ... will add the other machines

test/gw-ci/create_exp.sh
guillaumevernieres: no, but the intent is for the user to run what was created, what comes out of the `g-w` config is not what we want.

CoryMartin-NOAA (commented):
Instead of removing it, can we have the cmake define ICSDIR based on the machine it was built on?

guillaumevernieres:
> Instead of removing it, can we have the cmake define ICSDIR based on the machine it was built on?

this is done under the hood in the g-w

CoryMartin-NOAA (dismissed):

danholdaway (dismissed):

CoryMartin-NOAA (dismissed):

guillaumevernieres:
> > Instead of removing it, can we have the cmake define ICSDIR based on the machine it was built on?
> 
> this is done under the hood in the g-w

I thought I made sure I didn't have icsdir_root defined in the shell ... But apparently not, ICSDIR_ROOT needs to be defined.

AndrewEichmann-NOAA (approved):
looks reasonable

PR #1116, opened 2024-05-17
Add inverse distance weighting option to superob function

ypwang19:
**Description:**

1. The superob function now incorporates the capability for inverse distance weighted averaging. If optional parameters are provided when calling the function, it conducts weighted averaging; otherwise, it defaults to simple averaging.

3. The VIIRS converter has been enhanced to utilize this new functionality. Weighted averaging is only used in the computation of superobbed ObsValue and ObsError.


utils/obsproc/superob.h
```diff
@@ -8,11 +8,47 @@
...
+        double longitude;
+    };
+    const double EARTH_RADIUS_KM = 6371.0;
+
+    // Function to convert degrees to radians
+    double toRadians(double degree) {
+        return degree * (M_PI / 180.0);
+    }
+
+    // Function to calculate the Haversine distance between two points
```
guillaumevernieres: You should leverage functions from `oops` and/or `atlas` for this. 
Look in the `atlas` source code for `Earth::distance`.

utils/obsproc/superob.h
```diff
@@ -8,11 +8,47 @@
 
 namespace gdasapp {
   namespace superobutils {
+    // Function to compute distance between two points
+    struct GeoPoint {
```
guillaumevernieres: make use of `eckit::geometry::Point2` instead of declaring a new data structure.

utils/obsproc/superob.h
```diff
@@ -8,11 +8,47 @@
...
+        double latitude;
+        double longitude;
+    };
+    const double EARTH_RADIUS_KM = 6371.0;
```
guillaumevernieres: use `atlas::Earth` ... or something ... But the earht geometry is already defined in atlas, so no need to redefine the earht radius again.

guillaumevernieres (commented):
Looks good, but it might be a good idea to start poking at what's available in `atlas`. You shouldn't have to redefine things like the earth radius. Distance on the sphere is already an implemented function somewhere in atlas.

utils/obsproc/superob.h
ypwang19: Thank you for the suggestion, theses are very useful information. I will check functions available in altas and eckit.

utils/obsproc/superob.h
ypwang19: Distance function from atlas is used in the updated code.

guillaumevernieres (approved):
Nice!

CoryMartin-NOAA (approved):
Looks great, thanks!

PR #1129, opened 2024-05-24
Add optional testing block to 3dvar and lgetkf input YAMLs for jjobs

DavidNew-NOAA:
This PR turns on the testing blocks, added in jcb-algorithms #[1](https://github.com/NOAA-EMC/jcb-algorithms/pull/1), in the input YAMLs for the 3dvar and lgetkf jjob test, turning them into JEDI application tests that check for changes in the output states. This PR updates the jcb-algorithms hash to the latest develop as well.

The motivation for this PR is that we would like to know not only if changes to GDASApp cause run failures, but also whether they change outputs.

I used the same relative and absolute tolerance used in https://github.com/JCSDA-internal/fv3-jedi/blob/develop/test/testinput/lgetkf.yaml and https://github.com/JCSDA-internal/fv3-jedi/blob/develop/test/testinput/hyb-3dvar_gfs.yaml from FV3-JEDI. I don't know why one uses 1e-7 and the other 1e-6 for the absolute tolerances. Reviewers can tell me what they think is reasonable.

I ran all the jjob tests and they passed on Orion with the feature/atmensanlfv3inc branch of Global Workflow. That branch needs to be merged into develop with #[2592](https://github.com/NOAA-EMC/global-workflow/pull/2592).

Note: The JCB submodule in Global Workflow will need to updated to the latest develop branch due to change in this commit of jcb-algorithms https://github.com/NOAA-EMC/jcb-algorithms/commit/1a1497ced404545044497a475835970d5f43d2f3


danholdaway:
This might be a case for moving jcb to GDASapp instead of G-W. Perhaps that makes sense since only GDASapp actually uses it. What do you think @CoryMartin-NOAA? It seems like this creates a situation where we have to merge GDASapp with failing CI because of upstream dependency.

danholdaway (approved):
Really great to have these test be more regressive in nature!

CoryMartin-NOAA:
@danholdaway I think that depends on if it is expanded to use jcb with the model and other components. If just the 'DASApps', then I agree it should probably be brought inline with OOPS, IODA, etc.

CoryMartin-NOAA (approved):

PR #1136, opened 2024-05-31
Adjust absolute float tolerance for variational and ensemble DA jjob tests

DavidNew-NOAA:
This PR adjusts the absolute float tolerance for both variational and ensemble DA jjob tests to 10e-5 from 10e-6 and 10e-7 respectively. The old tolerances came from CTests in FV3-JEDI that involve very low resolution experiments. @danholdaway , suggested 10e-5 - 10e-4, since simple machine error could even cause the test to pass on one machine but fail on another.

Reviewers please let me know if the relative float tolerance (10e-3) makes sense as well.


RussTreadon-NOAA (approved):
Approve

CoryMartin-NOAA (approved):
Seems reasonable to me

danholdaway (approved):

PR #1140, opened 2024-06-03
Diffusion parameter yaml template

andytangborn:
This PR adds the yaml template aero_diffparm.yaml.j2 needed for running the diffusion parameter executable.

### DIFF
--- a/parm/aero/berror/aero_diffparm.yaml.j2
+++ b/parm/aero/berror/aero_diffparm.yaml.j2
@@ -1,0 +1,47 @@
+geometry:
+  fms initialization:
+    namelist filename: ./fv3jedi/fmsmpp.nml
+    field table filename: ./fv3jedi/field_table
+  akbk: ./fv3jedi/akbk.nc4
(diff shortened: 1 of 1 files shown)


parm/aero/berror/aero_diffparm.yaml.j2
```diff
@@ -0,0 +1,47 @@
+geometry:
+  fms initialization:
+    namelist filename: ./fv3jedi/fmsmpp.nml
+    field table filename: ./fv3jedi/field_table
+  akbk: ./fv3jedi/akbk.nc4
+  npx: {{ npx_ges }}
```
CoryMartin-NOAA, line 6: so I just had a thought here, somewhat unrelated to this PR but we need to test it, can we run the analysis at lower res than the background using these tools? In ops, we will be C768/C1152 but the analysis will need to be C384 or lower. Let's discuss in an hour.

CoryMartin-NOAA (approved):

PR #1141, opened 2024-06-03
Add yaml file for prepobsaero task

ypwang19:
This PR adds the yaml template prep_viirs.yaml.j2 needed for running the VIIRS aerosol observation preparation task.

### DIFF
--- a/parm/aero/prep/prep_viirs.yaml.j2
+++ b/parm/aero/prep/prep_viirs.yaml.j2
@@ -1,6 +1,6 @@
 provider: VIIRSAOD
-window begin: {{ window_begin }}
-window end: {{ window_end }}
+window begin: {{ window_begin | to_isotime }}
+window end: {{ window_end | to_isotime }}
 binning:
   cressman radius: 100
   stride: 132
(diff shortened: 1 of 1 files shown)


parm/aero/prep/prep_viirs.yaml.j2
```diff
@@ -0,0 +1,15 @@
+provider: VIIRSAOD
+window begin: {{ window_begin }}
```
CoryMartin-NOAA: do we need ` | to_isotime ` here on these two lines?
`{{ window_begin | to_isotime }}`

ypwang19: What's the purpose of` | to_isotime`? Now it works fine without it, but I can add it if needed.

CoryMartin-NOAA: I think it should ensure that the date is formatted `YYYY-MM-DDTHH:MM:SSZ`

CoryMartin-NOAA (approved):
looks great thanks!

Issue #1148, opened 2024-06-06
increase tolerance for lgetkf.x reference check


RussTreadon-NOAA:
`test_gdasapp_atm_jjob_ens_run` using GDASApp `develop` at 825f19c (update JEDI hashes) fails on Hercules.  This test passes on Hera and Orion.

The Hercules failure is due to the reference test after lgetkf runs.
```
0: OOPS_STATS Run end                                  - Runtime:    456.38 sec,  Memory: total:    22.59 Gb, per task: min =     3.36 Gb, max =     4.06 Gb
0: Run: Finishing oops::LocalEnsembleDA<FV3JEDI, UFO and IODA observations> with status = 0
0: terminate called after throwing an instance of 'oops::TestReferenceFloatMismatchError'
0:   what():  Test reference Float mismatch @ Line:149
0: Test Val : 3.9113397703941590e-04
0: Ref  Val : 3.9113337546012496e-04
0: Delta    : 6.0157929093924978e-10
0: Relative tolerance: 3.9113367624977039e-10
0: Absolute tolerance: 0.0000000000000000e+00
0: Test Line: 'cloud_liquid_ice                             | Min:+0.0000000000000000e+00 Max:+3.9113397703941590e-04 RMS:+1.0484802023479406e-05'
0: Ref Line : 'cloud_liquid_ice                             | Min:+0.0000000000000000e+00 Max:+3.9113337546012496e-04 RMS:+1.0484801913924773e-05'
srun: error: hercules-07-15: task 0: Aborted (core dumped)
```
The input yaml ends with
```
test:
  reference filename: /work2/noaa/da/rtreadon/git/global-workflow/pr2641_hercules/sorc/gdas.cd/test/atm/global-workflow/lgetkf.ref
  test output filename: ./lgetkf.out
  float relative tolerance: 1e-06
  float absolute tolerance: 0.0
  integer tolerance: 0
```
Increasing `float relative tolerance` to `1e-05` allows the reference check to pass.  

`1e-06` works on Orion and Hera.  Test `test_gdasapp_atm_jjob_ens_run` does not yet run on WCOSS2.  It is possible that a larger `float relative tolerance` is needed on WCOSS2.

RussTreadon-NOAA:
Repeat this test on Cactus.  `test_gdasapp_atm_jjob_ens_run` passes the reference check on Cactus with `float relative tolerance=1e-06`
```
[...]

PR #1154, opened 2024-06-07
Fix keyword for float tolerances in jjob tests

DavidNew-NOAA:
@RussTreadon-NOAA caught a bug in the JCB prototype files for the the jjob tests. They are missing "test_" in the keywords for the float tolerances, so that the jjob tests are just using the defaults. 

https://github.com/NOAA-EMC/GDASApp/issues/1148#issuecomment-2153252171


RussTreadon-NOAA (approved):
Thanks @DavidNew-NOAA for the quick fix.  Looks good to me.

Approve.

danholdaway (approved):

PR #1156, opened 2024-06-07
Use standard config files for jjob test and give output files better names

DavidNew-NOAA:
Now that G-W #[2661](https://github.com/NOAA-EMC/global-workflow/pull/2661) is merged, the custom config.atmanl and config.atmensanl used by the jjob tests can be deleted. This way, the custom files don't have to be updated everytime the standard ones are.

I also changed the sbatch submission, so the log files for each jjob test includes the name of the associated G-W job. This makes it a little easier to browse the logs, without having to guess which one is which.


CoryMartin-NOAA (dismissed):
I did not test but changes look good to me, and many thanks for that slurm log file name changeset!

test/atm/global-workflow/jjob_ens_final.sh
```diff
@@ -47,9 +47,9 @@ fi
 
 # Execute j-job
 if [[ $machine = 'HERA' ]]; then
-    sbatch --ntasks=1 --account=$ACCOUNT --qos=batch --time=00:10:00 --export=ALL --wait ${HOMEgfs}/jobs/JGLOBAL_ATMENS_ANALYSIS_FINALIZE
+    sbatch --ntasks=1 --account=$ACCOUNT --qos=batch --time=00:10:00 --export=ALL --wait --output=atmensanlfinal-%j.out ${HOMEgfs}/jobs/JGLOBAL_ATMENS_ANALYSIS_FINALIZE
```
RussTreadon-NOAA: The sbatch lines for hera, orion and hercules lines are identical.  I'm to blame for this.  We can add orion and hercules to the above machine test and then delete the elif for orion and hercules.  

I suspect this comment applies to all jjob*sh.

RussTreadon-NOAA (commented):
If you like, @DavidNew-NOAA , I can clean up the if-elif blocks in the jjob*sh.  Just let me know.

danholdaway (dismissed):
Nice simplification/cleanup.

test/atm/global-workflow/jjob_ens_final.sh
DavidNew-NOAA: @RussTreadon-NOAA Thanks for pointing that out. I saw it but wasn't sure if it was a placeholder for something in the future. I just committed some changes to remove the elif

RussTreadon-NOAA (approved):
Looks good.  Thank you @DavidNew-NOAA .

Approve.

PR #1158, opened 2024-06-10
Updates to aerosol yaml templates

andytangborn:
Changed name to aero_diffusionparm.yaml.j2 to match global workflow and changes to aero_diagb.yaml.j2 cycling.


ypwang19 (commented):
Renaming the diffusion yaml looks good to me

CoryMartin-NOAA (approved):

PR #1160, opened 2024-06-11
Remove N21 from the aero obs list 

CoryMartin-NOAA:
Remove N21 from the aero obs list until JCB can determine if the cycle we are running should have N21 data available.


ypwang19 (approved):
Thanks for fixing this issue. It looks good.

PR #1167, opened 2024-06-12
Addition of "Convert to Structured Grid" to gdas

jrocha22:
Added ConverttoStructuredGrid to gdas.cc


mains/gdas.cc
```diff
@@ -52,6 +53,9 @@ int runApp(int argc, char** argv, const std::string traits, const std::string ap
   // Define a map from app names to lambda functions that create unique_ptr to Applications
   std::map<std::string, std::function<std::unique_ptr<oops::Application>()>> apps;
 
+  apps["convertstructuredgrid"] = []() {
```
CoryMartin-NOAA: may want to make this and the other entry `converttostructuredgrid` just to be consistent with the App name

mains/gdas.cc
```diff
@@ -96,6 +100,7 @@ int main(int argc,  char ** argv) {
   // Check that the application is recognized
   // ----------------------------------------
   const std::set<std::string> validApps = {
+    "convertstructuredgrid",
```
CatherineThomas-NOAA: @jrocha22: I agree with @CoryMartin-NOAA's comment that this should probably be `converttostructuredgrid` for consistency.

mains/gdas.cc
jrocha22: okay will do!

mains/gdas.cc
jrocha22: sounds good!

CatherineThomas-NOAA: Looks great, thanks!

CatherineThomas-NOAA (approved):
Thanks for making the changes. Approve.

CoryMartin-NOAA:
Kicking off the automated tests. If things look good enough, I'll merge!

danholdaway (approved):

PR #1184, opened 2024-06-20
Add berror yaml for diffusion correlation operator.

### LINKED ISSUE #1177: Add option for diffusion in aero B error templates
We now have a template file to compute the diffusion parameters as well as one for the variance partitioning, for aerosol DA, but we do not have a template file for the use of the diffusion as a saber block in the variational DA. We need to add one of those before we can complete the workflow changes needed to test cycling with diffusion + variance partitioning.

andytangborn:
This PR adds the yaml template staticb_diffusion.yaml.j2  , to replace the bump yaml in the global workflow.


parm/aero/berror/staticb_diffusion.yaml.j2
```diff
@@ -0,0 +1,33 @@
...
+                       mass_fraction_of_hydrophilic_black_carbon_in_air,
+                       mass_fraction_of_hydrophobic_organic_carbon_in_air,
+                       mass_fraction_of_hydrophilic_organic_carbon_in_air,
+                       mass_fraction_of_dust001_in_air, mass_fraction_of_dust002_in_air,
+                       mass_fraction_of_dust003_in_air, mass_fraction_of_dust004_in_air,
+                       mass_fraction_of_dust005_in_air, mass_fraction_of_sea_salt001_in_air,
+                       mass_fraction_of_sea_salt002_in_air, mass_fraction_of_sea_salt003_in_air,
+                       mass_fraction_of_sea_salt004_in_air]
+      horizontal:
+        filepath: Data/diffusion/diffusion_hz
```
CoryMartin-NOAA: is this the right path? same for below, make them consistent with the others

CoryMartin-NOAA (approved):

PR #1191, opened 2024-06-21
Add layout to aero_diffusionparm.yaml.j2

andytangborn:
aero_diffusionparm.yaml.j2 needs a layout so that it can match the number of processors when it is run.


andytangborn:
I just noticed that the staticb_diffusion.yaml.j2 is shown as a new file here. It should already be in develop I think.

andytangborn:
> I just noticed that the staticb_diffusion.yaml.j2 is shown as a new file here. It should already be in develop I think.

Fixed it.

CoryMartin-NOAA (approved):

PR #1200, opened 2024-06-25
Update diffusion parameters yaml

andytangborn:
This PR updates the diffusion parameter path in the yaml so that it matches workflow changes. I separate global workflow PR will be issued shortly.

### DIFF
--- a/parm/aero/berror/aero_diffusionparm.yaml.j2
+++ b/parm/aero/berror/aero_diffusionparm.yaml.j2
@@ -40,11 +40,11 @@ background error:
       - horizontal:
           fixed value: {{ horiz_len }} 
         write:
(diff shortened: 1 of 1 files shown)


parm/aero/berror/aero_diffusionparm.yaml.j2
```diff
@@ -40,11 +40,11 @@ background error:
       - horizontal:
           fixed value: {{ horiz_len }} 
         write:
-          filepath: "{{ DATA }}/berror/diffusion_hz"
```
CoryMartin-NOAA, line 43: Do we also need this change in the Var YAML or does the code there copy it to berror?

CoryMartin-NOAA (approved):

CoryMartin-NOAA:
Unit tests are failing due to NCWCP network issues

PR #1201, opened 2024-06-26
Updated aero_diffusionparm.yaml.j2

andytangborn:
changed the parameter diff_iter to aero_diff_iter in aero_diffusionparm.yaml.j2    
I'll change this in the global workflow as well.


CoryMartin-NOAA:
Sorry I wasn't clear, can all of those variables be modified to be more explicit? `horiz_len` and `fixed_val` should also be more descriptive.

andytangborn:
> Sorry I wasn't clear, can all of those variables be modified to be more explicit? `horiz_len` and `fixed_val` should also be more descriptive.

ahh, I see.  I'll work on all of them.

PR #1207, opened 2024-07-03
Added the climatology read section

andytangborn:
This adds the read section for the climatology stddev, needed for the hybrid. It also includes the hybrid weight coefficient.

### DIFF
--- a/parm/aero/berror/aero_diagb.yaml.j2
+++ b/parm/aero/berror/aero_diagb.yaml.j2
@@ -54,6 +54,7 @@ variables:
 
 rescale: 2.0    # rescales the filtered std. dev. by "rescale"
 number of halo points: 4
+alpha_diagb: 0.5 
 number of neighbors: 16
 simple smoothing:
   horizontal iterations: 0 
--- a/parm/aero/berror/aero_diagb.yaml.j2
+++ b/parm/aero/berror/aero_diagb.yaml.j2
@@ -35,10 +35,10 @@ background error:
   filename_trcr: stddev.fv_tracer.res.nc
   filename_cplr: stddev.coupler.res
 
-Climate background error stddev:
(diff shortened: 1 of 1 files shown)


parm/aero/berror/aero_diagb.yaml.j2
```diff
@@ -35,6 +35,14 @@ background error:
   filename_trcr: stddev.fv_tracer.res.nc
   filename_cplr: stddev.coupler.res
 
+Climate background error stddev:
```
CoryMartin-NOAA: go with lowercase, also: 
```suggestion
climatological background error stddev:
```

parm/aero/berror/aero_diagb.yaml.j2
```diff
@@ -35,6 +35,14 @@ background error:
...
+  filetype: fms restart
+  skip coupler file: true
+  datapath: Data/staticb/
```
CoryMartin-NOAA: Make paths consistent

andytangborn: > Make paths consistent

I need to think about this a little bit. In the past the staticb directory was defined by the workflow . Should we continue that? Or have it be in ./staticb?  I just don't want it to get mixed up with the hybrid stddev which will be in ./stddev

CoryMartin-NOAA: We could call it something like clim_stddev, but regardless it shouldn't be Data/... I don't think

CoryMartin-NOAA:
@andytangborn is this PR still needed or can it be closed?

andytangborn:
> @andytangborn is this PR still needed or can it be closed?

closing it now.

andytangborn:
No longer needed.

PR #1216, opened 2024-07-09
Changes nc4 file suffix to nc on IODA observation files

### LINKED ISSUE #1215: Change IODA observation filename suffixes from nc4 to nc
There is rumored to be a directive to have all IODA observation files have the suffix `nc`, as opposed to `nc4`

AndrewEichmann-NOAA:
Tested with SOCA ctest on Hera

Resolves https://github.com/NOAA-EMC/GDASApp/issues/1215

### DIFF
--- a/parm/soca/obs/config/adt_rads_all.yaml
+++ b/parm/soca/obs/config/adt_rads_all.yaml
@@ -3,11 +3,11 @@ obs space:
   obsdatain:
     engine:
       type: H5File
-      obsfile: !ENV ${DATA}/obs/${OPREFIX}adt_rads_all.${PDY}${cyc}.nc4
+      obsfile: !ENV ${DATA}/obs/${OPREFIX}adt_rads_all.${PDY}${cyc}.nc
   obsdataout:
     engine:
       type: H5File
-      obsfile: !ENV ${DATA}/diags/adt_rads_all.${PDY}${cyc}.nc4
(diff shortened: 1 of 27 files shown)


guillaumevernieres (changes requested):
it's a bit more involved than this @AndrewEichmann-NOAA , prep and post will probably be affected and probably the archiving task in the g-w.

AndrewEichmann-NOAA:
> it's a bit more involved than this @AndrewEichmann-NOAA , prep and post will probably be affected and probably the archiving task in the g-w.

Is there a test (a manual one, a condition to be met) that would satisfy these issues?

guillaumevernieres:
> > it's a bit more involved than this @AndrewEichmann-NOAA , prep and post will probably be affected and probably the archiving task in the g-w.
> 
> Is there a test (a manual one, a condition to be met) that would satisfy these issues?

beyond the ctests that we already have?

guillaumevernieres:
my grepping says you missed "nc4" in `exgdas...post.py`

AndrewEichmann-NOAA:
The ctests work without the change in post, so maybe there should be something else considered, assuming this goes forward

guillaumevernieres:
> The ctests work without the change in post, so maybe there should be something else considered, assuming this goes forward

that's a problem for sure, our ctest only check if stuff ran without failure, it's useful but a bit primitive!

apchoiCMD (approved):

CoryMartin-NOAA:
Do we still need this PR when we are in the process of going to JCB?

PR #1220, opened 2024-07-11
Make aerosol diag B able to handle multiple resolutions + add in more config choices

CoryMartin-NOAA:
This PR adds to the chem_diagb code support for a different climatological std. dev. resolution than that of the analysis resolution. This also adds extra configuration choices in templated YAMLs.


andytangborn (approved):
Looks good.

PR #1226, opened 2024-07-17
Better filtering of D and revert obs QC

### LINKED ISSUE #1217: QA/QC for marine obs
### Insitu filters
- There's currently a bug in the insitu filters
- The obs error are too large
- The obs error are currently set to a constant for Temp and Salt, results in nonphysical analysis. Do something better. 

### SST QA/QC
Needs to be revised.
- filter out obs that are too close to shore
- increase the obs error globally  

To do this, one will have to update the obs `yamls` in 
```
gdas.cd/parm/soca/obs/config
```

### LINKED ISSUE #1219: Variance partitioning based on the correlation length scales
We currently have a discrepancy between the scales in the background error and the correlation operator used in the B-matrix which can result in large increments. Change the variance partitioning tool by filtering or binning using the correlation scales.

### LINKED ISSUE #1232: Miscellaneous bugs
- [x] **Running out of memory**
The 3DVAR went from using ~0.5TB in June to more than 2-2.5TB as of now (07/30/2024)
 
- [x] **Missing python path to pyioda**
Add/export PYTHONPATH in this script:
```
...../jobs/rocoto/prepoceanobs.sh
```
similarly to what is done for the atm obs.

- [x] **Looks like I was wrong about {{ FIXgfs }}**
I know, shocker!!! 
in the gfsv17 yaml, the  `SOCA_INPUT_FIX_DIR` key needs to be replaced:
```yaml
ocnanal:
  SOCA_INPUT_FIX_DIR: {{ HOMEgfs }}/fix/gdas/soca/1440x1080x75/soca
  SOCA_OBS_LIST: {{ HOMEgfs }}/sorc/gdas.cd/parm/soca/obs/obs_list.yaml
  SOCA_NINNER: 100
```

- [x] **soca MOM_input for the 1/4 deg is outdated**
1 fatal error and a few warnings
We need to remove ```DEFAULT_2018_ANSWERS``` (error) and ```ETA_TOLERANCE_AUX``` (warning)

- [x] **No error trapping in the checkpoint task**
One of the job within that task has been failing but the error was not reported by the script [exgdas_global_marine_analysis_chkpt.sh#L70](https://github.com/NOAA-EMC/GDASApp/blob/5583fbb6884a9b98fb3df76058ea2f51414299a2/scripts/exgdas_global_marine_analysis_chkpt.sh#L70)

- [x] **soca to cice variable change is sometimes failing when using 20+ pe's**
profiling for the 0.25deg res:
profiling-soca2cice
we're currently using 40pe's, which fails most of the time and according to the profiling figure above, it is not optimal.
Switch to 16 after stress testing that layout.  

- [x] **ice background fields are all zeros at 0.25deg res.**
Not quite sure what happened, the low-res seems to be ok.

guillaumevernieres:
This PR adds a bit more flexibility to the filtering of D, making use of the generic diffusion and allowing for much larger scales that the simple filtering that we used before.
I also reverted the sst obs error to something a bit more realistic.

- fixes #1219
- fixes #1217 
- fixes #1232 

**This work will now require a companion PR in the `global-workflow`**


AndrewEichmann-NOAA (approved):
nothing obviously wrong

JohnSteffen-NOAA (approved):

Issue #1182, opened 2024-06-20
More converters for sea-ice obs


guillaumevernieres:
- This is mainly for seaice concentration, but we need to look at what else is available (skin temp for ex.)
- Even though there are 8 different instruments/platforms, we should be able to get away with fewer converters, some of the files structure seem exactly the same across platform/inst.
 
Copy/paste of @ilianagenkova 's README:
```
1.ABI-L2-AICEF
2.AMSR2_SEAICE_NH
3.AMSR2_SEAICE_SH  
4.JPSSRR_IceConcentration
5.MIRS_AMSU-MHS_IMG
6.MIRS_ATMS_IMG  
7.MIRS_GMI_IMG
8.MIRS_High_Resolution_IMG 

The sub names (directory) are as follows
1.EMC_ABI-L2-AICEF (PULL/ABI_L2_AICEF)
2.EMC_AMSR2_SEAICE_NH (NDE/AMSR2)
3.EMC_AMSR2_SEAICE_SH (NDE/AMSR2)
4.EMC_JPSSRR_IceConcentration (PULL/JPSSRR_IceConcentration)
5.EMC_MIRS_AMSU-MHS_IMG (PULL/MIRS_AMSU)

8.EMC_MIRS_High_Resolution_IMG (PULL/MIRS_High_Resolution_IMG)
7.EMC_MIRS_GMI_IMG (PULL/MIRS_GMI_IMG)
6a.EMC_MIRS_ATMS_IMG_J1 (PULL/MIRS_ATMS_IMG/J1)
6b.EMC_MIRS_ATMS_IMG_J2 (PULL/MIRS_ATMS_IMG/J2)
6c.EMC_MIRS_ATMS_IMG_SNPP (PULL/MIRS_ATMS_IMG/SNPP)

For (6) MIRS_ATMS_IMG, 
3 subs were created for each satellite because each satellite produces ~2600 files/day.
If files from all 3 satellites were in the same directory, retrieval would have been slow.
```

TODO:
- [x] document the list of retrieved parameters (ice concentration, ice temp, ...)
- [x] make sure all the intrument/platform provided by @ilianagenkova have at least seaice concentration as retrieved obs. Communicate to Iliana all the instruments/platforms that we want as part of the obs processing.
- [x] design the converter(s)
   - [x] MIRS Instruments (NPP)
   - [x] ABI Instruments
   - [x] JPSSRR Instruments
   
- [x] test within a cycling experiment with S-NPP
[...]

PR #1238, opened 2024-07-30
Add MIRS sea-ice product to ioda converter

apchoiCMD:
#### Description
- Task for adding a new ioda converter for new sea-ice concentration
- For sea-ice, AMSR2 and MIRS product will be used

Partially addressed #1182


guillaumevernieres (commented):
Looks good but the test cdl files need to be smaller. 15,000 lines is definitely too large, target something ~1000 lines or less.

utils/obsproc/IcecMirs2Ioda.h
```diff
@@ -0,0 +1,165 @@
...
+        std::stringstream ss;
+        ss << std::setfill('0')
+           << std::setw(4) << year << '-'
+           << std::setw(2) << month << '-'
+           << std::setw(2) << day << 'T'
+           << std::setw(2) << hour << ':'
+           << std::setw(2) << minute << ':'
+           << std::setw(2) << second << 'Z';
+        std::string formattedDateTime = ss.str();
+        util::DateTime dateTime(formattedDateTime);
```
AndrewEichmann-NOAA, line 127: I wonder if there's a utility for this

AndrewEichmann-NOAA:
Looks reasonable

utils/obsproc/IcecMirs2Ioda.h
apchoiCMD: IODA util has it- We have used it to another converters, too

apchoiCMD:
> Looks good but the test cdl files need to be smaller. 15,000 lines is definitely too large, target something ~1000 lines or less.

Replaced with sub-sampled ones, now ~960 lines-

guillaumevernieres (approved):

PR #1250, opened 2024-08-14
fix for FMS paths that are too long in snow YAMLs

CoryMartin-NOAA:
FMS has a maximum of 128 characters for filenames... so instead of using full paths we will use relative paths for FMS and might as well do it with other things to be consistent.


jiaruidong2017 (approved):
Looks good to me. Thanks.

PR #1253, opened 2024-08-14
Add DA-Utils to GDASApp

CoryMartin-NOAA:
DA-Utils is intended to house EMC-specific code that can be shared across all DA apps. Things like, for example, IODA obs space statistics processing code.


ClaraDraper-NOAA (approved):
Looks good to me!

PR #1255, opened 2024-08-16
update da-utils hash

CoryMartin-NOAA:
da-utils hash was wrong and not building, this fixes that


RussTreadon-NOAA (approved):
Test change on Hera.  `./build.sh -f -v` works!

Approve.

RussTreadon-NOAA:
Thank you @CoryMartin-NOAA for the quick fix!

PR #1257, opened 2024-08-20
Yamls for MIRS type sea-ice obs in cycling test

apchoiCMD:
#### This PR is to add the yaml key for cycling test

- Yaml keys for MIRS_NPP Ioda Converters
- Generate `gdas.t06z.icec_mirs.2021063006.nc4` for the cycling test

#### Result looks like below
GFSv17-3DVAR-C384mx025_gdasprepoceanobs_202106300600


JohnSteffen-NOAA (dismissed):
Looks good.

guillaumevernieres (commented):
Looks fine to me but was this tested @apchoiCMD ?

AndrewEichmann-NOAA (dismissed):
looks good

parm/soca/obs/obs_list.yaml
```diff
@@ -15,6 +15,7 @@ observers:
 # Ice concentration
 - !INC ${OBS_YAML_DIR}/icec_amsr2_north.yaml
 - !INC ${OBS_YAML_DIR}/icec_amsr2_south.yaml
+- !INC ${OBS_YAML_DIR}/icec_mirs.yaml
```
guillaumevernieres: could you comment that out @apchoiCMD , I don't think we want this to be part of the default obs list setup yet.

guillaumevernieres (changes requested):
Sorry for the late review @apchoiCMD , just a minor comment related to the obs list.

parm/soca/obs/obs_list.yaml
```diff
@@ -15,6 +15,7 @@ observers:
 # Ice concentration
 - !INC ${OBS_YAML_DIR}/icec_amsr2_north.yaml
 - !INC ${OBS_YAML_DIR}/icec_amsr2_south.yaml
+#- !INC ${OBS_YAML_DIR}/icec_mirs.yaml
```
guillaumevernieres: sorry for catching this just now @apchoiCMD , but the file names above do not fit our convention, see how sst is constructed: {variable}_{instrument}_{satellite}_{level}.yaml
We'll need to make these changes to the AMSR2 yaml as well but in a different PR.

guillaumevernieres (changes requested):
The naming convention you use is not correct, see comment.

parm/soca/obs/obs_list.yaml
apchoiCMD: Good point, I will modify accordingly-

guillaumevernieres: sorry, I'm just realizing that the markdown rendering didn't do what I expected.
We should follow the convention we have for the sst yaml's, for example:
`sst_avhrr_ma_l3u.yaml`
variable_instrument_satellite_level.yaml

apchoiCMD: made changes and could you please take a look?

AndrewEichmann-NOAA (dismissed):
Looks right

guillaumevernieres (approved):
Looks good, thanks @apchoiCMD .

JohnSteffen-NOAA (approved):

PR #1277, opened 2024-09-09
Add simple stats yaml

kevindougherty-noaa:
Add simple stats yaml to parm for global-workflow work.


parm/stats/stat_analysis.yaml.j2
```diff
@@ -0,0 +1,3 @@
+time window:
```
CoryMartin-NOAA: I think we are going to need one of these for each component, let's discuss later this morning

CoryMartin-NOAA (commented):
Each file is missing an End of Line. I don't think we should merge these stubs in yet until you've been able to test but you should be able to use this branch for your testing/development

CoryMartin-NOAA:
@kevindougherty-noaa has this been superseded by JCB work, and thus we can close this?

kevindougherty-noaa:
@CoryMartin-NOAA Yes, I will close now.

PR #1259, opened 2024-08-21
Add JPSSRR sea-ice product to ioda converter

apchoiCMD:
#### Description
- Task for adding a new ioda converter for new sea-ice concentration
- For sea-ice, AMSR2, MIRS and JPSSRR product will be used

Partially addressed #1182

PR #1300, opened 2024-09-26
Yamls for JPSSRR sea-ice obs in cycling test

apchoiCMD:
#### This PR is to add the yaml keys
- Yaml keys for JPSSRR sea-ice for the cycling
- The file names of JPSSRR type sea-ice are the same in `NH` and `SH`, please see in `yamls` and `obsprep_config.yaml`
- The preliminary results here/below
GFSv17-3DVAR-C384mx025_gdasprepoceanobs_202106300600_NH
- Obs for the SH is not ready yet

Related to https://github.com/NOAA-EMC/GDASApp/pull/1259


parm/soca/obs/config/icec_viirs_n20_l2_north.yaml
```diff
@@ -0,0 +1,45 @@
...
+obs filters:
+- filter: Domain Check
+  where:
+  - variable: {name: GeoVaLs/sea_area_fraction}
+    minvalue: 0.9
+- filter: Bounds Check
+  minvalue: 0.0
+  maxvalue: 1.0
+#- filter: Background Check
+#  threshold: 5.0
```
JohnSteffen-NOAA, line 30: Do we need to keep this background check filter commented out? If not, then delete.

parm/soca/obs/config/icec_viirs_n20_l2_south.yaml
```diff
@@ -0,0 +1,45 @@
...
+obs filters:
+- filter: Domain Check
+  where:
+  - variable: {name: GeoVaLs/sea_area_fraction}
+    minvalue: 0.9
+- filter: Bounds Check
+  minvalue: 0.0
+  maxvalue: 1.0
+#- filter: Background Check
+#  threshold: 5.0
```
JohnSteffen-NOAA, line 30: Same comment as the *north.yaml

parm/soca/obs/config/icec_viirs_n20_l2_south.yaml
```diff
@@ -0,0 +1,45 @@
...
+- filter: Domain Check
```
JohnSteffen-NOAA, line 31: Why is there a domain check for sea_surface_temperature > 2.0 and also > 0? Isn't one sufficient?

parm/soca/obs/config/icec_viirs_n20_l2_north.yaml
apchoiCMD: Good point but now as same as what `amsr2` has

parm/soca/obs/config/icec_viirs_n20_l2_south.yaml
apchoiCMD: the same answer above

parm/soca/obs/obs_list.yaml
```diff
@@ -15,6 +15,8 @@ observers:
 # Ice concentration
 - !INC ${MARINE_OBS_YAML_DIR}/icec_amsr2_north.yaml
 - !INC ${MARINE_OBS_YAML_DIR}/icec_amsr2_south.yaml
+#- !INC ${MARINE_OBS_YAML_DIR}/icec_viirs_n20_l2_north.yaml
```
KatherineLukens-NOAA, line 18: Did you mean to comment this out? Same question for "south"

apchoiCMD: Not part of `CI` yet, we will need to comment out when we have to use them-

guillaumevernieres:
Does the figure above has all the obs for the cycle? If so, there's probably a bug somewhere @apchoiCMD .

AndrewEichmann-NOAA (commented):
Nothing to add

guillaumevernieres (approved):
:+1:

apchoiCMD:
> Does the figure above has all the obs for the cycle? If so, there's probably a bug somewhere @apchoiCMD .

GFSv17-3DVAR-C384mx025_gdasprepoceanobs_20240618 00Z_SH

From the original files, zero values appears-

PR #1312, opened 2024-10-08
JCB-based obs+bias staging

DavidNew-NOAA:
This PR is a companion to Global-Workflow PR [#2992](https://github.com/NOAA-EMC/global-workflow/pull/2992) and JCB-GDAS PR [#31](https://github.com/NOAA-EMC/jcb-gdas/pull/31).

It adds two parameters required for the above Global-Workflow PR to the JCB base YAML for the atmosphere.


CoryMartin-NOAA (dismissed):

danholdaway (dismissed):

DavidNew-NOAA:
@CoryMartin-NOAA @danholdaway @RussTreadon-NOAA Can I get a final review on this, so the companion G-W can move forward? I already have a new G-W PR in preparation that depends on it.

Note (FYI @RussTreadon-NOAA):  This PR will temporarily break the jjob tests because of changes in how environment variables in the config.* files of G-W are named/defined, but reverting ```test/atm/global-workflow/config.yaml``` will fix this if you need to run jjob tests.

RussTreadon-NOAA:
@DavidNew-NOAA , g-w PR [#2978](https://github.com/NOAA-EMC/global-workflow/pull/2978) breaks various job.  I'm working on resolving these failures.

parm/atm/jcb-base.yaml.j2
```diff
@@ -94,6 +98,16 @@ atmosphere_obsbiasout_suffix: ".satbias.nc"
 atmosphere_obsbiascovout_prefix: "{{APREFIX}}"
 atmosphere_obsbiascovout_suffix: ".satbias_cov.nc"
 
+bias_files:
```
RussTreadon-NOAA, line 101: Seems users will need to edit this file every time they add a new _sensor_satellite_ even though all radiance bias correction files are in the same tarball.  Is there a way to get around this?

DavidNew-NOAA:
Thanks for the update, @RussTreadon-NOAA

RussTreadon-NOAA (commented):
Looks good.  Question about listing all _sensor_satellite_ in `parm/atm/jcb-base.yaml.j2`

parm/atm/jcb-base.yaml.j2
DavidNew-NOAA: The alternative would be to place the name of the tarball in the observation template itself, for each observation. @CoryMartin-NOAA 's concern was that this would interfere with the JEDI YAML reader (see https://github.com/NOAA-EMC/jcb-gdas/pull/31#issuecomment-2400784464). I actually don't think this will be the case, but my issue is more that the observation template should contain the information JEDI needs to run, not information that G-W needs for staging.

Ultimately, since there will be at least one more tarball for other observations (namely aircraft), information about what tarball is associated with which observation is going to have to be stored in some list.

CoryMartin-NOAA: I think, while @RussTreadon-NOAA makes a good point, I think this approach is the most explicit and gives us the most flexibility. For example, this means, if we decide in the future we want to group the varbc files by sensor or platform we could do so. I could see benefit in doing it by platform so that we can expect when new coefficients need to be spun up when a new platform comes online

RussTreadon-NOAA: @DavidNew-NOAA and @CoryMartin-NOAA , thank you for your replies.  We can / will likely revisit this as we add aircraft bias correction and start cycling.

RussTreadon-NOAA (approved):
Approve

CoryMartin-NOAA (approved):

DavidNew-NOAA:
Thanks @RussTreadon-NOAA and @CoryMartin-NOAA for the quick action on this

PR #1339, opened 2024-10-18
Add variable rescaling for aerosol species

andytangborn:
This is a draft PR for additions to chem_diagb.h that could allow for a variable rescaling (rs) for each aerosol species. It does not yet include a way to read the values in from a yaml or config file. The it compiles and runs, but I have not yet tested it using 3dvar.  And it still has lots of print statements used for testing. It runs in about 40 seconds at C96, and is probably slowed down by the print statements.   Any comments will be helpful.   

Also, note the if statement. I was unsure of the dimensions of an array, so I put in a test. I'll remove it along with the print statements.


utils/chem/chem_diagb.h
```diff
@@ -230,15 +236,66 @@ namespace gdasapp {
...
+
       // Rescale
+      //
       if (fullConfig.has("rescale")) {
-        double rescale;
-        fullConfig.get("rescale", rescale);
-        util::multiplyFieldSet(bkgErrFs, rescale);
-      }
+    // Define rescaling factors for each species
+        std::vector<double> rescaleFactors = {.5, 1.0, 1.0, 1.0, 1.0, 2.0, 2.0, 2.0, 2.0, 2.0, 1.0, 1.0, 1.0, 1.0};
```
CoryMartin-NOAA: This should be read in from the YAML that way we can test/tune without needing to recompile

CoryMartin-NOAA: Never mind, I didn't read, you said that in the description :-)

andytangborn: Yes, I'm working on that next.

utils/chem/chem_diagb.h
```diff
@@ -230,15 +236,66 @@ namespace gdasapp {
...
+      for (size_t i = 0; i < chemVars.size(); ++i) {
+        std::cout << "Variable " << i << ": " << chemVars[i].name() << std::endl;
+}
+
       // Rescale
+      //
       if (fullConfig.has("rescale")) {
-        double rescale;
-        fullConfig.get("rescale", rescale);
-        util::multiplyFieldSet(bkgErrFs, rescale);
```
CoryMartin-NOAA, line 237: What about using `multiplyFieldSets`? I think this would require creating a 3D array of values in a new field set, but that would then set us up for the ability to scale by height/latitude/location at a later date. Thoughts?

andytangborn: I can look at it and see how that would work. It took me some time to make something that could run....

andytangborn: I did try multiplyFieldSets, but had trouble getting it to compile.

CoryMartin-NOAA: I wonder if we just read in another 'state' that is the scaling factors and be proactive on the 3D front?

andytangborn: So have a file with the scaling factors, and read it in from there?

CoryMartin-NOAA: I think so, that way we can easily scale by lat,lon,height and species

andytangborn: Oh, I think I understand. A full background state file, filled with rescaling values for each aerosol species and at all locations.   I'd like to try just 14 values first and see if we can do better.

CoryMartin-NOAA: yeah they can be constant in x,y,z to start

utils/chem/chem_diagb.h
```diff
@@ -78,6 +78,11 @@ namespace gdasapp {
       oops::Log::info() << "Background:" << std::endl;
       oops::Log::info() << xb << std::endl;
 
+      /// Read the global rescale
```
CoryMartin-NOAA: Remove this section that is added?

CoryMartin-NOAA (commented):
One minor cleanup thing and then I think this is ready to go

utils/chem/chem_diagb.h
andytangborn: Just those 4 lines?

andytangborn: L81-L84

CoryMartin-NOAA: and 85

utils/chem/chem_diagb.h
```diff
@@ -230,13 +230,38 @@ namespace gdasapp {
...
+        const eckit::LocalConfiguration GlobalRescaleGeomConfig(GlobalRescaleConfig, "geometry");
+        const fv3jedi::Geometry GlobalRescaleGeom(GlobalRescaleGeomConfig, this-> getComm());
+        fv3jedi::Increment global_rescale(GlobalRescaleGeom, chemVars, cycleDate);
+        global_rescale.zero();
+        const eckit::LocalConfiguration GlobalRescaleStdConfig(GlobalRescaleConfig,
+                                                    "rescale stddev");
+        // Get the 'datapath' and 'filename_trcr' from the YAML configuration
+        std::string datapath, filename_trcr;
+        GlobalRescaleStdConfig.get("datapath", datapath);
+        GlobalRescaleStdConfig.get("filename_trcr", filename_trcr);
```
CoryMartin-NOAA: making a note, that if we were to use 'histories' and not 'restarts' this would cause it to fail. Do we need the filename printed out?

andytangborn: Sure. Like 

oops::Log::info() <<"filename_trcr", filename_trcr << std::endl;     ?

CoryMartin-NOAA: my point was that if you use a different IO class, `filename_trcr` won't exist, so it will fail.

utils/chem/chem_diagb.h
```diff
@@ -230,13 +230,38 @@ namespace gdasapp {
...
+
+        // Combine the path and filename to get the full file path
+        std::string fullPath = datapath + filename_trcr;
+
+        // Print out the full path to verify
+        std::cout << "Attempting to read the global rescale file from: " << fullPath << std::endl;
```
CoryMartin-NOAA:
```suggestion
```

CoryMartin-NOAA (approved):

PR #1365, opened 2024-11-08
Add more CI to GW-CI

DavidNew-NOAA:
This PR is a companion to GW PR [#3120](https://github.com/NOAA-EMC/global-workflow/pull/3120).

It does a couple things:

1. 5 GW CI tests are added/extended as CTests in GDASApp, running through to the fcst jobs in the first full-cycle. These CI tests are: ```C96C48_hybatmDA```, ```C96C48_ufs_hybatmDA```, ```C96C48_hybatmaerosnowDA```,  ```C48mx500_3DVarAOWCDA```, and ```C48mx500_hybAOWCDA```.
2. Test references are added for ```C96C48_ufs_hybatmDA``` and ```C48mx500_3DVarAOWCDA```, so that we're actually testing the output.
3. These CTests are turned on by default in a workflow build, rather than having to mess with the ```CMakeCache.txt``` file and re-running make. This will allow us to use these tests in nightly testing.
4. ```test/gw-ci/CMakeLists.txt``` is refactored quite a bit.
5. There are 89 CTests, but for 5 CI tests, but I added task dependencies, so they can be run in parallel.

The primary motivation for this PR is that we can run CI for our nightly testing of GDASApp. Also, anyone with a PR can easily do CI testing through CTests.


test/gw-ci/CMakeLists.txt
```diff
@@ -99,40 +202,6 @@ if (WORKFLOW_TESTS)
...
-  add_cycling_tests(${pslot} ${YAML_PATH} ${HOMEgfs} ${RUNTESTS} ${PROJECT_SOURCE_DIR} "${TASK_LIST}")
-
-  # Atm DA, C96/C48
-  # ---------------
-  set(pslot "Atm-hyb-C96C48")
-  set(YAML_PATH ${HOMEgfs}/ci/cases/pr/C96C48_ufs_hybatmDA.yaml)
-  set(TASK_LIST)  # empty list for now
-  add_cycling_tests(${pslot} ${YAML_PATH} ${HOMEgfs} ${RUNTESTS} ${PROJECT_SOURCE_DIR} "${TASK_LIST}")
-
-  # GFSv17, 3DVAR prototype
```
guillaumevernieres, line 122: Maybe I missed it, but the gfsv17 prototype test doesn't seem to be in.
Add it back and wrap it under a specific flag so we don't run it for every pr.

guillaumevernieres (commented):
Looks good @DavidNew-NOAA  except for the missing gfsv17 test (c384/0.25). We don't want to run it a every pr's but it certainly needs to be checked regularly.

test/gw-ci/CMakeLists.txt
DavidNew-NOAA: Ah yes, OK, will do

RussTreadon-NOAA:
Question:  do we expect reference checks to pass on all platforms?  If not, on which platform(s) are the reference checks to be performed?

DavidNew-NOAA:
@RussTreadon-NOAA I would expect them to pass within the tolerance that we set

test/gw-ci/CMakeLists.txt
DavidNew-NOAA: @guillaumevernieres OK, I put it back under the flag ```TEST_GFS17```

test/gw-ci/CMakeLists.txt
```diff
@@ -18,78 +48,151 @@ function(add_cycling_tests pslot YAML_PATH HOMEgfs RUNTESTS PROJECT_SOURCE_DIR T
...
+  set(HALF_CYCLE_TASKS
+    "gdas_stage_ic"
+    "gdas_fcst"
+    "gdas_atmos_prod"
+    "gdas_aeroanlgenb"
+    "enkfgdas_stage_ic"
+    "enkfgdas_fcst"
+    "enkfgdas_echgres"
+    "enkfgdas_epmn")  
+  set(FULL_CYCLE_TASKS
```
CoryMartin-NOAA, line 323: any chance there is a way to automate this so that we don't have to manually change these if/when jobs are added/removed/renamed?

DavidNew-NOAA: @CoryMartin-NOAA Yeah, I thought about that. Could probably do the same thing ```setup_xml.py``` does to check dependencies. The problem is that I want to skip certain jobs like archiving.

CoryMartin-NOAA: a list of jobs to exclude, but otherwise it checks dependencies and includes all other jobs? or is that pain not worth it? I just fear that we will be doing a lot of maintenance on this file in the future to keep it up to date

DavidNew-NOAA: I think maybe this can be an iterative process. I have another PR planned to test the memory and time usage, so maybe I can add it there. It will be a bit of a pain because I'll have to interface the CMake and Python

guillaumevernieres: Keep it simple and definitely don't add unnecessary obfuscation to the code! We've been using this for a while and the maintenance required was trivial.

DavidNew-NOAA: Yeah, I mean we'll know if something breaks when the tests stop passing

DavidNew-NOAA:
I'm marking this PR as ready for review, because it won't break anything that isn't already broken

test/gw-ci/CMakeLists.txt
```diff
@@ -1,11 +1,41 @@
+# Function that adds a test for a given GW task
+function(add_task task_name test_prefix cycle pslot RUNTESTS PROJECT_SOURCE_DIR)
```
CoryMartin-NOAA: thoughts on adding `run` also?

DavidNew-NOAA: @CoryMartin-NOAA What do you mean?

CoryMartin-NOAA: `gdas` or `enkfgdas` or `gfs`

CoryMartin-NOAA: in the workflow the full jobs are defined as `$run$task` so the task name is just `fcst` but the job would be `gdasfcst`

DavidNew-NOAA: I think that would probably add unnecessary complexity. The ```add_task``` function is meant to split metatasks into subtasks, which doesn't really matter for ```gfs```, because I think none of the  ```gfs```  tasks are indeed metatasks. If you want to add ```gfs``` tasks for example,  you would just add ```gfs_prep```, ```gfs_anal```, etc to the ```FULL_CYCLE_TASKS``` list

DavidNew-NOAA: And ```gdas``` and ```enkfgdas``` metatasks are very different

CoryMartin-NOAA (dismissed):

DavidNew-NOAA:
Reminder DO NOT MERGE this PR until GW PR [#2992](https://github.com/NOAA-EMC/global-workflow/pull/2992) is merged.

test/gw-ci/CMakeLists.txt
guillaumevernieres: ha, good idea @CoryMartin-NOAA . We keep breaking the gfs cycle, might be good to test every once in a while.
ps: doesn't have to be implemented in this pr

parm/atm/jcb-base.yaml.j2
```diff
@@ -27,6 +27,11 @@ final_prints_frequency: PT3H
 number_of_outer_loops: 2
 analysis_variables: [ua,va,t,ps,sphum,ice_wat,liq_wat,o3mr]
 
+# Testing
```
danholdaway, line 30: Since jcb-algorithms will default `do_testing` to False itself I think you can move this key into your test/*/global-workflow/jcb-*.yaml files instead. This way we won't see anything about testing above the testing directory.

parm/atm/jcb-prototype_3dvar-fv3inc.yaml.j2
```diff
@@ -0,0 +1,10 @@
+# Algorithm
+# ---------
+algorithm: fv3jedi_fv3inc_variational
+
+# Testing things
```
danholdaway, line 5: Similarly with these. Can we get away with only seeing all these things about testing in the JCB YAML files that live in the testing directory?

parm/atm/jcb-prototype_lgetkf-fv3inc.yaml.j2
```diff
@@ -0,0 +1,10 @@
+# Algorithm
+# ---------
+algorithm: fv3jedi_fv3inc_lgetkf
+
+# Testing things
```
danholdaway, line 5: ⬆️

parm/atm/jcb-base.yaml.j2
DavidNew-NOAA: @danholdaway Let's tag up and discuss in-person

danholdaway (dismissed):

DavidNew-NOAA:
@CoryMartin-NOAA @guillaumevernieres @RussTreadon-NOAA @danholdaway

I made a few modifications to this PR, namely adding the task dependencies for each CTest so that they can be run in parallel. There are 89 CTests for 5 different CI tests, and usually you can run the deterministic/variational and ensemble components in parallel, so if you run ```ctest -j10``` it goes pretty fast. 

I have to re-run the CTests one more time because my session timed out while I was away, but I just want to get your feedback on one thing. The task dependencies are hardcoded, which turns out to be like 150 lines of CMake, which is a bit embarrassing, but it works. The proper way to do it would be to write a Python script that interfaces with this file:

https://github.com/NOAA-EMC/global-workflow/blob/develop/workflow/rocoto/gfs_tasks.py

which would then be called by ```CMakeLists.txt``` to set the CTests dependencies. It's complicated though because some tasks depend on the availability of files in the COM directory rather than depending directly on tasks.

I think I want to just get this PR merged soon as is, because I'm going to make a GW companion PR, and it's not clear how long that will take to get reviewed. Any future PRs after this can be done strictly done on GDASApp and will go a lot faster.

RussTreadon-NOAA:
@DavidNew-NOAA , I agree with the incremental approach.  If what you have works, let's get it merged.

DavidNew-NOAA:
@CoryMartin-NOAA @RussTreadon-NOAA 

OK, all 89 CTest-based CI tests pass before the hash associated with PR [#1339](https://github.com/NOAA-EMC/GDASApp/pull/1339). With that hash merged in, I get the following error when running `gdas_aeroanlgenb` with CI test `C96C48_hybatmaerosnowDA` (cc @andytangborn):
```
yaml.constructor.ConstructorError: while constructing a mapping
in "<unicode string>", line 88, column 10:
пхр: {{ npx_rescale }}
found unhashable key
in "‹unicode string>", line 88, column 11:
пхр: {{ npx_rescale }}
```

I haven't tried re-running the CTests with GDASApp PR [#1380](https://github.com/NOAA-EMC/GDASApp/pull/1380) merged in. 

Should we merge this PR anyway and fix with another PR? 

Also, this PR now has a Global-Workflow companion [#3120](https://github.com/NOAA-EMC/global-workflow/pull/3120).

DavidNew-NOAA:
I'm keeping the DO NOT MERGE label while I do a bit more testing

DavidNew-NOAA:
@CoryMartin-NOAA @RussTreadon-NOAA @danholdaway @guillaumevernieres

OK, I think this PR is now ready for final review and merger.

I made some additional changes:

1. ```C48mx500_3DVarAOWCDA``` now has a test reference for ```marineanlvar```
2.  I added CTest dependencies for the ```atm_jjob``` tests so they can be run in parallel.
3. The locations of test references and test output files in GDASApp are centralized (```test/testreference``` and ```build/gdas/test/testoutput``` respectively) to make updating references a bit easier. This mimics what the various JEDI repos do.
4. I removed the old ```jediinc2fv3jedi.py``` script (which meant to delete a long time ago) and removed its CTest. I also removed the standalone CTest for ```gdas_fv3jedi_fv3inc.x``` since that application already tested twice in the ```atm_jjob``` and CI tests.
5. The nightly testing scripts will use 12 processors to run the CTests to take advantage of the new ability to parallelize the tests.

ci/run_ci.sh
```diff
@@ -61,7 +61,7 @@ module use $GDAS_MODULE_USE
 module load GDAS/$TARGET
 echo "---------------------------------------------------" >> $outfile
 rm -rf log.ctest
-ctest -R gdasapp --output-on-failure &>> log.ctest
+ctest -j12 -R gdasapp --output-on-failure &>> log.ctest
```
CoryMartin-NOAA: Can we make the number of tasks platform dependent?

`ctest -j$NTASKS_TESTS`...
In case we need fewer on Hera than Hercules, etc.

ci/run_gw_ci.sh
```diff
@@ -62,7 +62,7 @@ module use $repodir/sorc/gdas.cd/modulefiles
 module load GDAS/$TARGET
 echo "---------------------------------------------------" >> $outfile
 rm -rf log.ctest
-ctest -R gdasapp --output-on-failure &>> log.ctest
+ctest -j12 -R gdasapp --output-on-failure &>> log.ctest
```
CoryMartin-NOAA: same comment

test/atm/global-workflow/CMakeLists.txt
```diff
@@ -1,63 +1,86 @@
...
-add_test(NAME test_gdasapp_atm_jjob_ens_obs
-  COMMAND ${PROJECT_SOURCE_DIR}/test/atm/global-workflow/jjob_ens_obs.sh
-          ${PROJECT_BINARY_DIR} ${PROJECT_SOURCE_DIR}
-          WORKING_DIRECTORY ${PROJECT_BINARY_DIR}/test/atm/global-workflow/testrun)
-
-add_test(NAME test_gdasapp_atm_jjob_ens_sol
-  COMMAND ${PROJECT_SOURCE_DIR}/test/atm/global-workflow/jjob_ens_sol.sh
-          ${PROJECT_BINARY_DIR} ${PROJECT_SOURCE_DIR}
-          WORKING_DIRECTORY ${PROJECT_BINARY_DIR}/test/atm/global-workflow/testrun)
+ecbuild_add_test(TARGET test_gdasapp_setup_atm_cycled_exp
```
CoryMartin-NOAA, line 4: can we create a macro/function to simplify these lines and then loop to create  the tests? I ask because they follow a pattern, not sure if it's possible or not?

test/gw-ci/CMakeLists.txt
```diff
@@ -1,11 +1,220 @@
...
+  if(${is_full_cycle})
+    set(cycle ${FULL_CYCLE})
+  else()
+    set(cycle ${HALF_CYCLE})
+  endif()
+  
+  # Add Task
+  set(test_name ${test_prefix}_${task_name}_${cycle})
+  message(STATUS "preparing ${subtask_names} for ${test_prefix} ctest")
+  ecbuild_add_test(TARGET ${test_name}
```
CoryMartin-NOAA, line 201: ok so you basically have this here, so I suggest reproducing it in the aforementioned CMakeLists.txt

test/testreference/C96C48_ufs_hybatmDA_fv3inc.ref
CoryMartin-NOAA, line 1: empty file? intended to be empty?

test/testreference/C96C48_ufs_hybatmDA_lgetkf.ref
CoryMartin-NOAA, line 1: empty file?

ci/run_ci.sh
guillaumevernieres: ~Isn't that fixing a problem that probably doesn't exist? I don't know what the optimum thread number is, but it shouldn't be hpc dependent. The one possible issue is the memory required to run more threads, but that hasn't been an issue so far on the hpc we use.~

CoryMartin-NOAA: it's probably not an issue for ctest, but it's absolutely an issue and machine dependent when running `make`. But, this was just a suggestion, not a requirement

guillaumevernieres (dismissed):
Thanks @DavidNew-NOAA , looks good.

ci/run_ci.sh
guillaumevernieres: Scratch the above, I'm again selecting snippets of text and extrapolating the rest without reading .. I'm realizing this is for the test, not make.
@CoryMartin-NOAA , I don't see why make -jN should be machine dependent.

danholdaway (dismissed):
Thanks @DavidNew-NOAA. How long does it take to run all the ctests now? Not including the code building?

ci/run_ci.sh
CoryMartin-NOAA: @guillaumevernieres well Hera, for example, has fewer cores and less memory per login node than Hercules...

DavidNew-NOAA:
@danholdaway Let me run them all one more time to give you an exact answer

test/testreference/C96C48_ufs_hybatmDA_fv3inc.ref
DavidNew-NOAA: That reference file doesn't do anything, so deleted

test/testreference/C96C48_ufs_hybatmDA_lgetkf.ref
DavidNew-NOAA: Again as above, useless file

DavidNew-NOAA:
Let me run the test_gdasapp tests one more time, just for certainty. Then I'll merge

ci/run_ci.sh
guillaumevernieres: > @guillaumevernieres well Hera, for example, has fewer cores and less memory per login node than Hercules...

I haven't played with this in a while, but the optimum N is well below the number of available threads we have access to, regardless of what hpc. Memory wise, 12 threads is an issue on my laptop but has never been an problem on hpc. One possible issue is how io may differ between hpc ... But I still think this is over-engineering.  KIS if we can!

DavidNew-NOAA:
I broke something somewhere, causing all ```test_gdasapp``` tests to fail. working on it

CoryMartin-NOAA (approved):

DavidNew-NOAA:
@danholdaway My last run of the 119 ```test_gdasapp``` CTests took 1 hour and 23 minutes

test/gw-ci/CMakeLists.txt
```diff
@@ -99,40 +381,31 @@ if (WORKFLOW_TESTS)
...
-    "gdas_marineanlfinal"
-    "gdas_ocnanalvrfy"
-    "gdas_prep"
-    "gdas_anal"
-    )
-  add_cycling_tests(${pslot} ${YAML_PATH} ${HOMEgfs} ${RUNTESTS} ${PROJECT_SOURCE_DIR} "${TASK_LIST}")
+  option(TEST_GFS17 "Enable the GFSv17 prototype CI tests" OFF)
+  if(TEST_GFS17)
+    set(pslot "GFSv17-3DVAR-C384mx025")
+    set(YAML_PATH ${HOMEgfs}/ci/cases/gfsv17/${pslot}.yaml)
```
shlyaeva, line 392: I am trying to run this set of tests, and it can't find the `GFSv17-3DVAR-C384mx025.yaml` in `global-workflow/ci/cases/gfsv17`. I don't see this file on develop, and I don't see it in https://github.com/NOAA-EMC/global-workflow/pull/3120. What am I missing? (I have not yet tried to run with https://github.com/NOAA-EMC/global-workflow/pull/3120)

DavidNew-NOAA: @shlyaeva Yeah, that was a mistake on my part. I assumed the GFSv17 test name had the same naming convention as the low resolution marine CI. Thanks for pointing this out, I'll make a bugfix PR tomorrow morning.

shlyaeva: Oh good, thank you for the clarification! I am still figuring out how to run things, so I wasn't sure whether I missed something.

PR #1372, opened 2024-11-13
Fixes gridgen problem and localization for LETKF

AndrewEichmann-NOAA:
Adds the `gridgen.yaml ` back after being removed to` jcb-gdas`, to allow LETKF to work with a fix within GDASApp. This is a temporary fix until the LETKF task can be converted to using the jcb yamls.

Also adds `obs localization` to remaining obs config files - this will have to likewise be carried over to whatever is being used with the obs list.


apchoiCMD (approved):
It appears as intended.

guillaumevernieres (approved):
I'm merging this @AndrewEichmann-NOAA , but FYI the marine obs yaml in the gdasapp are/should be deprecated.

PR #1389, opened 2024-11-26
Update JCB-GDAS for Aerosol DA

CoryMartin-NOAA:
Depends on https://github.com/NOAA-EMC/jcb-gdas/pull/48


CoryMartin-NOAA:
Note to self (and reviewers) we will need to clean up most of these YAML templates once JCB migration is complete

TravisElless-NOAA (approved):
Changes allow aerosol DA to run. Can revisit and update as needed in future PRs when tuning is more complete

PR #1390, opened 2024-11-26
Make sure gfs_* DA jobs are not regression tests

DavidNew-NOAA:
This PR addresses the bug @RussTreadon-NOAA found that ```gfs_atmanlvar``` was being run as a regression test and using the same test reference as ```gdas_atmanlvar``` in GW PR [#3120](https://github.com/NOAA-EMC/global-workflow/pull/3120). 

See https://github.com/NOAA-EMC/global-workflow/pull/3120#issuecomment-2501783018

I've moved all activation of testing mode in JCB out of the JCB base YAMLs and into the JCB algorithm YAMLs. I test the ```RUN``` variables to make sure it's not equal to ```gfs```.

I re-ran all the regression tests, and they all passed.

### DIFF
--- a/parm/atm/jcb-base.yaml.j2
+++ b/parm/atm/jcb-base.yaml.j2
@@ -30,7 +30,7 @@ analysis_variables: [ua,va,t,ps,sphum,ice_wat,liq_wat,o3mr]
 # Testing
 # -------
 
-do_testing: {{ DO_TEST_MODE | default(false, true) }}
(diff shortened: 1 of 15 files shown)


RussTreadon-NOAA (approved):
Did not test but @DavidNew-NOAA did and found that all regression tests pass.

We need the changes in this PR to move g-w PR [#3120](https://github.com/NOAA-EMC/global-workflow/pull/3120) forward.

Approve.

PR #1394, opened 2024-12-02
Make separate JCB algorithm YAMLs for CI tests

DavidNew-NOAA:
The motivation for having separate JCB algorithm YAMLs for each CI test is that in the future, we may want different regression tests for various resolutions, obs configurations, etc. The primary JCB algorithm YAMLs should be generic.


test/atm/global-workflow/jcb-prototype_3dvar-fv3inc.yaml.j2
```diff
@@ -4,7 +4,7 @@ algorithm: fv3jedi_fv3inc_variational
 
 # Testing things
 # --------------
-{% if DO_TEST_MODE and RUN == 'gdas'  %}
+{% if DO_TEST_MODE and RUN == 'gdas' and current_cycle | to_YMDH == '2021032318' %}
```
DavidHuber-NOAA: I recommend not relying on Jinja's implicit boolean checks and instead to check them explicitly.  This is because the wxflow extension of Jinja will initialize uninitialized values to, e.g., `"{{ DO_TEST_MODE }}"`, which `if` will evaluate as `True`.  Instead, use `if DO_TEST_MODE == True and ....`.  I'm going to open a wxflow issue to see if this can be changed so undefined values will return `False`.

danholdaway (dismissed):
Do you want to further your if statement around the testing to also include resolution? Especially since that's hardcoded in the path for the reference files. And/or template the reference file names to include the resolution?

DavidNew-NOAA:
@danholdaway Interesting idea. Let me give that a try

guillaumevernieres (dismissed):
Thanks @DavidNew-NOAA . That should fix the issue that we have when testing the 3dfgat in static and hybrid mode, it currently points to the same ref file.

DavidNew-NOAA:
@danholdaway OK, I implemented your suggestion

test/atm/global-workflow/jcb-prototype_3dvar-fv3inc.yaml.j2
DavidHuber-NOAA: After some wxflow testing, I see that this is not true (or no longer true).  That said, there is no harm in explicit boolean tests.

DavidHuber-NOAA (approved):

test/atm/global-workflow/jcb-prototype_3dvar-fv3inc.yaml.j2
DavidNew-NOAA: Thanks @DavidHuber-NOAA !

guillaumevernieres (dismissed):
:+1:

guillaumevernieres (approved):
:+1:

test/atm/global-workflow/jcb-prototype_3dvar.yaml.j2
```diff
@@ -17,7 +17,7 @@ atmosphere_obsdatain_suffix: ".{{ current_cycle | to_YMDH }}.nc"
 
 # Testing things
 # --------------
-{% if DO_TEST_MODE and RUN == 'gdas'  %}
+{% if DO_TEST_MODE == true and RUN == 'gdas' and current_cycle | to_YMDH == '2021032318' %}
```
CoryMartin-NOAA, line 20: suggestion for future, what about making the test reference file names include RUN and cycle times?

CoryMartin-NOAA (approved):

test/atm/global-workflow/jcb-prototype_3dvar.yaml.j2
DavidNew-NOAA: Good idea

guillaumevernieres:
I did approve, but that was with the understanding that we should be running the gdasapp test @DavidNew-NOAA . Are these broken? I see that the label I set in my pr has not been updated.

DavidNew-NOAA:
@guillaumevernieres I'm actually refactoring and testing those scripts right now, so they won't work right now with the CI labels. I wasn't aware anyone was using those scripts currently.

PR #1395, opened 2024-12-02
update hash for land-imsproc

CoryMartin-NOAA:
Update the hash for land-imsproc to use `snowdl` instead of `snwdph` which is needed for the ensemble mean update.


CoryMartin-NOAA:
This is in draft until I can test it and make sure everything is working as intended.

CoryMartin-NOAA:
Superseded by #1410

Issue #1404, opened 2024-12-06
Wrong dependency in the marine hyb. var. ctest


guillaumevernieres:
Thanks for catching this @RussTreadon-NOAA .

The bmat job needs to wait until the ens. fcst is done.

From @DavidNew-NOAA , example of the implementation of a fix is here:
https://github.com/NOAA-EMC/GDASApp/blob/ea25002a066a42986ae5c5dd030713c39288bbca/test/gw-ci/CMakeLists.txt#L103

PR #1405, opened 2024-12-06
Encapsulate the marine 3dvar and hyb ctests under different cmake conditionals

guillaumevernieres:
That's not fixing #1404 , but that should allow the nightly `ci` to not complain while @AndrewEichmann-NOAA or myself  figure out a fix.


DavidNew-NOAA (approved):

RussTreadon-NOAA (approved):
Didn't test but looks OK.   Thanks for making a quick change @guillaumevernieres

guillaumevernieres:
Thanks for the quick review @RussTreadon-NOAA and @DavidNew-NOAA . I'll configure and check that the correct list of ctest is generated before I merge this.

guillaumevernieres:
I played with the switches in the CMakeCache, it does what we want so merging!

PR #1409, opened 2024-12-10
MIRS ICEC filename flexibility

JohnSteffen-NOAA:
The filename nomenclature for MIRS sea ice concentration is modified to use a wildcard in place of the hardcoded "revision" number since this changes annually from the data provider.


apchoiCMD (approved):

guillaumevernieres (approved):
thanks @JohnSteffen-NOAA

PR #1417, opened 2024-12-16
Add u,v to the MOM6 IAU increment if it is not part of the DA control variables

guillaumevernieres:
Mostly as the title says, this branch will be used for the next batch of experiments (cp4). 
Only a draft because I need to update the #'s for `jcb-algorithm` and `jcb-gdas`.

- [ ] atm_jjob 
- [ ] C96C48_ufs_hybatmDA 
- [ ] C96C48_hybatmaerosnowDA  
- [x] C48mx500_3DVarAOWCDA 
- [x] C48mx500_hybAOWCDA 
- [ ] C96C48_hybatmDA


utils/soca/gdas_postprocincr.h
```diff
@@ -92,42 +92,49 @@ class PostProcIncr {
   }
 
   // -----------------------------------------------------------------------------
-  // Append layer thicknesses to increment
-  // TODO(guillaume): There's got to be a better way to append a variable.
-  soca::Increment appendLayer(const soca::Increment& socaIncr) {
+  // Append variable to increment
+  soca::Increment appendVar(const soca::Increment& socaIncr, const oops::Variables varToAppend) {
```
shlyaeva, line 96: Not for this PR: I agree, there's gotta be a better way to append a variable. Is this meant to add new zero variables? Looking through soca code, perhaps `soca::Increment::updateFields` could be enough? It looks like it's creating new variables if the argument contains new variables, and I think from the code that the fields for new variables would be set to zero.

shlyaeva (dismissed):
Did not run, only reviewed code changes, they look reasonable to me!

apchoiCMD (dismissed):

utils/soca/gdas_postprocincr.h
guillaumevernieres: @shlyaeva , It was meant to append the layer thickness state to the mom6 iau increment at first, but I now need to add u,v if it's not in the soca incr. I'll have a look at `updateFields`

shlyaeva: no worries about looking into it now. I am also happy to look into this in the next few days/weeks, let me know.

guillaumevernieres: We'll have to revisit this at some point for sure. The LETKF increment isn't used currently, it needs to be combined with the re-centering, which is the only thing that we currently apply to the ensemble members.

AndrewEichmann-NOAA (dismissed):
I'm not familiar with this code, but I see nothing objectionable

shlyaeva (approved):

apchoiCMD (approved):
Looks good-

PR #1445, opened 2025-01-15
Feature/b2iron

givelberg:
# Description


guillaumevernieres:
We talked about it multiple times @givelberg : Your work goes in `spoc`. Ping @rmclaren if you need guidance.

Issue #750, opened 2023-11-20
[End-to-End Test Code Sprint] Update GNSSRO converter and YAML


XuanliLi-NOAA:
GNSSRO API updates:

Updated the pccf flag to use the profile pccf for CDAAC RO data
Updated the qfro flag to use the profile qfro for EUMETSAT RO data 
Updated the satelliteAscending flag:  convert the 3rd bit of qfro to satelliteAscending flag 
Updated some information in the json file
Added QC filters in config gnssro yaml.

XuanliLi-NOAA:
GSI run obs vs. hofx plot before QC:
![scatter_obs_hofx_gsi](https://github.com/NOAA-EMC/GDASApp/assets/101414760/1ebc8b84-5301-4dc4-8b24-4e42d8c59f2f)

GSI run obs vs. hofx plot after QC:
![scatter_obs_hofx_gsi_qc](https://github.com/NOAA-EMC/GDASApp/assets/101414760/137818f1-869b-4354-9daf-a853f3fbe414)

XuanliLi-NOAA:
JEDI run obs vs. hofx plot before QC:
![scatter_obs_hofx0_fv3](https://github.com/NOAA-EMC/GDASApp/assets/101414760/0677cbf5-9e1f-4a4e-ad4b-323fc5f837d9)

JEDI run obs vs. hofx plot after applying pccf and qfro:
![scatter_obs_hofx0_fv3_pc](https://github.com/NOAA-EMC/GDASApp/assets/101414760/858dd056-38fc-4786-aacd-f6c5c54f77c2)

JEDI run obs vs. hofx plot after all QCs:
![scatter_obs_hofx0_fv3_qc](https://github.com/NOAA-EMC/GDASApp/assets/101414760/67993569-59a3-437c-aa7c-00402ee35325)

XuanliLi-NOAA:
GSI run hofx vs. JEDI run hofx before QCs:
![scatter_hofxgsi_hofx0fv3](https://github.com/NOAA-EMC/GDASApp/assets/101414760/eeb095f3-fc15-4520-bee8-d25835d8f087)

GSI run hofx vs. JEDI run hofx after QCs:
![scatter_hofxgsi_hofx0fv3_qc](https://github.com/NOAA-EMC/GDASApp/assets/101414760/ceb60634-8216-4369-be4a-67b010c835b5)


[...]

PR #1448, opened 2025-01-16
Add GNSSRO bufr2ioda and json files

XuanliLi-NOAA:
# Description

Add JSON and bufr2ioda python scripts for GNSSRO data. Each GNSSRO mission has a separate ioda obs file.  

# Companion PRs

N/A

# Issues

Refs #750 

- [ ] atm_jjob 
- [ ] C96C48_ufs_hybatmDA 
- [ ] C96C48_hybatmaerosnowDA  
- [ ] C48mx500_3DVarAOWCDA 
- [ ] C48mx500_hybAOWCDA 
- [ ] C96C48_hybatmDA


CoryMartin-NOAA:
Should these go in here or in `spoc`? @emilyhcliu what are your thoughts?

emilyhcliu:
@CoryMartin-NOAA 
The json and Python scripts in this PR are written in bufr-query low-level Python API.
These are good to use with the current setup in GDASApp.  So, they should be checked in here in GDASApp so Xuanli can keep track of changes and repeat her tests.

In the meanwhile, @nicholasesposito will create these low-level Pythons into one Python script with high-level API and mapping files.  These will be checked into SPOC.  

Once we have most of the Python Script, mapping files in SPOC, we will work on GDASApp to use them from SPOC.

CoryMartin-NOAA:
ok sounds good thanks @emilyhcliu I just wanted to double check

parm/atm/jcb-prototype_3dvar.yaml.j2
```diff
@@ -14,7 +14,18 @@ observations:
...
+  - gnssro_s6
+  - gnssro_geoopt
+  - gnssro_grace
+  - gnssro_k5
+  - gnssro_metop
+  - gnssro_paz
+  - gnssro_piq
+  - gnssro_tsx
+  - gnssro_tdm
+#  - gnssro
```
CoryMartin-NOAA:
```suggestion
```

ush/ioda/bufr2ioda/bufr2ioda_gnssro_cosmic2.py
```diff
@@ -0,0 +1,805 @@
+#!/usr/bin/env python3
+# (C) Copyright 2023 NOAA/NWS/NCEP/EMC
```
CoryMartin-NOAA: if we are putting dates in all of these, they should probably be 2025 not 2023. We can also just not put dates in it at all, we don't need to put a license string in every file like the JCSDA does

CoryMartin-NOAA (commented):
a few minor comments, and I didn't closely look at the converters but assume they have been thoroughly tested

emilyhcliu:
@XuanliLi-NOAA  one question.
Is it possible to write one Python to convert all platforms?
GPSRO data is in one subset.  It should be possible to write one Python to handle all platforms.
Are these platforms using the same BUFR table?

No worries, what you have in this PR is OK. But. we should try to create one Python script with high-level Python API for GPSRO. 
@nicholasesposito will work with you on this.

ush/ioda/bufr2ioda/bufr2ioda_gnssro_cosmic2.py
XuanliLi-NOAA: The date line has been removed from the python scripts.

XuanliLi-NOAA:
@emilyhcliu , it would be great to have one Python script to convert all platforms. Yes, let's do that with the with high-level Python API.

emilyhcliu:
> @emilyhcliu , it would be great to have one Python script to convert all platforms. Yes, let's do that with the with high-level Python API.

Do all GPSRO platform use the same bufr table?

XuanliLi-NOAA:
@emilyhcliu, yes, for the test date 00z 2024-02-19, they use the same bufr table. But after June 2024, we have new EUMETSAT Spire data, this has an additional BUFR mnemonic "SASBID". Should we consider that in the high level python API?

@XuanliLi-NOAA   Yes, we should include the new one.  It is good that all platforms share the same bufr table mnemonics.  We can have one mapping file to handle all of them.  So, the one Python script with high-level API should be doable for GPSRO.
Thanks for the information.

@emilyhcliu Sounds good, thank you for confirming that.

parm/ioda/bufr2ioda/bufr2ioda_gnssro_cosmic2.json
```diff
@@ -0,0 +1,20 @@
+{
+  "data_format"      : "bufr_d",
+  "data_type"        : "gpsro",
+  "cycle_type"       : "{{ RUN }}",
+  "cycle_datetime"   : "{{ current_cycle | to_YMDH }}",
+  "dump_directory"   : "{{ DMPDIR }}",
+  "ioda_directory"   : "{{ COM_OBS }}",
+  "subsets"          : [ "NC003010" ],
+  "data_description" : "Satellite radio occultation data", 
+  "data_provider"    : "U.S. NOAA",
```
nicholasesposito: For the data_provider, I believe we are using "U.S. NOAA NCEP". I won't comment on every single one.

ush/ioda/bufr2ioda/bufr2ioda_gnssro_cosmic2.py
```diff
@@ -0,0 +1,804 @@
...
+    path, fname = os.path.split(OUTPUT_PATH)
+    if path and not os.path.exists(path):
+        os.makedirs(path)
+
+    # Create IODA ObsSpace
+    obsspace = ioda_ospace.ObsSpace(OUTPUT_PATH, mode='w', dim_dict=dims)
+
+    # Create Global attributes
+    logger.debug(f" ... ... Create global attributes")
+    obsspace.write_attr('source_file', bufrfile)
```
nicholasesposito, line 544: I think the global attributes need to be closer to the standards.
At a minimum, 
data_format -> dataOriginalFormatSpec
subsets -> platforms
data_type should be deleted altogether since it's redundant with "source"
data_provider -> dataProviderOrigin

I'm unsure of the others. but I would look at the global attributes tab of this document to help figure it out.
https://docs.google.com/spreadsheets/d/1fHb2QIHOwgp4d6plyn9Dj9HZFZsDRxcPaUV6hdTb5BU/edit?gid=46200857#gid=46200857

@emilyhcliu do you agree?

I won't mark each script.

nicholasesposito (commented):
Recommend changing the global attributes, but @emilyhcliu should comment on that. The rest looked fine with the exception of I think changing the data_format description.

ush/ioda/bufr2ioda/bufr2ioda_gnssro_cosmic2.py
emilyhcliu: I think fixing it to the minimum as @nicholasesposito  is good.
@XuanliLi-NOAA could you modify three global attribute name based on nick's suggestions.  
Thanks!

XuanliLi-NOAA:
@emilyhcliu and @nicholasesposito _data_provider_ has been corrected in the json files to reflect the actual originating center. _data_format_ has been changed to _dataOriginalFormatSpec_ in the python scripts. I'm not sure if we should replace _subsets_ with _platforms_, as _subsets_ is a bufr message for the name of the subset.

ush/ioda/bufr2ioda/bufr2ioda_gnssro_cosmic2.py
```diff
@@ -0,0 +1,804 @@
...
+    sclf = r.get('satelliteConstellationRO', 'latitude')
+    ptid = r.get('satelliteTransmitterId', 'latitude')
+    elrc = r.get('earthRadiusCurvature', 'latitude')
+    seqnum = r.get('sequenceNumber', 'latitude')
+    geodu = r.get('geoidUndulation', 'latitude')
+    heit = r.get('height', 'height', type='float32').astype(np.float32)
+    impp1 = r.get('impactParameterRO_roseq2repl1', 'latitude')
+    impp2 = r.get('impactParameterRO_roseq2repl2', 'latitude')
+    impp3 = r.get('impactParameterRO_roseq2repl3', 'latitude')
+#    impp1 = r.get('impactParameterRO_roseq2repl1', 'latitude', type='double')
```
nicholasesposito: Do we need these comments?

ush/ioda/bufr2ioda/bufr2ioda_gnssro_cosmic2.py
```diff
@@ -0,0 +1,804 @@
...
+        # Create the dimensions
+    if nobs > 0:
+        dims = {'Location': np.arange(0, nobs)}
+        print(' ... dim = ', nobs)
+    else:
+        dims = {'Location': nobs}
+        print(' ... dim = ', nobs)
+
+        # iodafile = f"{cycle_type}.t{hh}z.{data_type}.tm00.{data_format}.nc"
+        # iodafile = f"{cycle_type}.t{hh}z.{ioda_data_type}.{mission}.tm00.{data_format}.nc"
```
nicholasesposito: Unneeded comments I think

nicholasesposito (dismissed):
I think just clean up some of the comments unless they are needed. Otherwise no glaring issues from me. Looks very complicated so thank you for taking the time to do all this!

XuanliLi-NOAA:
@nicholasesposito @emilyhcliu Thank you so much, Nick. I cleaned up the python scripts and also uncommented _mefr1_ to maintain consistency with the other variables.

emilyhcliu (approved):

nicholasesposito (approved):

CoryMartin-NOAA (approved):

nicholasesposito (approved):
Approved.

PR #1488, opened 2025-02-10
JEDI-based ensemble recentering and analysis calculation

DavidNew-NOAA:
# Description

This PR creates a JEDI-based ensemble recentering application. It computes a "correction increment", which when added the the ensemble increments, yields the recentered increments. Some miscellaneous JCB parameters are also created.

# Companion PRs

Global Workflow PR [#3312](https://github.com/NOAA-EMC/global-workflow/pull/3312)
JCB-GDAS PR [#72](https://github.com/NOAA-EMC/jcb-gdas/pull/72)

# Issues

Global Workflow Issue [#3248](https://github.com/NOAA-EMC/global-workflow/issues/3248)

- [ ] atm_jjob 
- [X] C96C48_ufs_hybatmDA 
- [ ] C96C48_hybatmaerosnowDA  
- [ ] C48mx500_3DVarAOWCDA 
- [ ] C48mx500_hybAOWCDA 
- [ ] C96C48_hybatmDA


parm/atm/jcb-base.yaml.j2
```diff
@@ -24,11 +24,13 @@ bound_to_include: begin
 minimizer: DRPCG
 final_diagnostics_departures: anlmob
 final_prints_frequency: PT3H
-number_of_outer_loops: 2
 analysis_variables: [ua,va,t,ps,sphum,ice_wat,liq_wat,o3mr]
+number_of_outer_loops: 2
```
CoryMartin-NOAA: reason to move this line?

utils/fv3jedi/fv3jedi_ecen.h
```diff
@@ -0,0 +1,121 @@
...
+        const eckit::LocalConfiguration ensMeanAnlConfig(additionsConfig[ihrs], \
+                                                         "ensemble mean analysis");
+        const eckit::LocalConfiguration corIncrConfig(additionsConfig[ihrs], \
+                                                      "correction increment");
+
+        // Initialize increment
+        fv3jedi::Increment dxVar(varIncrGeom, incrVars, currentCycle);
+        dxVar.read(varIncrConfig);
+
+        // Initialize backgroun
```
CoryMartin-NOAA:
```suggestion
        // Initialize background
```

utils/fv3jedi/fv3jedi_ecen.cc
```diff
@@ -0,0 +1,8 @@
+#include "fv3jedi_ecen.h"
+#include "oops/runs/Run.h"
+
+int main(int argc,  char ** argv) {
+  oops::Run run(argc, argv);
+  gdasapp::ecen ecen;
```
CoryMartin-NOAA: we should 'kill' these 4 letter names, and just call it recenter everywhere instead of ecen

parm/atm/jcb-base.yaml.j2
DavidNew-NOAA: Mistake. I'll move back when I fix other stuff

utils/fv3jedi/fv3jedi_ecen.h
```diff
@@ -0,0 +1,121 @@
+#pragma once
```
danholdaway: This doesn't seem like quite the right organization of this code. I think `fv3jedi_ecen.h` should contain only the definition of the class. You shouldn't 'implement' the methods in a .h file. Since .h files can be included from several different places it would mean that code being compiled multiple times (and could not be compiled with a static library). For most optimal compilation the class methods would be implemented in the .cc file. If the code in your class is intended to be reusable then you might want to have the class definition in the .h. The class implementation in the .cc and then another .cc in a mains directory (or wherever) that contains the main function. if the code is not really reusable then you could skip the class altogether and more your functions into the .cc file directly and remove the .h. You would lost the inheritance of the oops application though.

CoryMartin-NOAA: I think he's just following the convention we have already in GDASApp which is as you pointed out, wrong...

danholdaway: Ah ok, I see. Lets leave it for now then and we can fix in a batch if it becomes necessary. Perhaps as part of the trudge towards static compilation.

DavidNew-NOAA:
I made a custom `fv3jedi_add_increments` utility in GDASApp to be used in the analysis calculation, because I talked with @danholdaway yesterday, and it's not a good idea physically to interpolate increments rather that states. Now, I add the cubed-sphere backgrounds and increments, interpolate the resulting analysis to the Gaussian grid, and insert the analysis variables into the UFS Gaussian histories to get analyses acceptable to UPP.

The reason I don't use the OOPS `addincrement` utility is that you can only add one state and one increment, and a I need to add several for different forecast hours. My utility allows one to specify a list, much like `convertstate` allows.

utils/fv3jedi/fv3jedi_ecen.cc
DavidNew-NOAA: Done

utils/fv3jedi/fv3jedi_ecen.h
DavidNew-NOAA: Done

parm/atm/jcb-base.yaml.j2
DavidNew-NOAA: done

RussTreadon-NOAA:
@DavidNew-NOAA :  I cloned g-w branch [`DavidNew-NOAA:feature/calcanl`](https://github.com/DavidNew-NOAA/global-workflow/tree/feature/calcanl) on Cactus.  The clone of `sorc/gdas.cd` was incomplete.  `sorc/oops` was not successfully cloned
```
Submodule 'sorc/oops' (https://github.com/jcsda/oops.git) registered for path 'sorc/gdas.cd/sorc/oops'
Cloning into '/lfs/h2/emc/da/noscrub/russ.treadon/git/global-workflow/pr3312/sorc/gdas.cd/sorc/oops'...
fatal: Fetched in submodule path 'sorc/gdas.cd/sorc/oops', but it did not contain 860bfd77d14a4dd697e809650d7d9220671d367d. Direct fetching of that commit failed.
```
GDASApp branch [`feature/calcanl`](https://github.com/NOAA-EMC/GDASApp/tree/feature/calcanl) points at [`oops @ 860bfd7`](https://github.com/jcsda/oops/tree/860bfd77d14a4dd697e809650d7d9220671d367d).  This hash is not found.

DavidNew-NOAA:
@RussTreadon-NOAA That makes sense. I currently have the oops hash pointing to feature/ens_app_full_config. When OOPS PR #[2835](https://github.com/JCSDA-internal/oops/pull/2835) is merged, I can point it to develop.

Don't bother testing this PR for now. I'm in the process right now of revamping it quite a bit. I'm hoping I will be finished by the end of the day.

RussTreadon-NOAA:
@DavidNew-NOAA :  Thank you for the reminder.  I forgot that we pick up most of the JEDI repos from https://github.com/jcsda, not https://github.com/JCSDA-internal/.  I'll stand down on testing g-w PR [#3312](https://github.com/NOAA-EMC/global-workflow/pull/3312).

DavidNew-NOAA:
@CoryMartin-NOAA @RussTreadon-NOAA @danholdaway OK, next up to get https://github.com/NOAA-EMC/global-workflow/pull/3312 ready is this PR. Ready for final review and merger

CoryMartin-NOAA (approved):
Looks good to me

PR #1554, opened 2025-03-18
apply ci path patch only when running g-w ctests

### LINKED ISSUE #1553: Restrict CI g-w path patch to g-w tests
PR #1531 added the following logic to `ci/run_ci.sh`
```
# PATCH START
# HERA role.jedipara can not use /scratch1/NCEPDEV/global. MSU role-da
# can not use /work2/noaa/global. The logic below modifies the paths so
# role.jedipara and role-da can run g-w based ctests.
if [[ "${TARGET}" = "hera" ]]; then
    echo "***WARNING*** apply ${TARGET} global-->da patch to $workflow_dir/workflow/hosts/${TARGET}.yaml"
    sed -i "s|/scratch1/NCEPDEV/global/\${USER}|/scratch1/NCEPDEV/da/\${USER}|g" $workflow_dir/workflow/hosts/${TARGET}.yaml
fi
if [[ "${TARGET}" = "orion" || "${TARGET}" = "hercules" ]]; then
    echo "***WARNING*** apply MSU stmp patch to $workflow_dir/workflow/hosts/${TARGET}.yaml"
    sed -i "s|/work2/noaa/global/\${USER}|/work2/noaa/da/\${USER}|g" $workflow_dir/workflow/hosts/${TARGET}.yaml
fi
# PATCH END
```

This logic is required due to changes in g-w PR [#3430](https://github.com/NOAA-EMC/global-workflow/pull/3430).  The above changes enable `role.jedipara` and `role-da` to run g-w GDASApp ctests.  

We can run GDASApp CI with or without g-w tests included.  GDASApp CI fails when run without g-w tests because `workflow_dir` is not defined.  For example, here is output from Hera CI run without g-w tests:
```
+ echo 'Build: Completed at Tue Mar 18 11:45:26 UTC 2025'
+ [[ 0 == 1 ]]
+ [[ hera = \h\e\r\a ]]
/scratch1/NCEPDEV/da/role.jedipara/CI/GDASApp/repo/ci/run_ci.sh: line 100: workflow_dir: unbound variable
+ ci_status=1
++ date
+ echo 'Finished running run_ci.sh with ci_status 1 at Tue Mar 18 11:45:26 UTC 2025'
Finished running run_ci.sh with ci_status 1 at Tue Mar 18 11:45:26 UTC 2025
```

Variable `TEST_WORKFLOW` is 0 when running GDASApp CI without g-w tests.  `TEST_WORKFLOW=1` when running GDASApp CI with g-w tests.  The path patch is only required when running GDASApp CI with g-w tests.  Thus, the fix is simple.  The patch block needs to be enclosed in 
```
+if [[ $TEST_WORKFLOW == 1 ]]; then

...

 fi
 # PATCH END
```

This issue is opened to document this bug and its resolution.

RussTreadon-NOAA:
# Description

This PR resolves an undefined variable bug in `ci/run_ci.sh`

# Companion PRs

None

# Issues

Resolves #1553

- [ ] atm_jjob 
- [ ] C96C48_ufs_hybatmDA 
- [ ] C96C48_hybatmaerosnowDA  
- [ ] C48mx500_3DVarAOWCDA 
- [ ] C48mx500_hybAOWCDA 
- [ ] C96C48_hybatmDA

### DIFF
--- a/ci/run_ci.sh
+++ b/ci/run_ci.sh
@@ -96,13 +96,15 @@ fi
 # HERA role.jedipara can not use /scratch1/NCEPDEV/global. MSU role-da
 # can not use /work2/noaa/global. The logic below modifies the paths so
 # role.jedipara and role-da can run g-w based ctests.
-if [[ "${TARGET}" = "hera" ]]; then
+if [[ $TEST_WORKFLOW == 1 ]]; then
+  if [[ "${TARGET}" = "hera" ]]; then
     echo "***WARNING*** apply ${TARGET} global-->da patch to $workflow_dir/workflow/hosts/${TARGET}.yaml"
     sed -i "s|/scratch1/NCEPDEV/global/\${USER}|/scratch1/NCEPDEV/da/\${USER}|g" $workflow_dir/workflow/hosts/${TARGET}.yaml
-fi
-if [[ "${TARGET}" = "orion" || "${TARGET}" = "hercules" ]]; then
+  fi
+  if [[ "${TARGET}" = "orion" || "${TARGET}" = "hercules" ]]; then
     echo "***WARNING*** apply MSU stmp patch to $workflow_dir/workflow/hosts/${TARGET}.yaml"
     sed -i "s|/work2/noaa/global/\${USER}|/work2/noaa/da/\${USER}|g" $workflow_dir/workflow/hosts/${TARGET}.yaml
+  fi
 fi
 # PATCH END
 


RussTreadon-NOAA:
The modified `ci/run_ci.sh` was manually installed in the role.jedipara (Hera) and role-da (MSU) accounts.  GDASApp RT was triggered via labels.  GDASApp tests passed on Hera, Hercules, and Orion.

This PR is ready for review.

ci/run_ci.sh
```diff
@@ -96,13 +96,15 @@ fi
 # HERA role.jedipara can not use /scratch1/NCEPDEV/global. MSU role-da
 # can not use /work2/noaa/global. The logic below modifies the paths so
 # role.jedipara and role-da can run g-w based ctests.
-if [[ "${TARGET}" = "hera" ]]; then
+if [[ $TEST_WORKFLOW == 1 ]]; then
```
CoryMartin-NOAA, line 99: seems strange to set this equal to 1, but I guess that is outside of the control of GDASApp/this PR?

CoryMartin-NOAA (approved):

ci/run_ci.sh
RussTreadon-NOAA: `TEST_WORKFLOW` are local to `ci/run_ci.sh` and `ci/driver.sh`.  Both scripts were already using `TEST_WORKFLOW=1` trigger the running of g-w based GDASApp ctests.  I simply replicated existing logic here.  That said, one could replace the `0/1` toggle with `NO/YES`.

CoryMartin-NOAA: No, this is fine, for all I know I was the one who did that in the first place (haha)

DavidNew-NOAA (approved):

RussTreadon-NOAA:
Thank you, @CoryMartin-NOAA and @DavidNew-NOAA .  Merging now.

PR #1511, opened 2025-02-25
Marine in situ obs update (SSS, moorings, drifters)

JohnSteffen-NOAA:
# Description

This PR resurrects the SSS SMAP/SMOS retrievals in monitoring mode and adds tropical moorings and surface drifters to the list of assimilated in situ observations.

# Issues

Resolves #1461 
Refs #1301 
-->

# Automated CI tests to run in Global Workflow

- [ ] C48mx500_3DVarAOWCDA

PR #1560, opened 2025-03-19
Bugfix SSS obs for LETKF

### LINKED ISSUE #1559: Salinity obs configuration is missing obs localization section
which leads to LETKF failing in some experiments.

Fix by adding localization section similar to all other obs.

shlyaeva:
# Description

Bugfixes SSS obs for LETKF by adding "obs localizations" yaml section similar to other yamls, otherwise LETKF using these yamls fails.

# Issues

Resolves #1559 

- [ ] atm_jjob 
- [ ] C96C48_ufs_hybatmDA 
- [ ] C96C48_hybatmaerosnowDA  
- [ ] C48mx500_3DVarAOWCDA 
- [x] C48mx500_hybAOWCDA 
- [ ] C96C48_hybatmDA

### DIFF
--- a/parm/soca/obs/config/sss_smap_l2.yaml
+++ b/parm/soca/obs/config/sss_smap_l2.yaml
@@ -42,3 +42,10 @@ obs filters:
   where:
   - variable: {name: GeoVaLs/distance_from_coast}
     minvalue: 100e3
+obs localizations:
+- localization method: Rossby
+  base value: 100.0e3
+  rossby mult: 1.0
+  min grid mult: 2.0
+  min value: 200.0e3
+  max value: 900.0e3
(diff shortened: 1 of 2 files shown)


JohnSteffen-NOAA:
If I recall correctly, @guillaumevernieres requested I remove that section from my original SSS PR #1511 . Also, if we want to add in the localization section, we need to make the same change in jcb-gdas/observations/marine j2 yamls as well.

shlyaeva:
Where are jcb-gdas/observations/marine j2 yamls used?
@guillaumevernieres why not have obs localizations for SSS?
If we don't want to assimilate SSS in LETKF, what do I need to change so that the list of obs is different between LETKF and Var?

guillaumevernieres (approved):
Thanks @shlyaeva !

PR #1574, opened 2025-03-26
Feature/oops profiling cmpexps

rtodling:
# Description

This adds capabilities to the profiling program to allow for comparison of timings between two (or more experiments). The program was reorganized a little, putting the pie-generation original code into a function and adding another function to handle the new option. The only difference from the original implementation is that to generate a pie plot the your has to specify the "plot type" in the yaml.

A yaml illustrating the use of a comparison is added to the repo.

Here is a figure showing what a comp plot looks like.

timings

# Companion PRs

None

# Issues

No issues associated with this.

Refs NOAA-EMC/repo#5678
-->

- [x] atm_jjob 
- [ ] C96C48_ufs_hybatmDA 
- [ ] C96C48_hybatmaerosnowDA  
- [ ] C48mx500_3DVarAOWCDA 
- [ ] C48mx500_hybAOWCDA 
- [ ] C96C48_hybatmDA


danholdaway:
Thanks for this addition to the profiling tool @rtodling. Can you change the base branch from develop to feature/oops_profiling?

PR #1597, opened 2025-04-08
Adds some paths for aircraft profiles

### LINKED ISSUE #1596: Add new path names for acftprofiles
Some lines need to be added to run_ufo_hofx_test.sh in order to reflect the update of aircraft profiles prepbufr phase one yaml to jcb-gdas. The additions will be for the aircraft bias file.

nicholasesposito:
# Description

Add some lines to the run_ufo_hofx_test script to take care of the addition of using the aircraft bias file.

# Companion PRs
TBD

# Issues
Fixes #1596 

None


CoryMartin-NOAA (commented):
its already long enough, why not `aircraft` instead of `acft`?

nicholasesposito:
> its already long enough, why not `aircraft` instead of `acft`?

acft_out is what the files were called originally. aircraft_out makes sense though especially when it comes to consistency within JEDI/GDASApp. At the same time, "acft" comes from the prepbufr files being named gdas.t00z.prepbufr.acft_profiles. 

I did a recursive grep into all files in GDASApp for occurrences of that file, and only found that the acft_converter.yaml in iodaconv uses this (as well as the associated CMake file). 

All that being said, changing it to aircraft probably wouldn't be the hardest thing in the world. What do we think?

CoryMartin-NOAA (approved):
I'm happy either way, thanks

RussTreadon-NOAA:
@nicholasesposito :  What's the status of this PR? 
- ready to review & merge?
- work continues so keep in draft mode?
- not needed so close without merge?
- something else?

nicholasesposito:
I have newer changes for this. Let's close it and I'll make a newer one.

PR #1611, opened 2025-04-10
Add marine ensemble variance output to recentering task

### LINKED ISSUE #1540: add ensemble variance computation/save to the recenter task
In prep for removing/not running letkf tasks for v17, add a way to compute the ensemble variance (in the recenter task).

shlyaeva:
# Description

Adds marine ensemble variance output to the recentering task, so we can monitor ensemble spread even in the ensemble experiments without LETKF.

The background ensemble variance for the ?? cycle is saved in `enkfgdas.*/??/ensstat/analysis/ocean/` and `enkfgdas.*/??/ensstat/analysis/ice/` directories.

# Issues

Resolves #1540 

- [ ] atm_jjob 
- [ ] C96C48_ufs_hybatmDA 
- [ ] C96C48_hybatmaerosnowDA  
- [x] C48mx500_3DVarAOWCDA 
- [x] C48mx500_hybAOWCDA 
- [ ] C96C48_hybatmDA

### DIFF
--- a/ush/soca/marine_recenter.py
+++ b/ush/soca/marine_recenter.py
@@ -263,7 +263,28 @@ class MarineRecenter(Task):
             mem_dir_list.append(mem_dir_real)
             copy_list.append([f'ocn.recenter.incr.{str(mem)}.nc',
                               os.path.join(mem_dir_real, incr_file)])
-
+        # Copy the ensemble variance
+        ensvar_date = self.task_config.MARINE_WINDOW_END.strftime('%Y-%m-%dT%H:%M:%SZ')
+        stats_dir = os.path.join(self.task_config.ROTDIR,
+                                 f'enkf{RUN}.{PDYstr}',
+                                 f'{cyc}',
+                                 'ensstat',
+                                 'analysis',
+                                 'ocean')
+        stats_dir_real = os.path.realpath(stats_dir)
+        mem_dir_list.append(stats_dir_real)
+        copy_list.append([f'ocn.ensvar.incr.{ensvar_date}.nc',
+                         os.path.join(stats_dir_real, f'enkf{RUN}.t{cyc}z.ocn.bg_ensvar.nc')])
+        stats_dir = os.path.join(self.task_config.ROTDIR,
+                                 f'enkf{RUN}.{PDYstr}',
+                                 f'{cyc}',
+                                 'ensstat',
+                                 'analysis',
+                                 'ice')
+        stats_dir_real = os.path.realpath(stats_dir)
+        mem_dir_list.append(stats_dir_real)
+        copy_list.append([f'ice.ensvar.incr.{ensvar_date}.nc',
+                         os.path.join(stats_dir_real, f'enkf{RUN}.t{cyc}z.ice.bg_ensvar.nc')])
         FileHandler({'mkdir': mem_dir_list}).sync()
         FileHandler({'copy': copy_list}).sync()
 
--- a/parm/soca/berror/soca_ensrecenter.yaml
+++ b/parm/soca/berror/soca_ensrecenter.yaml
@@ -117,3 +117,9 @@ output increment:
   type: incr
   output file: 'recenter.incr.%mem%.nc'
   pattern: '%mem%'
+
+ensemble variance output:
+  datadir: ./
+  date: '{{ MARINE_WINDOW_BEGIN_ISO }}'
+  exp: ensvar
+  type: incr
--- a/parm/soca/berror/soca_ensrecenter.yaml
+++ b/parm/soca/berror/soca_ensrecenter.yaml
@@ -120,6 +120,6 @@ output increment:
 
 ensemble variance output:
   datadir: ./
-  date: '{{ MARINE_WINDOW_BEGIN_ISO }}'
+  date: '{{ MARINE_WINDOW_END_ISO }}'
   exp: ensvar
   type: incr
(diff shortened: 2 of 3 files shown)


guillaumevernieres (approved):

ush/soca/marine_recenter.py
```diff
@@ -263,7 +263,28 @@ def finalize(self):
...
                               os.path.join(mem_dir_real, incr_file)])
-
+        # Copy the ensemble variance
+        ensvar_date = self.task_config.MARINE_WINDOW_END.strftime('%Y-%m-%dT%H:%M:%SZ')
+        stats_dir = os.path.join(self.task_config.ROTDIR,
+                                 f'enkf{RUN}.{PDYstr}',
+                                 f'{cyc}',
+                                 'ensstat',
+                                 'analysis',
+                                 'ocean')
```
AndrewEichmann-NOAA, line 273: Is there a template to make this path from? If not, we'll need to add it

ush/soca/marine_recenter.py
```diff
@@ -263,7 +263,28 @@ def finalize(self):
...
+        stats_dir_real = os.path.realpath(stats_dir)
+        mem_dir_list.append(stats_dir_real)
+        copy_list.append([f'ocn.ensvar.incr.{ensvar_date}.nc',
+                         os.path.join(stats_dir_real, f'enkf{RUN}.t{cyc}z.ocn.bg_ensvar.nc')])
+        stats_dir = os.path.join(self.task_config.ROTDIR,
+                                 f'enkf{RUN}.{PDYstr}',
+                                 f'{cyc}',
+                                 'ensstat',
+                                 'analysis',
+                                 'ice')
```
AndrewEichmann-NOAA, line 283: Same comment about the template

ush/soca/marine_recenter.py
shlyaeva: I don't know what templates for the path are, can you tell me some more?

AndrewEichmann-NOAA: They are used to create the paths under ROTDIR, and start under https://github.com/NOAA-EMC/global-workflow/blob/940e3539c174b05564f0652aad4bb41af0e8f697/parm/config/gfs/config.com#L84
The actual paths are declared in the job script and then accessed from the task_config in the script. We don't have to hold things up for that (it requires a PR to g-w) but it's something to address in an issue. I can do it.

AndrewEichmann-NOAA: Unless we want to use` analysis/ocean`?

shlyaeva: Oh, cool, I didn't know this, thank you for pointing to it!
It _may be_ COM_OCEAN_ANALYSIS_TMPL when MEMDIR is "ensstat"? I'm not sure.
Thank you for offering to do it! 🙏

AndrewEichmann-NOAA: That is, `COMOUT_OCEAN_ANALYSIS` and `COMOUT_ICE_ANALYSIS`, which are already available?

guillaumevernieres: I think @AndrewEichmann-NOAA might be talking about making use of this:
https://github.com/NOAA-EMC/global-workflow/blob/940e3539c174b05564f0652aad4bb41af0e8f697/jobs/JGLOBAL_MARINE_ANALYSIS_LETKF#L31

Is that correct @AndrewEichmann-NOAA ?

AndrewEichmann-NOAA: So if it needs its own `ensstat` subdir, then that would require a PR to g-w, otherwise you can use `COMOUT_*` as above. Let me know which you want.

AndrewEichmann-NOAA: @guillaumevernieres More like https://github.com/NOAA-EMC/global-workflow/blob/940e3539c174b05564f0652aad4bb41af0e8f697/jobs/JGDAS_GLOBAL_OCEAN_ANALYSIS_ECEN#L29

shlyaeva: I saw that the atmospheric ensemble stats go to a similar directory as in this PR:
`enkfgdas*/[hour]/ensstat/[ocean|atm|ice]/analysis`
while the ensemble members go to
`enkfgdas*/[hour]/mem*/[ocean|atm|ice]/analysis`

Note that this is different from `gdas*/[hour]/[ocean|ice]/analysis/` which I think correspond to `COMOUT_OCEAN_ANALYSIS` and `COMOUT_ICE_ANALYSIS`?

AndrewEichmann-NOAA: Ah, right. I'd have to dig into how enkfgdas is handled.

AndrewEichmann-NOAA: I'm happy to approve this and handle the templating later.

AndrewEichmann-NOAA (approved):

ush/soca/marine_recenter.py
guillaumevernieres: > Ah, right. I'd have to dig into how enkfgdas is handled.

That's why I pointed to the enkf job earlier. I'll create an issue out of your earlier comment @AndrewEichmann-NOAA . ~This isn't in the v17 critical path so tech debt is ok imo.~ Friday fog. We don't make use of the spread, but monitoring it is in the critical path.

guillaumevernieres: Just wrote the above but see that @JessicaMeixner-NOAA flagged this as a v17 issue ... Oops. I'll head scratch a bit more before merging this work.

JessicaMeixner-NOAA: I flagged this because @shlyaeva flagged this to me as something that would be helpful for v17.

guillaumevernieres: You flagged it correctly @JessicaMeixner-NOAA .

shlyaeva: yep, would be good to still have spread diagnostics like the ones we looked at this week when we don't run letkf (in retro and parallel runs and in operations eventually).

PR #1618, opened 2025-04-14
make GDASapp compile on NOAA CSPs with Rocky 8

### LINKED ISSUE #1617: make GDASapp compile on NOAA CSPs with Rocky 8
After OS on NOAA cloud updated to Rocky 8, need to update GDASapp for noaacloud to compile on CSPs.

weihuang-jedi:
Update noaacloud modulefiles so GDASapp can compile on CSPs with Rocky 8 (originally CentOS).

Resolves #1617

- [ ] atm_jjob 
- [ ] C96C48_ufs_hybatmDA 
- [ ] C96C48_hybatmaerosnowDA  
- [ ] C48mx500_3DVarAOWCDA 
- [ ] C48mx500_hybAOWCDA 
- [ ] C96C48_hybatmDA


CoryMartin-NOAA (commented):
Will the gdas_common approach work on all platforms? I don't like the idea of doing it for only the cloud and calling it 'common'

weihuang-jedi:
The whole log file is too long. The last few line of the compile output shows:

make[2]: Nothing to be done for 'da-utils/src/ioda-extract/CMakeFiles/meanioda.x.dir/build'.
make[2]: Leaving directory '/contrib/Wei.Huang/dev/gdasapp-cloud/build'
make[2]: Leaving directory '/contrib/Wei.Huang/dev/gdasapp-cloud/build'
[100%] Built target meanioda.x
[100%] Built target ioda-stats.x
make[1]: Leaving directory '/contrib/Wei.Huang/dev/gdasapp-cloud/build'
/contrib/spack-stack-rocky8/spack-stack-1.6.0/envs/ue-intel/install/intel/2021.10.0/cmake-3.23.1-2vxp2hq/bin/cmake -E cmake_progress_start /contrib/Wei.Huang/dev/gdasapp-cloud/build/CMakeFiles 0
+ cd ../
+ set +x
Finish ... Mon Apr 14 16:28:45 UTC 2025

This is how I see that the compile is successful.

weihuang-jedi:
@CoryMartin-NOAA 
Call it "common" is probably not right at this stage, as I have not check on other platforms (but only on CSPs).
But if we (say you) like this approach, then it can be test out, and eventually become "common".

If you do not like the idea of this approach, I can use the same thing as other platforms.

Please let me know  your preference.

CoryMartin-NOAA:
I like the approach if it works for all platforms, because then you only need to change one file when the version of a library changes.

weihuang-jedi:
Since most platforms use spack-stack 1.6.0, so the libraries is likely the same.
By the way, this is copied from ufs-weather-model, where they use this approach.

weihuang-jedi:
@CoryMartin-NOAA 
Sorry we did not get this PR merged, or closed, but left it dandling here.
Recently Maoyi finds this and asked what the situation, and what to do this.
Since there are very few users compile/run GDAS on CSPs
(actually only have Taylor Roper form QOSAP/NOAA is working on it on AWS),
we did not come back to check.

Now, since we are moving to spack-stack 1.9.2 on many platfomrs, including CSP,
can we keep this PR open for a little bit longer, and once SS 1.9.2 PR merged,
then this PR become obsolete, and we close it then with msg like:
Obsolete with SS 1.9.2 upgrade.

Do you agree with this?

Thanks,

Wei

CoryMartin-NOAA:
I'm fine with that, besides WCOSS2, all NOAA on prem HPC that we support for GDASApp is already using 1.9.2

TaylorRoper-NOAA:
I'll add that #1797 ports the cloud to 1.9.2.

CoryMartin-NOAA:
Now that #1797 is merged, does that mean this can be closed?

RussTreadon-NOAA:
@weihuang-jedi :  What's the status of this PR?   May this PR be closed?

PR #1151, opened 2024-06-06
Changes name of rossrad file for letkf

AndrewEichmann-NOAA:
Changes soca fix file staging yaml for rossrad.dat -> rossrad.nc

Another little step towards https://github.com/NOAA-EMC/GDASApp/issues/1091

PR #1659, opened 2025-05-05
Use UFS MOM6 and Icepack

DavidNew-NOAA:
# Description

This PR modifies `build.sh` so that if `$WORKFLOW_BUILD == 'YES'`, symbolic links in SOCA are created to the MOM6 and Icepack submodules in in UFS. This is part of a broader effort of forking SOCA for the purpose of EE2 compliance.

I've tested `C48mx500_3DVarAOWCDA` successfully on Hera with the workflow build.

# Companion PRs

SOCA PR [#1151](https://github.com/JCSDA-internal/soca/pull/1151)

- [ ] atm_jjob 
- [ ] C96C48_ufs_hybatmDA 
- [ ] C96C48_hybatmaerosnowDA  
- [x] C48mx500_3DVarAOWCDA 
- [ ] C48mx500_hybAOWCDA 
- [ ] C96C48_hybatmDA


RussTreadon-NOAA (commented):
It looks like the soca submodule change is on a branch, not soca `develop`.  Is this correct?  

Just curious, what's the plan moving forward?  Will the soca changes in this PR be merged into JCSDA's soca repo or will we fork and merge these changes into our soca fork?

Changes look straight foward.  We can trigger automated CI to ensure everything builds and runs as intended.

build.sh
```diff
@@ -108,6 +108,14 @@ mkdir -p ${BUILD_DIR} && cd ${BUILD_DIR}
 WORKFLOW_BUILD=${WORKFLOW_BUILD:-"OFF"}
 CMAKE_OPTS+=" -DWORKFLOW_TESTS=${WORKFLOW_TESTS:-${WORKFLOW_BUILD}}"
 
+# If cloned with the global-workflow, link MOM6 and Icepack in SOCA to submodules in the UFS repo
+if [[ $WORKFLOW_BUILD == 'YES' ]]; then
+  rm -rf $dir_root/sorc/soca/external/MOM6/mom6
+  rm -rf $dir_root/sorc/soca/external/icepack/Icepack
+  ln -sf $dir_root/../ufs_model.fd/MOM6-interface/MOM6/ $dir_root/sorc/soca/external/mom6/MOM6
```
CoryMartin-NOAA, line 115:
```suggestion
  ln -sf $dir_root/../ufs_model.fd/MOM6-interface/MOM6/ $dir_root/sorc/soca/external/MOM6/mom6
```

CoryMartin-NOAA: This is the opposite of the `rm` statement. I don't know which is correct but they should be consistent, no?

DavidNew-NOAA:
@RussTreadon-NOAA that's correct. The SOCA branch to be merged in PR [#1151]((https://github.com/JCSDA-internal/soca/pull/1151) has CMake changes required to link to the UFS MOM6 and Icepack submodules. And yes, my understanding is that we fork after these changes. The changes to the fork will be two-fold (again to my understanding):
1. Delete the MOM6 and Icepack submodules so they're not cloned twice
2. Delete the LFS test data

build.sh
DavidNew-NOAA: @CoryMartin-NOAA The `ln` was right and the `rm` was wrong rather, which why testing didn't catch this. Good catch

DavidNew-NOAA:
The SOCA PR companion is now merged, so triggering CI now for this PR

CoryMartin-NOAA (dismissed):

danholdaway (commented):
Seems reasonable but perhaps this PR should just wait until the soca fork is ready and then do everything in one PR? There's a removal of things from SOCA but this shouldn't be necessary once we have the fork, which wouldn't have these directories.

guillaumevernieres (dismissed):

guillaumevernieres (dismissed):

build.sh
```diff
@@ -108,6 +108,14 @@ mkdir -p ${BUILD_DIR} && cd ${BUILD_DIR}
 WORKFLOW_BUILD=${WORKFLOW_BUILD:-"OFF"}
 CMAKE_OPTS+=" -DWORKFLOW_TESTS=${WORKFLOW_TESTS:-${WORKFLOW_BUILD}}"
 
+# Link MOM6 and Icepack in SOCA to submodules in the UFS repo
+if [[ $WORKFLOW_BUILD == 'ON' ]]; then
```
danholdaway, line 112: Perhaps I'm missing something but I'm not sure I understand this logic. Now that the submodule points to the dev/emc branch of the NOAA-EMC fork there is no `$dir_root/sorc/soca/external/mom6/MOM6` to remove.  Wouldn't the logic need to be inverted with the work to do when `$WORKFLOW_BUILD == 'OFF'`?  Then the work is to add back by removing the version of SOCA that was cloned at checkout out and instead clone from JCSDA-internal?

The problem though is that `build.sh` could in theory run on a compute node that does not have access to the internet. So the non-workflow build might need to be a bit more manual than that.

DavidNew-NOAA: @danholdaway You're right, but this branch stopped building successfully, so with this last commit, I'm testing the possibility that when the CI driver checks out GW, it's checking out a hash of that still has the mom6 and icepack submodules in its soca submodules, and somehow that's interfering with the linking.

DavidNew-NOAA: @danholdaway To clarify my last comment, G-W develop currently points to a GDASApp hash which points to a soca hash which still contains mom6 and icepack submodules. So the CI driver checks out G-W develop and _then_ it checks out the new GDASApp hash which does not have mom6 and icepack submodules. This is somehow interfering with the linking for a reason I don't understand. And it appears that adding the `rm -rf $dir_root/sorc/soca/external/mom6/MOM6` indeed fixed the error, because I'm looking at the build log from my last triggering of CI. 

Ultimately the `-f` means `rm` won't fail if `$dir_root/sorc/soca/external/mom6/MOM6` doesn't exist, so while redundant, it allows CI to pass.

DavidNew-NOAA:
@CoryMartin-NOAA See /scratch1/NCEPDEV/da/role.jedipara/CI/GDASApp/workflow/PR/1659/global-workflow/sorc/gdas.cd/build/log.ctest for failure of `test_gdasapp_util_ioda_example`. This branch includes your fixes to da-utils

@RussTreadon-NOAA `C48mx500_3DVarAOWCDA` didn't run, so I think we need to update the `grep` command on Hera like you did on MSU

build.sh
danholdaway: Thanks @DavidNew-NOAA. It doesn't hurt to have the rm so that's fine. But we probably still need some kind of else to this if though. Some logic that replaces SOCA with the JCSDA-internal version or clones MOM6 into the place where SOCA expects it to be. Otherwise I think the WORKFLOW_BUILD = OFF mode would be broken.

CoryMartin-NOAA:
`Reason: Input file is not readable, will stop execution. File: ../../../sorc/soca/test/Data/obs/gmi_gpm_obs.nc`
sounds like a missing test file, probably because soca has been pruned?

build.sh
DavidNew-NOAA: @danholdaway I was discussing this with @CoryMartin-NOAA and we think that `WORKFLOW_BUILD = OFF` would be more of a "power user" option, whereby SOCA is manually cloned by the user before building. The reason is this: The build script would need to know which SOCA hash to clone, since SOCA develop won't alway work with G-W/GDASApp develop.

DavidNew-NOAA:
@CoryMartin-NOAA Righttt that's the reason. Maybe we can put the data for that test in GDASApp itself?

CoryMartin-NOAA:
@DavidNew-NOAA then we are back at square 1, we can't have git-lfs files here, but we can add them to a tarball and download it (or stage them on supported platforms)

DavidNew-NOAA:
I disabled `test_gdasapp_util_ioda_example` for now because of its reference to soca test data. `C48mx500_3DVarAOWCDA` passed, so I'd say this PR is ready to merge after a final review

DavidNew-NOAA:
@CoryMartin-NOAA @danholdaway @guillaumevernieres This PR is ready for final review

CoryMartin-NOAA (approved):

guillaumevernieres (approved):
Thanks @DavidNew-NOAA

build.sh
shlyaeva: Can we avoid having a fork by just doing the `rm -rf` here in all conditions?

CoryMartin-NOAA: @shlyaeva that still requires git-lfs to clone all the data first, which leads to a waste of bandwidth that has to be paid for

PR #1683, opened 2025-05-11
fixes bug

givelberg:
# Description
fixes the bug where ocean basin array may be of incorrect size. 

# Companion PRs

# Issues

- [ ] atm_jjob 
- [ ] C96C48_ufs_hybatmDA 
- [ ] C96C48_hybatmaerosnowDA  
- [x] C48mx500_3DVarAOWCDA 
- [ ] C48mx500_hybAOWCDA 
- [ ] C96C48_hybatmDA

### DIFF
--- a/ush/ioda/bufr2ioda/marine/b2i/b2iconverter/ocean.py
+++ b/ush/ioda/bufr2ioda/marine/b2i/b2iconverter/ocean.py
@@ -61,10 +61,10 @@ class OceanBasin:
         dlon = self.__longitudes[1] - self.__longitudes[0]
 
         # the data may be a masked array
-        ocean_basin = []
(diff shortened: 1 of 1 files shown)


guillaumevernieres:
@givelberg , it looks short, I could probably reverse engineer what it is you are doing ... But why?
Where's the description? issue?

givelberg:
@guillaumevernieres Sorry, I thought I put in a description;
it is a small bug that could under some circumstances make the ocean basin array in the code to be of the wrong size; in that case the code would fail. I don't think you ever ran into this bug because it would break things, but since you are running the code and the data is suspect, it would be good to merge it in. There was no issue. Should I create an issue in a case like this?

PR #1759, opened 2025-06-18
Update gaeac6.intel.lua following maintenance

CoryMartin-NOAA:
# Description
Hat tip to @aerorahul for finding this is the needed solution to enable building to resume on GaeaC6


RussTreadon-NOAA (approved):
Approve.  @DavidHuber-NOAA found this change works.

Issue #1763, opened 2025-06-24
FV3 surface files have the wrong sea ice when cycling the coupled model


guillaumevernieres:
## Description
We need to replace the aggregated sea ice in the `FV3` surface restarts with the sea ice analysis produced by `SOCA`.

## Proposed Solution
Develop an application that uses `soca`, `fv3-jedi`, `oops`' unstructured interpolation, and `atlas` to interpolate the `SOCA` sea ice concentration field onto the `FV3` surface restart grid.

## Tasks
- [x] Step 1: Add a `CTest` that sets up all required resource files to run `fv3-jedi` and `soca`. Use a placeholder (e.g., a minimal dummy application that builds both geometries) to validate the test setup. — @guillaumevernieres

- [ ] Step 2: Implement the interpolation of sea ice concentration from the `SOCA` geometry to the `FV3` surface geometry. — @shlyaeva

shlyaeva:
Interpolation app: https://github.com/JCSDA-internal/oops/pull/2921. I'm going to test it with interpolating ice for high-res files to make sure it works OK with land masks that may be misaligned between fv3 and soca.


Which variables in fv3 files need to be updated with soca analysis? @guillaumevernieres and @rgrumbine brought up `fice` (total ice concentration). Anything else? Anyone else to check in with?

CatherineThomas-NOAA:
> Which variables in fv3 files need to be updated with soca analysis? @guillaumevernieres and @rgrumbine brought up fice (total ice concentration). Anything else? Anyone else to check in with?

Any input @NeilBarton-NOAA or @GeorgeGayno-NOAA ?

NeilBarton-NOAA:
I do not know the FV3 sea ice variables. @DeniseWorthen likely is a good resource as I assume you'll want to update the same variables as what is being coupled.

DeniseWorthen:
I know the coupling variables, but you're asking for "aggregated" variables from CICE which are written to the ATM sfc restarts? I would guess that would be ``fice`` and ``hice`` and ``tiice``. Looking at a surface restart file, it appears there may be others, but I'm not sure of the exact provenance in the restart. For example, does "sncovr_ice" come from CICE?

shlyaeva:
From some of @guillaumevernieres's digging: looks like we also need to update `slmsk` because it includes an ice mask?

And to second @DeniseWorthen's question: do we need to update `sncovr_ice`?

guillaumevernieres:
[...]

PR #1765, opened 2025-06-25
Add a simple soca fv3 "coupled" ctest

guillaumevernieres:
# Description
See issue #1763


guillaumevernieres:
FYI @shlyaeva , I'll add the other stuff that `fv3-jedi` needs to allow us to read a surface restart. That's indeed a lot simpler than I expected!

guillaumevernieres:
Questions/Notes to reviewers:
- I added the orography files, but they are not used in this PR, are we ever going to need them?
- We're planning to start interpolating ice concentration from soca to fv3jedi, but the example here is for temperature, I suspect the masking will be more problematic in the seaice case.
- ...

utils/soca/CMakeLists.txt
```diff
@@ -27,5 +27,8 @@ target_link_libraries( gdas_soca_obsstats.x PUBLIC NetCDF::NetCDF_CXX oops ioda)
 # Diag B
 add_subdirectory( diagb )
 
+# Coupled stuff
+add_subdirectory( coupled )
+
```
CoryMartin-NOAA, line 32: Missing EOL at the end here

utils/soca/coupled/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+ecbuild_add_executable( TARGET gdas_soca_to_fv3.x
+                        SOURCES gdas_soca_to_fv3.cc )
+target_compile_features( gdas_soca_to_fv3.x PUBLIC cxx_std_17)
+target_link_libraries( gdas_soca_to_fv3.x PUBLIC NetCDF::NetCDF_CXX oops fv3jedi soca)
```
CoryMartin-NOAA: No EOL

utils/soca/coupled/gdas_soca_to_fv3.h
```diff
@@ -0,0 +1,79 @@
...
+#include "oops/mpi/mpi.h"
+#include "oops/runs/Application.h"
+#include "oops/util/DateTime.h"
+#include "oops/util/Logger.h"
+
+#include "soca/Geometry/Geometry.h"
+#include "soca/State/State.h"
+
+namespace gdasapp {
+namespace coupled {
```
CoryMartin-NOAA, line 22: should this be indented for readability?

danholdaway:
@shlyaeva could this ultimately be done generically in OOPS? I.e. a coupled ConvertState application that can go from any model to any model?

shlyaeva:
@danholdaway yes, good point, I'm pretty sure it can!

utils/soca/coupled/gdas_soca_to_fv3.h
shlyaeva: I think across jedi code in general it's not.

shlyaeva (dismissed):
Thank you for adding this @guillaumevernieres. I didn't review the fv3 resource files, and didn't review the application code in detail since I'm working on changing it. Everything else looks good to me. We may be able, as you and @danholdaway suggested before, to have a generic oops application that can do state interpolation from MODEL1 to MODEL2, in which case `gdas_soca_to_fv3.h` will disappear, but this can be done in the future PR when the interpolation functionality is actually added.

shlyaeva (approved):

shlyaeva:
oops app here: https://github.com/JCSDA-internal/oops/compare/develop...feature/models_interp and coupling test with artificially chosen variables here: https://github.com/JCSDA-internal/coupling/compare/develop...feature/models_interp
I've tested with soca->fv3 on those coupling low res tests. It seems reasonable in general, but the masking looks  somewhat inconsistent. I'm not confident in that low res test data, I'll do more tests before opening oops PR.

PR #1774, opened 2025-06-27
Diagnostic application

guillaumevernieres:
# Description
This will be spruced up at some point, all it does for now is compute geostrophy, in a buggy way. The barotropic fields might be OK but the 3D fields can't be derived with the hybrid vertical coordinate used. A PR for an other day.


utils/soca/test/testinput/gdas_soca_diagnostics.yaml
```diff
@@ -0,0 +1,21 @@
...
+  #ice_filename: cice.res.nc
+  date: *date
+  state variables:
+  - sea_surface_height_above_geoid
+
+output:
+  datadir: ./
+  date: *date
+  exp: barotropic
+  type: incr
```
CoryMartin-NOAA: missing EOL

CoryMartin-NOAA (approved):

PR #1799, opened 2025-07-15
Support chem B matrix rescaling through YAML

CoryMartin-NOAA:
# Description

This PR adds support to the chem diagb code the ability to rescale tracer standard deviations from YAML rather than reading in a netCDF file.

# Companion PRs

N/A but note we will need to do subsequent PRs that are coordinated to global-workflow, GDASApp, jcb-gdas to use this

# Issues

Did not make one (oops!)

- [ ] atm_jjob 
- [ ] C96C48_ufs_hybatmDA 
- [ ] C96C48_hybatmsnowDA 
- [ ] C96_gcafs_cycled 
- [ ] C48mx500_3DVarAOWCDA 
- [ ] C48mx500_hybAOWCDA 
- [ ] C96C48_hybatmDA


andytangborn (changes requested):
I think that the default rescaling values should be set to 1 for all species. I know that if you leave out the rescaling then they will all automatically be 1, but is that the best solution. In case the rescaling needs to be changed in the future, would it be clear where to do it if they are not explicitly shown.

CoryMartin-NOAA:
@ypwang19 @andytangborn I've updated jcb-gdas to not rescale, and the workflow `develop` branch can stage the orography files now. I believe this is now ready for final review.

andytangborn (approved):
This looks fine. I take it that that the default for high latitude is not sat at +- 90, and the rescaling is 1.0 , so by default it has no effect. This is likely fine, unless we find out otherwise.

CoryMartin-NOAA:
Thanks @andytangborn we can always revisit the default values later in a subsequent JCB-GDAS PR.

PR #1728, opened 2025-06-03
GDASApp companion PR for direct installation of GDASApp executables in Global Workflow 

DavidNew-NOAA:
# Description

The PR, along with companion https://github.com/NOAA-EMC/global-workflow/pull/3760, seeks to achieve direction installation of GDASApp executable in the Global Workflow. Mostly, this is achieved in `build.sh` by running `make install` for `WORKFLOW_BUILD=ON`. Also, all GDASApp utilities are renamed to include the prefix `gdas_` + (if they're OOPS applications) the name of the JEDI model interface they use: `fv3jedi_`, `soca_`, `ioda_`. Then, they're all moved to `$HOMEgfs/exec` from `$HOMEgfs/bin` in `build.sh`, and `$HOMEgfs/bin` is deleted.

# Companion PRs

https://github.com/NOAA-EMC/global-workflow/pull/3760

# Issues

https://github.com/NOAA-EMC/GDASApp/issues/1678

# Automated CI tests to run in Global Workflow

- [ ] atm_jjob 
- [ ] C96C48_ufs_hybatmDA 
- [ ] C96C48_hybatmaerosnowDA  
- [ ] C48mx500_3DVarAOWCDA 
- [ ] C48mx500_hybAOWCDA 
- [ ] C96C48_hybatmDA

Issue #161, opened 2022-10-19
Create backround seaice file readable by soca


guillaumevernieres:
## Description
In the past we've relied in reading the CICE restarts to generate the aggregated background. This is fine, but not optimum since it requires dumping intermittent restarts which can be slow at 1/4 deg resolution.  

## Solution
Instead of doing the above, manipulate the already aggregated CICE diag files to make them readable by soca/fms.

guillaumevernieres:
CICE lacks the required flexibility to allow saving averages (used for diagnostics) and the snap shots needed for the DA.

PR #1822, opened 2025-07-29
A C++ OOPS utility for IMS SCF preprocessing

CoryMartin-NOAA:
# Description

This PR introduces a C++ OOPS based utility to preprocess SCF (such as from IMS) based on the model background and produces pseudo observations in the IODA format. The intention is to replace a two step Fortran and then Python approach with this single C++ executable. Once tested and merged, this utility should remove the need for the sorc/land-imsproc (and hopefully the sorc/iodaconv) submodules.

# Companion PRs

None

# Issues

None, whoops.

- [ ] atm_jjob 
- [ ] C96C48_ufs_hybatmDA 
- [ ] C96C48_hybatmsnowDA 
- [ ] C96_gcafs_cycled 
- [ ] C48mx500_3DVarAOWCDA 
- [ ] C48mx500_hybAOWCDA 
- [ ] C96C48_hybatmDA


CoryMartin-NOAA:
An example YAML to run this utility:
```
geometry:
  fms initialization:
    namelist filename: ./fv3jedi/fmsmpp.nml
    field table filename: ./fv3jedi/field_table
  akbk: ./fv3jedi/akbk.nc4
  layout:
  - 1
  - 1
  npx: 97
  npy: 97
  npz: 127
time invariant fields:
  state fields:
    datetime: '2021-12-21T00:00:00Z'
    filetype: fms restart
    skip coupler file: true
    state variables:
    - filtered_orography
    - fraction_of_land
    field io names:
      filtered_orography: orog_filt
      fraction_of_land: land_frac
    datapath: /scratch3/NCEPDEV/stmp/Cory.R.Martin/aug2025/g-w-aero-bug/fix/orog/C96/
    filename_orog: C96.mx500_oro_data.nc
    filename_sfcd: 20211221.000000.sfc_data.nc
    filename_cplr: 20211221.000000.coupler.res
date: '2021-12-21T00:00:00Z'
output ioda file: './obs/gdas.oops.t00z.ims_snow.tm00.nc'
input scf file: './obs/ims2021355_4km_v1.3.asc'
mapping file: './obs/IMS4km_to_FV3_mapping.C96_oro_data.nc'
variables:
- totalSnowDepth
- vtype
- slmsk
- sheleg
- filtered_orography
- fraction_of_ice
- fraction_of_land
- stc
background:
  datapath: ./bkg/
  filetype: fms restart
  skip coupler file: true
  datetime: '2021-12-21T00:00:00Z'
  state variables:
  - totalSnowDepth
  - vtype
  - slmsk
  - sheleg
  - filtered_orography
  - fraction_of_ice
  - fraction_of_land
  - stc
  field io names:
    totalSnowDepth: snodl
    filtered_orography: orog_filt
    fraction_of_ice: fice
    fraction_of_land: land_frac
  filename_sfcd: 20211221.000000.sfc_data.nc
  filename_cplr: 20211221.000000.coupler.res
  filename_orog: C96.mx500_oro_data.nc
```

CoryMartin-NOAA:
I have confirmed that results from this code and the existing Fortran + Python converter approach are identical down to several decimal places for a C96 case that is in the Global-Workflow CI, which I think is "good enough" to say it is working as intended.

CoryMartin-NOAA:
Note that this is just the first step. Subsequent PRs will need to be made to global-workflow, jcb-gdas, and GDASApp to actually use this code.

utils/land/gdas_calc_scf_to_ioda.cc
```diff
@@ -0,0 +1,724 @@
...
+  auto bkg_snd = atlas::array::make_view<double, 2>(xBfs["totalSnowDepth"]);
+  auto bkg_density = atlas::array::make_view<double, 2>(xBfs["snowDensity"]);
+  // now compute density
+  for (atlas::idx_t jnode = 0; jnode < xBfs["totalSnowDepth"].shape(0); ++jnode) {
+    if (bkg_snd(jnode, 0) > 0.01) {
+      // snow is present, compute density
+      bkg_density(jnode, 0) = bkg_swe(jnode, 0) / bkg_snd(jnode, 0);
+    } else {
+      // snow is not present, use average from snow forecasts over land
+      double tmp_density = 67.92 + 51.25 * std::exp((bkg_stc(jnode, 0)- 273.15) / 2.59);
```
DavidNew-NOAA: In general, wherever physical constants occur, I would suggest referencing those from https://github.com/JCSDA-internal/fv3-jedi/blob/develop/src/fv3jedi/Utilities/Constants.cc

utils/land/CMakeLists.txt
DavidNew-NOAA, line 1: One pedantic suggestions

In https://github.com/NOAA-EMC/GDASApp/pull/1728 we established a naming protocol for GDASApp utilities of `gdas_(jedi submodule or model interface)_yada_yada.x`. In this case you're using both FV3-JEDI and IODA, but these applications are really FV3-JEDI based, so I would suggest `gdas_fv3jedi_scf_to_ioda.cc` (or `gdas_fv3jedi_scf2ioda.cc` if you feel like we're getting too wordy)

DavidNew-NOAA (commented):
Minor comments

utils/land/gdas_calc_scf_to_ioda.cc
CoryMartin-NOAA: Are any of these in that file? Note this is the average snow density not air density

utils/land/CMakeLists.txt
CoryMartin-NOAA: ok yes I can/should do this.

CoryMartin-NOAA: @copilot can you commit the suggestion @DavidNew-NOAA has here (rename the files and change the CMakeLists.txt)?

RussTreadon-NOAA:
**WCOSS2 GDASApp ctests**

Install `feature/land-scf-cpp` at 49f5c01 inside g-w from g-w PR [#3951](https://github.com/NOAA-EMC/global-workflow/pull/3951).  Run  GDASApp ctests.   All tests _Passed_
```
Test project /lfs/h2/emc/da/noscrub/russ.treadon/git/global-workflow/pr3951/sorc/gdas.cd/build
        Start 1967: test_gdasapp_C96C48_ufs_hybatmDA
        Start 1948: test_gdasapp_C96C48_hybatmDA
        Start 1991: test_gdasapp_C96_gcafs_cycled
        Start 2006: test_gdasapp_C96C48_hybatmsnowDA
        Start 2027: test_gdasapp_C48mx500_3DVarAOWCDA
        Start 2038: test_gdasapp_C48mx500_hybAOWCDA
        Start 2054: test_gdasapp_setup_atm_jjob_cycled_exp
        Start 1557: test_gdasapp_util_prepdata
        Start 1556: test_gdasapp_util_coding_norms
        Start 1941: test_gdasapp_check_python_norms
        Start 1942: test_gdasapp_check_yaml_keys
        Start 1943: test_gdasapp_jcb_installation
  1/136 Test #1942: test_gdasapp_check_yaml_keys .............................................   Passed    1.39 sec
        Start 1944: test_gdasapp_snow_create_ens
  2/136 Test #1943: test_gdasapp_jcb_installation ............................................   Passed    2.63 sec

...

135/136 Test #1990: test_gdasapp_C96C48_ufs_hybatmDA_enkfgdas_fcst_202402240000 ..............   Passed  261.56 sec
136/136 Test #1982: test_gdasapp_C96C48_ufs_hybatmDA_gdas_fcst_202402240000 ..................   Passed  424.76 sec

100% tests passed, 0 tests failed out of 136

Label Time Summary:
gdas-utils    =  25.88 sec*proc (10 tests)
gdasapp       = 21088.69 sec*proc (116 tests)
script        = 21114.58 sec*proc (126 tests)

Total Test time (real) = 2881.21 sec
```

utils/land/gdas_calc_scf_to_ioda.cc
DavidNew-NOAA: Triple point/melting point of water `273.15`, thought Constants.cc has `273.16`. But you're right. It just makes me nervous to see physical constants as numerical values floating around like these. Could you create constants in the .h file and reference them here?

CoryMartin-NOAA: This isn't the triple point of water, this is a C to K conversion factor. I'd have to ask @ClaraDraper-NOAA what/where exactly these values come from since I "borrowed" them from her original Fortran code

CoryMartin-NOAA: image

I think my Google-fu has found that this is just the fit to this curve and not physical constants

DavidNew-NOAA: Right, yeah I glanced over this line too fast and just saw 273.15. 
If these values are never going to change, ie no tuning, then a code comment about where the formula comes from is probably fine. If they will ever be adjusted in the future, then I would suggest making them constants in a header file

DavidNew-NOAA (dismissed):

parm/jcb-gdas
RussTreadon-NOAA, line 1: `feature/land-scf-cpp` points at [jcb-gdas @ 7cfaaa2](https://github.com/noaa-emc/jcb-gdas/tree/7cfaaa28c2ac977cbe5fcd7c7eba678ea94c7480).  This hash is in jcb-branch `feature/land-scf-cpp`.  Once we verify that the changes in this PR work as intended, we need to merge jcb -gdas PR [#161](https://github.com/NOAA-EMC/jcb-gdas/pull/161) into jcb-gdas `develop` and update the jcb-gdas hash in this PR.

RussTreadon-NOAA (commented):
Do any of the existing GDASApp ctests exercise the changes in this PR?

parm/jcb-gdas
CoryMartin-NOAA: No, they do not exercise it yet. I still need to make the necessary global-workflow changes. I thought it would be easier to get the code in first, to reduce circular dependency woes

RussTreadon-NOAA: Makes sense.  We can update GDASApp `develop` with the target changes and then test with a locally modified g-w.  This will tell us whether or not additional changes are needed in jcb-gdas or GDASApp.  If no changes are needed, we can open a g-w issue and PR to make the g-w changes needed to exercise this new functionality.

Do we plan to merge jcb -gdas PR [#161](https://github.com/NOAA-EMC/jcb-gdas/pull/161) into jcb-gdas `develop` and update the jcb-gdas hash in `feature/land-scf-cpp` before we merge this PR into GDASApp `develop`?

CoryMartin-NOAA: No, I think I'd prefer to get the source code in first, just in case I need to modify the templates in jcb-gdas at all as part of the workflow integration

RussTreadon-NOAA: OK

CoryMartin-NOAA: I will revert this hash update

RussTreadon-NOAA (dismissed):
Approve to get source code changes into GDASApp `develop`

We will need to carefully watch subsequent GDASApp PRs to see if they alter the jcb-gdas hash.  If they do, the source code added by this PR won't contain the expected templates.

parm/jcb-gdas
RussTreadon-NOAA, line 1: [jcb-gdas @ 6ae2a24](https://github.com/noaa-emc/jcb-gdas/tree/6ae2a24aa6815079dd190391598d2f76919a6f24) is one jcb-gdas hash ahead of GDASApp `develop`.  This should be OK.

RussTreadon-NOAA (dismissed):
Approve.

parm/jcb-gdas
CoryMartin-NOAA: Yes it adds something I just merged in an hour ago that is irrelevant to this change

RussTreadon-NOAA: Agreed.  This is fine.

utils/land/gdas_fv3jedi_calc_scf_to_ioda.h
```diff
@@ -0,0 +1,81 @@
...
+     private:
+      const fv3jedi::Geometry & geom_;
+      std::string imspath_;
+      std::string weightspath_;
+      std::vector<std::vector<int>> IMS_flag;
+      std::vector<std::vector<std::vector<int>>> IMS_index;
+      std::vector<std::vector<std::vector<float>>> lonFV3, latFV3, oroFV3;
+      static constexpr int nodata_int = -999;
+      static constexpr float nodata_float = -999.0f;
+      static constexpr float nodata_tol = 0.01f;  // Tolerance for nodata checks
```
jiaruidong2017: Reminder: In the original code, **nodata_tol** was set to 0.1.

CoryMartin-NOAA:
please don't merge until we have @ClaraDraper-NOAA 's approval since this is a rewrite of her original code into C++

RussTreadon-NOAA:
Thanks @CoryMartin-NOAA for the note.  We can remove the **DO NOT MERGE** label after @ClaraDraper-NOAA reviews and approves.

utils/land/gdas_fv3jedi_calc_scf_to_ioda.h
```diff
@@ -0,0 +1,81 @@
...
+    };
+    static constexpr std::array<float, 20> scffac_table = {
+        0.005f, 0.005f, 0.005f, 0.005f, 0.005f, 0.008f,
+        0.008f, 0.010f, 0.010f, 0.010f, 0.010f, 0.007f, 0.021f,
+        0.013f, 0.015f, 0.008f, 0.015f, 0.015f, 0.015f, 0.015f
+    };
+    float oberr_scf = 0.0f;
+    float oberr_snd = 80.0f;
+    static constexpr float nodata_float = -999.0f;
+    static constexpr float nodata_tol = 0.01f;  // Tolerance for nodata checks
```
jiaruidong2017: Same as above

ClaraDraper-NOAA:
> I have confirmed that results from this code and the existing Fortran + Python converter approach are identical down to several decimal places for a C96 case that is in the Global-Workflow CI, which I think is "good enough" to say it is working as intended.

I agree, this is good enough.

ClaraDraper-NOAA (commented):
Thanks for doing this @CoryMartin-NOAA . What are the consequences for the offline workflow? I am working hard to return to both workflows calling the exact same code, so if we adopt this in the global_workflow, we'll need to use it offline too. What additional dependencies might this introduce?

CoryMartin-NOAA:
@ClaraDraper-NOAA the plan is to modify the snow_analysis class in the global-workflow to call this instead. I think the long term result will be simplifying the way in which the workflow generates the IMS obs. Right now we have that IMS SCF Fortran code and the ioda-converter, so it is a 2 step process. This makes it a one step process. Only issue is that this has to run with 6 PEs, instead of 1. Since you are already using GDASApp for JEDI executable maintenance, there should not be any additional dependencies.

CoryMartin-NOAA:
I'll add that you would have to either 1) use JCB or 2) have the script in your workflow that calls this produce this YAML instead of the namelist for the existing Fortran code

ClaraDraper-NOAA:
> I'll add that you would have to either 1) use JCB or 2) have the script in your workflow that calls this produce this YAML instead of the namelist for the existing Fortran code

Thanks. I'm currently adding JCB, and an separate copy of the snow_analysis task - so this should all be fine. 

Note: I've re-arranged the snow_analysis task to call prep_IMS before initialize, so that we can then use the obs chronicle for IMS instead of treating it separately. I'll start on a draft PR tomorrow now.

ClaraDraper-NOAA (dismissed):

jiaruidong2017 (dismissed):
Great efforts!

utils/land/gdas_fv3jedi_calc_scf_to_ioda.h
CoryMartin-NOAA: Fantastic catch! Thanks.

DavidNew-NOAA (approved):

PR #1830, opened 2025-07-31
Update paths for run_ufo_hofx_test.sh

ADCollard:
Updating paths so that test data is visible from URSA.

Note this does not update the paths in `config_gsi.yaml` and `config.jedi.yaml`

### DIFF
--- a/ush/ufoeval/run_ufo_hofx_test.sh
+++ b/ush/ufoeval/run_ufo_hofx_test.sh
@@ -67,9 +67,9 @@ obtype=$1
 
 #--------------- User modified options below -----------------
 
-machine=${machine:-orion} 
+machine=${machine:-ursa} 
 
-if [ $machine = orion || $machine = hercules ]; then
+if [[ "$machine" = "orion" || "$machine" = "hercules" ]]; then
    if [ $run_filtering == NO ]; then
       workdir=/work2/noaa/da/$LOGNAME/ufoeval/$cycle/${obtype}_noqc
       echo "Run without data filtering"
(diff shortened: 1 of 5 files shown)


ush/ufoeval/run_ufo_hofx_test.sh
```diff
@@ -79,11 +79,11 @@ if [ $machine = orion ]; then
    GDASApp=${GDASApp:-/work2/noaa/da/$LOGNAME/git/GDASApp/} # Change this to your own branch
    JCBinstall=${JCBinstall:-/work2/noaa/da/cmartin/CI/GDASApp/opt}
    JCBpylib=$JCBinstall/lib/python3.7/site-packages
-elif [ $machine = hera ]; then
```
CoryMartin-NOAA, line 82: Can we still support Hera? The compute nodes aren't going away immediately

RussTreadon-NOAA: Correct.  Hera compute will remain past 8/5/2025

> 08/05* Hera scratch 1-2 disks decommissioned (Hera CPU nodes still available)
>*CPU allocations remain on Hera and users can run jobs using the cross-mounted Ursa disks. Hera access
will remain for as long as there is funding.

ADCollard: > Can we still support Hera? The compute nodes aren't going away immediately

OK, I added it back and also added hercules

ush/ufoeval/run_ufo_hofx_test.sh
```diff
@@ -114,12 +114,12 @@ else
...
 fi
 
-if [ $machine = orion ]; then
+if [ $machine = orion || $machine = hercules ]; then
     export Datapath='/work2/noaa/da/acollard/UFO_eval/data/gsi_geovals_l127/nofgat_feb2024/'$dataprocdate 
     FixDir=/work2/noaa/da/cmartin/GDASApp/fix
-elif [ $machine = hera ]; then
-    export Datapath='/scratch1/NCEPDEV/da/Emily.Liu/UFO_eval/data/gsi_geovals_l127/nofgat_aug2021/'$dataprocdate
-    FixDir=/scratch1/NCEPDEV/da/Cory.R.Martin/GDASApp/fix
+elif [ $machine = ursa || $machine" = hera ]; then
```
CoryMartin-NOAA:
```suggestion
elif [ $machine = ursa || $machine = hera ]; then
```

RussTreadon-NOAA (dismissed):
Didn't test but looks good to me.

Approve.

ADCollard:
Converting to draft while I update the test data and re-test.

PR #1958, opened 2025-10-28
Add 3DVar atmosphere regression test

### LINKED ISSUE #1953: Add atmospheric 3DVar regression test
To facilitate end-to-end testing and prototype cycling, the underpinnings for a regression test need to be added. This ctest will run from ObsForge generated observations, and use JCB to create the YAML, and run an atmospheric 3DVar. Thus, we can know for sure before cycling in the global workflow if the observations will work properly. Eventually, this will be the basis for a regression test to ensure results do not change when updating the software versions.

CoryMartin-NOAA:
# Description

This PR adds the framework for an atmospheric C384 3DVar regression test. In the short term, this can be used to turn on additional observation types for @RussTreadon-NOAA 's cycling prototype but longer term can serve as the basis for a more rigorous, realistic regression testing system.

This is in draft until

- [ ] input data are staged on supported platforms
- [ ] compute resource requirements are fleshed out for supported platforms

But it works on Ursa and would appreciate any general feedback so far.

# Companion PRs

None

# Issues
Closes #1953 

- [ ] atm_jjob 
- [ ] C96C48_ufs_hybatmDA 
- [ ] C96C48_hybatmsnowDA 
- [ ] C96_gcafs_cycled 
- [ ] C48mx500_3DVarAOWCDA 
- [ ] C48mx500_hybAOWCDA 
- [ ] C96C48_hybatmDA


CoryMartin-NOAA:
This will remain in draft until input data can be copied to all supported platforms but that should wait until any issues with https://github.com/NOAA-EMC/JEDI-T2O/issues/147 are resolved and the file naming conventions are agreed to

DavidNew-NOAA:
Why need to do so much manually in `regression_test_3dvar.py`? Why can't you use existing infrastructure in GW and GDASApp like the `AtmAnalysis` class and `jcb_base.yaml.j2`?

CoryMartin-NOAA:
@DavidNew-NOAA I explicitly do not want to use the global workflow infrastructure for this, but rather a regression test to validate the results of the code independent of a workflow, such as when we update the hashes weekly.  We could possibly use jcb-base.yaml but would still require a preprocessor to define all of the environment variables.

DavidNew-NOAA:
@CoryMartin-NOAA OK. My preference would be to define all the required environment variables and render `jcb-base.yaml.j2`, but I'm not married to the idea. Up to you.

CoryMartin-NOAA:
@DavidNew-NOAA you mean use `wxflow` and create a jcb-base.yaml from a template?

DavidNew-NOAA:
@CoryMartin-NOAA Yes, or python `Jinja` if you consider `wxflow` to be part of the "workflow"

test/atm/regression_test_3dvar.sh
```diff
@@ -0,0 +1,290 @@
...
+cp -rf $fix_root/fv3jedi/${fv3jedi_ver}/fv3files/fmsmpp.nml $rundir/fv3jedi/fmsmpp.nml
+cp -rf $fix_root/fv3jedi/${fv3jedi_ver}/fv3files/field_table_gfdl $rundir/fv3jedi/field_table
+
+# Copy observations
+mkdir -p $rundir/obs
+cp -rf $GDASAPP_REGRESSION_TEST_DATA_PATH/obs/gdas.${PDY}/${cyc}/atmos/* $rundir/obs/.
+
+# Copy backgrounds
+mkdir -p $rundir/bkg
+cp -rf $GDASAPP_REGRESSION_TEST_DATA_PATH/bkg/${CASE}/gdas.${gPDY}/${gcyc}/model/atmos/history/${GPREFIX}cubed_sphere_grid_atmf006.nc $rundir/bkg/.
```
DavidNew-NOAA: I would still use the new workflow file-naming conventions (csg_atm.f006)

CoryMartin-NOAA: agreed

CoryMartin-NOAA:
@DavidNew-NOAA thinking about this a bit more now that I'm revisiting it. Are you sure it makes sense to create a template with the variables, define the variables in this script, then generate a YAML rather than just creating the jcb-base.yaml here in the bash script? I don't want to sound like a luddite but it seems like that would require extra steps from doing it this way, but maybe I'm being dense.

DavidNew-NOAA:
@CoryMartin-NOAA I'm not married to the idea of using the `jcb-base.yaml.j2` from GDASApp. It really depends on which infrastructure you're trying to test. Since this is a regression test which tests whether results are changing, likely due to obs or JEDI algorithm changes, than it's probably not necessary. I thought it might make things easier, but if it's over complicating things, let's keep it simple.

CoryMartin-NOAA:
@DavidNew-NOAA oh I see, you want to use the jcb-base.yaml.j2 that is here: https://github.com/NOAA-EMC/GDASApp/blob/develop/parm/atm/jcb-base.yaml.j2 rather than making another one. I think there are a few things we need to make more modular before we can do that (observations: all_observations for example) but I like that idea longer term

PR #1974, opened 2025-11-10
Bugfix for cice restart fix utility

shlyaeva:
# Description

The cice restart fix utility on develop compares ice and snow volumes (instead of thicknesses) to the threshold. This PR fixes it to compare the thicknesses.

The behavior on develop leads to this script missing some of the worst conditions. I have encountered a case, e.g., where we have ice concentration of 0.00055 in one of ice categories and ice volume of 0.4 for that category, resulting in ~700m ice thickness in less than tenth of percent of a gridpoint (I'd like to see that ice tower in real life!). This gridpoint doesn't get picked up by the script as on develop, but gets fixed with the update here.

- [ ] atm_jjob 
- [ ] C96C48_ufs_hybatmDA 
- [ ] C96C48_hybatmsnowDA 
- [ ] C96_gcafs_cycled 
- [ ] C48mx500_3DVarAOWCDA 
- [ ] C48mx500_hybAOWCDA 
- [ ] C96C48_hybatmDA

### DIFF
--- a/utils/soca/fixcicerst/fix_cice_restart.py
+++ b/utils/soca/fixcicerst/fix_cice_restart.py
@@ -39,11 +39,12 @@ def load_config(config_file):
         sys.exit(1)
 
 
-def calculate_aggregated_thickness(vicen, vsnon):
+def calculate_aggregated_thickness(aicen, vicen, vsnon):
     """
     Calculate aggregated ice and snow thickness across categories.
 
     Parameters:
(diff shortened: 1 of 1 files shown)


guillaumevernieres (approved):
Oops ... Good catch and thanks again for digging into this.

JohnSteffen-NOAA (approved):

PR #1988, opened 2025-11-22
Add soil temp QC to IMS pre-processing

### LINKED ISSUE #1986: Add soil temp QC to IMS pre-processing
Snow depth increments are not added if the soil temperature is above freezing. 

We then need to also QC out non-zero snow depth obs if the soil temperature is above freezing. This is difficult to do in JEDI, so for now we will do it in the IMS pre-processing stage - which already reads in the soil temperatures. 

See https://github.com/NOAA-EMC/jcb-gdas/issues/210

jiaruidong2017:
# Description

This PR adds soil temperature QC to IMS pre-processing by removing the snow obs if the IMS derived snow depth > 0 and model soil temp in layer 1 is > 273.155.

# Issues
Resolves #1986 

- [ ] atm_jjob 
- [ ] C96C48_ufs_hybatmDA 
- [ ] C96C48_hybatmsnowDA 
- [ ] C96_gcafs_cycled 
- [ ] C48mx500_3DVarAOWCDA 
- [ ] C48mx500_hybAOWCDA 
- [ ] C96C48_hybatmDA

### DIFF
--- a/utils/land/gdas_fv3jedi_calc_scf_to_ioda.cc
+++ b/utils/land/gdas_fv3jedi_calc_scf_to_ioda.cc
@@ -677,7 +677,7 @@ void gdasapp::CalcSCFtoIODA::IMSscf::updateIMSsd(fv3jedi::State &state,
         this->sndIMS[fv3_j][fv3_i] = -10.0f;
       }
       if ((bkg_stc(jnode, 0) > 273.155) && (this->sndIMS[fv3_j][fv3_i] > 0)) {
-         // if soil too warm and snow depth > 0,
+         // if model soil is too warm to add snow, and the observed snow depth is >0,
          // set the IMS snow depth to a fixed value to QC in JEDI
         this->sndIMS[fv3_j][fv3_i] = -20.0f;
       }
--- a/utils/land/gdas_fv3jedi_calc_scf_to_ioda.cc
+++ b/utils/land/gdas_fv3jedi_calc_scf_to_ioda.cc
@@ -121,7 +121,7 @@ void gdasapp::CalcSCFtoIODA::calc_fcst_snow_density(fv3jedi::State & bkgState,
       // snow is not present, use average from snow forecasts over land
       // The below line comes from NOAH-MP, fit to a curve of snow density from:
       // hedstrom nr and jw pomeroy (1998), hydrol. processes, 12, 1611-1625
-      double tmp_density = 67.92 + 51.25 * std::exp((bkg_stc(jnode, 0)- 273.155) / 2.59);
+      double tmp_density = 67.92 + 51.25 * std::exp((bkg_stc(jnode, 0)- 273.15) / 2.59);
       bkg_density(jnode, 0) =
         std::max(80.0, std::min(120.0, tmp_density)) / 1000.0;
     }
--- a/utils/land/gdas_fv3jedi_calc_scf_to_ioda.cc
+++ b/utils/land/gdas_fv3jedi_calc_scf_to_ioda.cc
@@ -121,7 +121,7 @@ void gdasapp::CalcSCFtoIODA::calc_fcst_snow_density(fv3jedi::State & bkgState,
       // snow is not present, use average from snow forecasts over land
       // The below line comes from NOAH-MP, fit to a curve of snow density from:
       // hedstrom nr and jw pomeroy (1998), hydrol. processes, 12, 1611-1625
-      double tmp_density = 67.92 + 51.25 * std::exp((bkg_stc(jnode, 0)- 273.15) / 2.59);
+      double tmp_density = 67.92 + 51.25 * std::exp((bkg_stc(jnode, 0)- 273.155) / 2.59);
       bkg_density(jnode, 0) =
         std::max(80.0, std::min(120.0, tmp_density)) / 1000.0;
     }
@@ -654,7 +654,8 @@ void gdasapp::CalcSCFtoIODA::IMSscf::updateIMSsd(fv3jedi::State &state,
   // convert the state to an atlas fieldset
   atlas::FieldSet xBfs;
   state.toFieldSet(xBfs);
-  // Get the snow cover fraction and depth fields from the state
+  // Get the necessary fields from the state
+  auto bkg_stc = atlas::array::make_view<double, 2>(xBfs["stc"]);
   auto bkg_scf = atlas::array::make_view<double, 2>(xBfs["surface_snow_area_fraction"]);
   auto bkg_snd = atlas::array::make_view<double, 2>(xBfs["totalSnowDepth"]);
   const auto bkg_idx =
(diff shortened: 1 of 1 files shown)


yuanxue2870:
@jiaruidong2017: thanks for getting in the PR so quickly! The revised code itself looks OK to me. My only concern is this additional if condition seems a bit too conservative, which may QC out lots of observations. Since the issue mainly occurs when consistently high biased (O-F) present when IMS SCF=100% but the modeled SCF is unable to catch up (for various reasons). Shall we consider a revised if statement like: model soil temp layer1 > 273.155 & ims_scf > 95%? Which means we are still trying to retain the points where has 50%-95% IMS SCF (with converted snow depth > 0) and the modeled soil temperature may be warm in this case. Just my two cents. Feel free to reject if you disagree.

jiaruidong2017:
Thanks @yuanxue2870 for your quick review and suggestions. 

I checked the two time-slot stats files. For 2024-11-01 00Z, there are 1,120 observations assigned a value of –20, out of 181,725 total observations, with 47,873 passing. This means only 2.35% of the passed IMS observations were masked by this condition.

For 2024-11-02 00Z, 1,773 observations were masked with –20, compared to 47,552 passed observations, which corresponds to 3.73%.

Therefore, only a very small portion of the IMS observations is masked out by this condition.

utils/land/gdas_fv3jedi_calc_scf_to_ioda.cc
```diff
@@ -654,7 +654,8 @@ void gdasapp::CalcSCFtoIODA::IMSscf::updateIMSsd(fv3jedi::State &state,
   // convert the state to an atlas fieldset
   atlas::FieldSet xBfs;
   state.toFieldSet(xBfs);
-  // Get the snow cover fraction and depth fields from the state
+  // Get the necessary fields from the state
+  auto bkg_stc = atlas::array::make_view<double, 2>(xBfs["stc"]);
```
ClaraDraper-NOAA, line 658: I'm not familiar with Atlas or C++. Is it necessary here to indicate that STC has an additional vertical dimension (I'm wondering is that "2" is the dimension.

ClaraDraper-NOAA:
> @jiaruidong2017: thanks for getting in the PR so quickly! The revised code itself looks OK to me. My only concern is this additional if condition seems a bit too conservative, which may QC out lots of observations. Since the issue mainly occurs when consistently high biased (O-F) present when IMS SCF=100% but the modeled SCF is unable to catch up (for various reasons). Shall we consider a revised if statement like: model soil temp layer1 > 273.155 & ims_scf > 95%? Which means we are still trying to retain the points where has 50%-95% IMS SCF (with converted snow depth > 0) and the modeled soil temperature may be warm in this case. Just my two cents. Feel free to reject if you disagree.

We should keep the check as Jiarui has it. The reason we're seeing consistent positive O-F at many locations is because the temp is above freezing, and no snow is added. By QCing out the non-zero obs, we solve this problem.

utils/land/gdas_fv3jedi_calc_scf_to_ioda.cc
```diff
@@ -675,6 +676,11 @@ void gdasapp::CalcSCFtoIODA::IMSscf::updateIMSsd(fv3jedi::State &state,
          // set the IMS snow depth to a fixed value to QC in JEDI
         this->sndIMS[fv3_j][fv3_i] = -10.0f;
       }
+      if ((bkg_stc(jnode, 0) > 273.155) && (this->sndIMS[fv3_j][fv3_i] > 0)) {
+         // if soil too warm and snow depth > 0,
```
ClaraDraper-NOAA: Can you please change this to "is model soil is too warm to add snow, and the observed snow depth is >0"

utils/land/gdas_fv3jedi_calc_scf_to_ioda.cc
CoryMartin-NOAA: Pretty much all of the Atlas fieldsets in JEDI are "2D". The first dimension is the unstructured grid dimension and the 2nd is the vertical. It is size 1 if the field is not 3D in the model.

ClaraDraper-NOAA: Got it, thanks. I see the jnode dimension now!

ClaraDraper-NOAA:
Thanks @jiaruidong2017. Have you run any experiments to check that the new QC is behaving as expected?

utils/land/gdas_fv3jedi_calc_scf_to_ioda.cc
CoryMartin-NOAA: yeah by far the most complicated part of this code for me was figuring out how to go to/from Atlas "j-node" space and FV3 cube i,j,tile space

utils/land/gdas_fv3jedi_calc_scf_to_ioda.cc
jiaruidong2017: Updated as suggested. Thanks.

jiaruidong2017:
> Thanks @jiaruidong2017. Have you run any experiments to check that the new QC is behaving as expected?

I ran several cycles to confirm the updated codes work, but I haven’t evaluated their behavior yet. I was struggling with the G-W debugging related to the recent rename changes, but that issue is finally resolved.

Since Yuan identified the IMS issues from the RT runs, do you think I should run tests for the same period to check whether the updates address the problem? Do you have any suggestions on how I should proceed with the validation?

ClaraDraper-NOAA:
> > Thanks @jiaruidong2017. Have you run any experiments to check that the new QC is behaving as expected?
> 
> I ran several cycles to confirm the updated codes work, but I haven’t evaluated their behavior yet. I was struggling with the G-W debugging related to the recent rename changes, but that issue is finally resolved.
> 
> Since Yuan identified the IMS issues from the RT runs, do you think I should run tests for the same period to check whether the updates address the problem? Do you have any suggestions on how I should proceed with the validation?

If you have confirmed that this check is successfully preventing the expected obs from being assimilated, I'm OK to proceed. we can check the behavior in the offline system. 

I'm approving now, and will defer to @CoryMartin-NOAA on what testing he'd like.

ClaraDraper-NOAA (approved):
Thanks @jiaruidong2017 !

jiaruidong2017:
> > > Thanks @jiaruidong2017. Have you run any experiments to check that the new QC is behaving as expected?
> > 
> > 
> > I ran several cycles to confirm the updated codes work, but I haven’t evaluated their behavior yet. I was struggling with the G-W debugging related to the recent rename changes, but that issue is finally resolved.
> > Since Yuan identified the IMS issues from the RT runs, do you think I should run tests for the same period to check whether the updates address the problem? Do you have any suggestions on how I should proceed with the validation?
> 
> If you have confirmed that this check is successfully preventing the expected obs from being assimilated, I'm OK to proceed. we can check the behavior in the offline system.
> 
> I'm approving now, and will defer to @CoryMartin-NOAA on what testing he'd like.

Out of the 1120 IMS observations marked as –20, about 300 that were previously assimilated were removed by this condition. Others should be removed previously by other filters. See the parallel run comparison below.
**Without the warm soil condition:** 
image
**With the warm soil condition:**
image

CoryMartin-NOAA (approved):
This looks good to me, since it is confirmed to work. Thanks @jiaruidong2017

PR #2015, opened 2025-12-19
Remove GSI-B from release branch

CoryMartin-NOAA:
# Description

Removes gsibec as it is not needed for GFSv17 plus add a readme for release tags for each repo


RussTreadon-NOAA:
Are we removing these modules to simplify GFS v17 EE2 compliance?

I view GFS v17 as the path to implement JEDI components.  While GFS v18 may not use GSI-B, it could.  The same remark could apply to land-imsproc, (though I don't know where/how this module is used in GDASApp `develop.`).

CoryMartin-NOAA:
@RussTreadon-NOAA for GSI-B, yes, and for the land one, I think we can remove it from `develop` also, but I want to double check that when @ClaraDraper-NOAA is back from leave.

Now that we are in this release branch, it is intended to diverge somewhat from `develop`

RussTreadon-NOAA:
@CoryMartin-NOAA . OK. I try to keep implementations branches as close as possible to `develop` for as long as possible.  While diverging is inevitable, managing divergent branches can become tedious.

CoryMartin-NOAA:
@RussTreadon-NOAA do you think it makes more sense to, for example, open a PR to both develop and the release branch removing the Land submodule? I'm pretty sure we are already past the point of no divergence for this branch

RussTreadon-NOAA:
@CoryMartin-NOAA :  If `land-imsproc` isn't needed in `develop`, I'm all for removing it.  I agree.  It would be good to check with @ClaraDraper-NOAA

RussTreadon-NOAA (approved):
This PR does what it advertises:  Submodules `gsibec` and `land-imsproc` are removed from `release/gfs.v17`.

I didn't attempt to build `release/gfs.v17` or run ctests.  Multiple ctests will fail after `gsibec` is removed.

Approve.

CoryMartin-NOAA:
Thanks @RussTreadon-NOAA but you raised a good point about the land repo, let's not merge this in and see if we can do a separate PR to both develop and the release branch that removes the land repo, in the new year.

CatherineThomas-NOAA:
FWIW I agree with the path forward to remove the land repo from both develop and the release branch at the same time if able.  It'll be a cleaner github path to trace later down the road.

ClaraDraper-NOAA:
Confirming that we can remove the land-imsproc from GDASApp. 

Once we have the IMS netcdf file reading added to land-imsproc, we can also delete it (as it's been replaced by the c++ version). I'll create an issue for that now.

ClaraDraper-NOAA:
Not sure if this is appropriate here, but if we're doing clean-up then [the regridding code in da-utils](https://github.com/NOAA-EMC/DA-utils/tree/develop/src/regridStates) can also be removed, as it's now in UFS_UTILS.

CoryMartin-NOAA:
Thanks @ClaraDraper-NOAA , I think we need a da-utils PR to remove that. We can do that soon.

CoryMartin-NOAA:
Also noting that I will open a PR into develop that mirrors this PR.

ClaraDraper-NOAA:
> Thanks @ClaraDraper-NOAA , I think we need a da-utils PR to remove that. We can do that soon.

I created an issue [here](https://github.com/NOAA-EMC/DA-utils/issues/54). I'll create a PR soon.

CoryMartin-NOAA:
Closing this PR as it is superseded by https://github.com/NOAA-EMC/GDASApp/pull/2025 and https://github.com/NOAA-EMC/GDASApp/pull/2026

PR #2158, opened 2026-06-01
Add Hyundeok's changes for insitu marine obs

CoryMartin-NOAA:
# Description

Correct insitu profile converter names after provider renaming

# Companion PRs

https://github.com/NOAA-EMC/obsForge/pull/232/changes
# Issues


apchoiCMD (approved):
Thanks for the update @CoryMartin-NOAA , I think this will fix marine in-situ obs-

PR #2253, opened 2026-09-14
Updated marine DA configuration

guillaumevernieres:
# Description                                                                                                                                                                                                           
                                                                                                                                                                                                                        
Marine DA update covering three largely independent threads: sea ice                                                                                                                                         
freeboard assimilation, a rebuilt static background error, and a substantially                                                                                                                                          
expanded in situ and altimeter observation set.                                                                                                                                                                
Sorry in advance for the verbosity of the AI generated PR description, but these changes are significant and require a detailed review.

**Results here:** [https://drive.google.com/file/d/1BdMnjtX9r26C8hZNyeSbB0NtaTzL13hu/view?usp=drive_link](https://drive.google.com/file/d/1BdMnjtX9r26C8hZNyeSbB0NtaTzL13hu/view?usp=drive_link)
                                                                                                                                                                                                                        
## 1. Sea ice freeboard assimilation (new)                                                                                                                                                                              
                                                                                                                                                                                                                        
Adds freeboard as an assimilated variable, from two instrument classes with                                                                                                                                             
different physics:                                                                                                                                                                                                      
                                                                                                                                                                                                                        
- `icefb_icesat2_{north,south}` -- laser altimeter, **total** freeboard                                                                                                                                                 
  (ice + snow).                                                                                                                                                                                                         
- `icefb_cryotempo_{north,south}` -- radar altimeter, **ice-only** freeboard,                                                                                                                                           
  assimilated with its own snow depth.                                                                                                                                                                                  
                                                                                                                                                                                                                        
Each observation space uses a Composite operator (`SeaIceFreeboard` +                                                                                                                                                   
`Identity` on ice fraction), a PreQC and physical bounds screen, domain checks                                                                                                                                          
on sea area fraction / ice concentration / SST, a background check, and an                                                                                                                                              
additive error floor applied through `ObsFunction/Arithmetic` so that the final                                                                                                                                         
sigma_o cannot fall below a believable value regardless of what the file                                                                                                                                                
reports.                                                                                                                                                                                                                
                                                                                                                                                                                                                        
Densities are 917 / 330 / 1026 kg/m^3 (ice / snow / seawater) and **must stay                                                                                                                                           
in sync** with `rho ice`, `rho snow`, `rho water` in the diagB configuration                                                                                                                                            
(see below).                                                                                                                                                                                                            
                                                                                                                                                                                                                        
## 2. Background error                                                                                                                                                                                                  
                                                                                                                                                                                                                        
### diagB (`utils/soca/diagb/`)                                                                                                                                                                                         
                                                                                                                                                                                                                        
- **hi/hs variance clipping** (new `gdas_soca_diagb_hihs_clip.h`). For laser                                                                                                                                            
  freeboard `H = alpha*hi + beta*hs`, so the freeboard-induced increment ratio                                                                                                                                          
  `delta_hi/delta_hs` is set by the variance ratio. A new `rmax` (default 3.0)                                                                                                                                          
  bounds that ratio, redistributing `sigma_hi^2` and `sigma_hs^2` so their                                                                                                                                              
  **sum is preserved** where the bound bites. Densities are configurable.                                                                                                                                               
- **Sea ice concentration floor** (`min aice`). `sea_ice_thickness` and                                                                                                                                                 
  `sea_ice_snow_thickness` are the ratios `hi/aice` and `hs/aice`, which blow                                                                                                                                           
  up as `aice -> 0`; a sliver at `aice ~ 1e-4` reads as hundreds of metres of                                                                                                                                           
  ice. Cells below the floor are excluded from the variance stencils and given                                                                                                                                          
  zero background error. `aice` itself is left unmasked so the analysis can                                                                                                                                             
  still move the ice edge.                                                                                                                                                                                              
- **Absolute stddev ceilings**: `max ssh`, `max aice`, `max hi`. `rmax` bounds                                                                                                                                          
  only the hi/hs *ratio*, so `max hi` is still required to stop both variances                                                                                                                                          
  growing together.                                                                                                                                                                                                     
- **Missing-value backward compatibility**: soca now fills masked/land cells                                                                                                                                            
  with a missing-value sentinel rather than 0; these are reset to 0 on read.                                                                                                                                            
                                                                                                                                                                                                                        
All new options are documented in `utils/soca/diagb/README.md`.                                                                                                                                                         
                                                                                                                                                                                                                        
### Static B (`marine_background_error_static_diffusion.yaml.j2`)                                                                                                                                                       
                                                                                                                                                                                                                        
Rewritten as a **dual-scale hybrid**: two BUMP_NICAS central blocks (200 km and                                                                                                                                         
100 km) combined by SABER's hybrid chain at 0.5/0.5 variance weights, with the                                                                                                                                          
outer blocks (StdDev, TorchBalance) applied **once** to the sum so the                                                                                                                                                  
torchscript emulators are loaded and evaluated a single time.                                                                                                                                                           
                                                                                                                                                                                                                        
**`BalanceSOCA` is replaced by a **TorchBalance** outer block carrying surface                                                                                                                                            
(ice fraction), vertical (steric SSH, heave salinity) and non-local                                                                                                                                                     
(geostrophy) emulators. The SOCA balance stanza is retained, commented out, so                                                                                                                                          
the two balance operators can be swapped.**   <-- I need to comment that out, `libtorch` is not part of the jedi stack yet                                                                                                                                                                           
                                                                                                                                                                                                                        
Measured at cycle 2025121906: the 100 km component reaches an effective                                                                                                                                                 
resolution of 2.75 against a `min effective resolution` of 2.7. That margin is                                                                                                                                          
thin, and 100 km is about the shortest scale the 1/2-degree analysis grid can                                                                                                                                           
support -- going shorter means refining `anl_geom`, not tuning NICAS.                                                                                                                                                   
                                                                                                                                                                                                                        
### Ensemble B (`marine_background_error_hybrid_diffusion_diffusion.yaml.j2`)                                                                                                                                           
                                                                                                                                                                                                                        
Replaced by a pure ensemble covariance with BUMP/NICAS localization (300 km),                                                                                                                                           
ported from the `envar_L300` configuration validated by a dirac test. No static                                                                                                                                         
component and no BalanceSOCA. NICAS runs in calibration mode each cycle, so                                                                                                                                             
parameters are recomputed rather than read, at the cost of per-cycle                                                                                                                                                    
calibration time.                                                                                                                                                                                                       
                                                                                                                                                                                                                        
## 3. Observations                                                                                                                                                                                                      
                                                                                                                                                                                                                        
### In situ profiles (the bulk of the new obs)                                                                                                                                                                          
                                                                                                                                                                                                                        
The combined `insitu_profile_*` files are replaced by **separate temperature                                                                                                                                            
and salinity streams** per platform, twelve files in all: Argo, glider, TESAC,                                                                                                                                          
PIRATA, RAMA and TAO/TRITON. Surface drifter temperature                                                                                                                                                                
(`insitu_temp_surface_drifter`) is also activated. Every one of these was                                                                                                                                               
commented out on `develop`, so this is the first cycle they are live.                                                                                                                                                   
                                                                                                                                                                                                                        
Each profile stream carries the same QC chain:                                                                                                                                                                          
                                                                                                                                                                                                                        
1. **Pre-filters** (credit: Kriti Bhargava) -- global range test, then regional                                                                                                                                         
   range tests for the marginal seas (Red Sea, Mediterranean, and others),                                                                                                                                              
   a below-bathymetry rejection comparing observation depth against                                                                                                                                                     
   `sea_floor_depth_below_sea_surface`, and a **Spike and Step Check** to throw                                                                                                                                         
   out stair-stepped profiles reported at coarse rounding.                                                                                                                                                              
2. **Post-filters** -- `Profile Background Check` (absolute threshold 0.5),                                                                                                                                             
   then a per-observation `Background Check` at 5 K / 5 psu.                                                                                                                                                            
3. **Error model**, in three stages, and the order matters because                                                                                                                                                      
   `assign error` overwrites while `inflate error` multiplies:                                                                                                                                                          
   - a constant error is assigned, discarding whatever the file reported;                                                                                                                                               
   - a **linear depth ramp** replaces it, giving the largest error at the                                                                                                                                               
     surface and decaying with depth, with a fraction of the original error                                                                                                                                             
     added back to represent out-of-window error;                                                                                                                                                                       
   - the **tropical inflation** is applied last.                                                                                                                                                                        
                                                                                                                                                                                                                        
### Tropical observation error inflation                                                                                                                                                                                
                                                                                                                                                                                                                        
Present on all twelve in situ profile streams, in two forms.                                                                                                                                                            
                                                                                                                                                                                                                        
**Basin-scale platforms** (Argo, glider, TESAC) use a Gaussian in latitude:                                                                                                                                             
                                                                                                                                                                                                                        
```                                                                                                                                                                                                                     
factor(lat) = 1 + (A - 1) * exp(-lat^2 / (2*sigma^2))    A = 4, sigma = 15 deg                                                                                                                                          
```                                                                                                                                                                                                                     
                                                                                                                                                                                                                        
peaking at 4 on the equator and decaying to 2.8 at 15 deg, 2.2 at 20 deg,                                                                                                                                               
1.4 at 30 deg and effectively 1 by 50 deg. It carries **no `where` clause on                                                                                                                                            
purpose** -- the factor is continuous in latitude, so there is no discontinuity                                                                                                                                         
at a band edge, and squaring the latitude rather than taking its absolute value                                                                                                                                         
keeps the curve smooth across the equator as well.                                                                                                                                                                      
                                                                                                                                                                                                                        
**Tropical moored arrays** (PIRATA, RAMA, TAO/TRITON) sit entirely within about                                                                                                                                         
20 degrees of the equator, so there is no latitude structure for the curve to                                                                                                                                           
resolve. They apply the same peak amplitude of 4 as a flat factor.                                                                                                                                                      
                                                                                                                                                                                                                        
In both cases this filter **must remain the last error filter** in the file;                                                                                                                                            
the `assign error` filters above it would otherwise overwrite the inflation.                                                                                                                                            
                                                                                                                                                                                                                        
### Depth ramp coefficients                                                                                                                                                                                             
                                                                                                                                                                                                                        
| Stream | Surface error | Depth slope |                                                                                                                                                                                
|---|---|---|                                                                                                                                                                                                           
| Temperature, all platforms | 1.0 K | -0.00045 |                                                                                                                                                                       
| Salinity, all platforms | 0.5 psu | -0.00025 |                                                                                                                                                                        
                                                                                                                                                                                                                        
Note: `insitu_salt_profile_argo` currently uses the temperature slope                                                                                                                                                   
(-0.00045) rather than the salinity slope used by every other salinity stream.                                                                                                                                          
Flagging in case that is unintentional.                                                                                                                                                                                 
                                                                                                                                                                                                                        
### ADT                                                                                                                                                                                                                 
                                                                                                                                                                                                                        
Per-instrument ADT obs spaces are long-standing -- the `rads_adt_*` /                                                                                                                                                   
`adt_rads_*` files are all **symlinks to `adt_rads_all.yaml.j2`**, so the                                                                                                                                               
instruments are separated at the obs-space level while sharing one config.                                                                                                                                              
Nothing changes about that here. What changes is (a) `obs_list.yaml.j2` now                                                                                                                                             
names the eight ObsForge-convention streams (`rads_adt_3a`, `3b`, `6a`, `c2`,                                                                                                                                           
`j2`, `j3`, `sa`, `sw`) instead of the single `adt_rads_all` entry, which also                                                                                                                                          
retires the stale commented-out `adt_rads_sentinel3a`-style names, and (b) the                                                                                                                                          
shared config itself is retuned, below. Because of the symlinks, that retune                                                                                                                                            
applies to all eight streams at once.                                                                                                                                                                                   
                                                                                                                                                                                                                        
Adds **flow-dependent sigma_o**: the error is reduced where the background                                                                                                                                              
surface speed, a geostrophic proxy for the ADT gradient, is large, so the                                                                                                                                               
analysis draws harder to the altimeter in energetic regions. `ObsErrorModelRamp`                                                                                                                                        
decays 0.10 m down to a hard floor of 0.03 m, the along-track altimeter noise                                                                                                                                           
level.                                                                                                                                                                                                                  
                                                                                                                                                                                                                        
| Background surface speed | Assigned sigma_o | Relative weight |                                                                                                                                                       
|---|---|---|                                                                                                                                                                                                           
| <= 0.10 m/s (quiet ocean) | 0.100 m | 1.0x |                                                                                                                                                                          
| 0.25 m/s (western boundary current median) | 0.074 m | 1.8x |                                                                                                                                                         
| >= 0.50 m/s (jet cores) | 0.030 m | 11.1x |                                                                                                                                                                           
                                                                                                                                                                                                                        
The ramp is **deliberately broad and must not be sharpened**: the reduction                                                                                                                                             
lands where the *model* puts the current, not where it actually is, so the                                                                                                                                              
reduced-sigma_o footprint has to cover the energetic surround rather than just                                                                                                                                          
the jet core, wide enough to contain the 50-100 km axis error measured in the                                                                                                                                           
frontal analysis. Measured on 20251217/00 across 7 satellites: 35% of obs keep                                                                                                                                          
0.1 m, 3.6% reach the floor, global mean weight change x2.0. Median assigned                                                                                                                                            
sigma_o by region -- Agulhas 0.048, Brazil/Malvinas 0.062, EAC 0.071, Kuroshio                                                                                                                                          
0.072, ACC-Indian 0.075, Gulf Stream 0.076, quiet North Pacific 0.091.                                                                                                                                                  
                                                                                                                                                                                                                        
The background check tightens from 1.0 m to 0.1 m, and the 500 m minimum depth                                                                                                                                          
domain check is commented out.                                                                                                                                                                                          
                                                                                                                                                                                                                        
**Known gap: no tropical error inflation on ADT.** The Gaussian-in-latitude                                                                                                                                             
inflation described above is applied to all twelve in situ profile streams but                                                                                                                                          
to none of the ADT streams, which currently get only the flow-dependent ramp.                                                                                                                                           
The intent was for ADT to carry it as well; it was simply not added. Two                                                                                                                                                
details to settle when it is: the flow-dependent ramp already reduces                                                                                                                                                   
sigma_o toward its 0.03 m floor in the energetic parts of the tropics, so the                                                                                                                                           
two would need an agreed order and the inflation would have to come last, as                                                                                                                                            
it does in the profile files; and the combined effect where tropical in situ                                                                                                                                            
and ADT overlap should be checked so the region is not down-weighted twice.                                                                                                                                             
Not addressed in this PR.                                                                                                                                                                                               
                                                                                                                                                                                                                        
### SSS                                                                                                                                                                                                                 
                                                                                                                                                                                                                        
SMAP and SMOS L2 activated, both previously commented out.                                                                                                                                                              
                                                                                                                                                                                                                        
## 4. Supporting changes                                                                                                                                                                                                
                                                                                                                                                                                                                        
- `gdas_incr_qc.h`: state bounds are now applied to **any** increment variable                                                                                                                                          
  listed under `state bounds`, instead of being hardwired to temperature and                                                                                                                                            
  salinity. Bounds requested for the sea ice variables were previously ignored                                                                                                                                          
  silently. Malformed entries warn rather than abort.                                                                                                                                                                   
- `gdas_incr_handler.h`: sea ice postprocessing moves from the                                                                                                                                                          
  `VariableChange` path to soca's C++/atlas `PostProcessIce`, which writes a                                                                                                                                            
  CICE-consistent restart directly. Config key renamed                                                                                                                                                                  
  `ice analysis postprocessing` -> `postprocess ice`.                                                                                                                                                                   
- `fields_metadata.yaml`: adds atmospheric forcing fields (winds, humidity,                                                                                                                                             
  downwelling fluxes), ice strain rate diagnostics, ice salinity and                                                                                                                                                    
  `pole_selector`. Note `sea_ice_thickness` / `sea_ice_snow_thickness` now map                                                                                                                                          
  to `hi_div_aice_h` / `hs_div_aice_h` (previously `hicen` / `hsnon`).                                                                                                                                                  
- `jcb-base.yaml.j2`: velocities added to the analysis variables;                                                                                                                                                       
  `bulk_ice_salinity` renamed `sea_ice_salinity`; the deterministic state                                                                                                                                               
  variable list extended with the forcing and diagnostic fields above.                                                                                                                                                  
- `parm/marine/fms/input.nml` removed in favour of develop's `input.nml.j2`.                                                                                                                                            
                                                                                                                                                                                                                        
# Companion PRs                                                                                                                                                                                                         
                                                                                                                                                                                                                        
<!-- global-workflow PR, if any -->                                                                                                                                                                                     
                                                                                                                                                                                                                        
# Issues                                                                                                                                                                                                                
                                                                                                                                                                                                                        
<!-- Refs / Resolves -->                                                                                                                                                                                                
                                                                                                                                                                                                                        
- [ ] C48mx500_3DVarAOWCDA <!-- JEDI low-res marine 3DVar cycled DA  !-->                                                                                                                                               
- [ ] C48mx500_hybAOWCDA <!-- JEDI marine hybrid envar cycled DA !-->              

<!-- Which Global Workflow CI tests are required to adequately test this PR? -->
- [ ] atm_jjob <!-- JEDI atm single cycle DA !-->
- [ ] C96C48_ufs_hybatmDA <!-- JEDI atm cycled DA !-->
- [ ] C96C48_hybatmsnowDA <!-- JEDI snow cycled DA !-->
- [ ] C96_gcafs_cycled <!-- JEDI aerosol cycled DA !-->
- [ ] C48mx500_3DVarAOWCDA <!-- JEDI low-res marine 3DVar cycled DA  !-->
- [ ] C48mx500_hybAOWCDA <!-- JEDI marine hybrid envar cycled DA !-->
- [ ] C96C48_ufsgsi_hybatmDA <!-- JEDI atm Var with GSI EnKF cycled DA !-->
- [ ] C48_ufsenkf_atmDA <!-- JEDI atm EnKF cycled DA !-->
- [ ] C96C48_hybatmDA <!-- GSI atm cycled DA !-->


parm/jcb-gdas/model/marine/marine_background_error_static_diffusion.yaml.j2
```diff
@@ -27,26 +156,48 @@ saber outer blocks:
...
+    - sea_ice_snow_thickness
+    - sea_water_salinity
+    - sea_water_potential_temperature
+    jacobian masking:
+      variable: sea_ice_area_fraction
+      level: 0
+
+  vertical emulators:
+  - name: sea_surface_height_above_geoid
+    path: /scratch3/NCEPDEV/da/Guillaume.Vernieres/sandboxes/saber-pytorch/runs/phy_steric/steric.ts
```
AndrewEichmann-NOAA, line 175: Should this be hardcoded?

parm/jcb-gdas/model/marine/marine_background_error_static_diffusion.yaml.j2
```diff
@@ -27,26 +156,48 @@ saber outer blocks:
...
+    jacobian wrt:
+    - sea_water_potential_temperature
+    - sea_water_salinity
+
+  - name: sea_water_salinity
+    path: /scratch3/NCEPDEV/da/Guillaume.Vernieres/sandboxes/saber-pytorch/emulators/heave_salinity/scripts/heave_salinity.ts
```
AndrewEichmann-NOAA, line 181: Should this be hardcoded?

parm/jcb-gdas/model/marine/marine_background_error_static_diffusion.yaml.j2
```diff
@@ -27,26 +156,48 @@ saber outer blocks:
...
+    jacobian wrt:
+    - sea_water_potential_temperature
+
+  non-local emulators:
+  - name: geostrophy
+    path: /scratch3/NCEPDEV/da/Guillaume.Vernieres/sandboxes/saber-pytorch/emulators/geostrophy/scripts/geostrophy_nonlocal.ts
```
AndrewEichmann-NOAA, line 187: Should this be hardcoded?

parm/jcb-gdas/observations/marine/adt_rads_all.yaml.j2
```diff
@@ -34,39 +34,110 @@
...
+  #- filter: Perform Action
+  #  action:
+  #    name: assign error
+  #    error function:
+  #      name: ObsFunction/LinearCombination
+  #      options:
+  #        variables: [GeoVaLs/mesoscale_representation_error,
+  #                    ObsError/absoluteDynamicTopography]
+  #        coefs: [0.0,
+  #                1.0]
```
AndrewEichmann-NOAA, line 122: Is there expectation of using this again?

parm/jcb-gdas/observations/marine/adt_rads_all.yaml.j2
```diff
@@ -34,39 +34,110 @@
...
+#  - filter: BlackList
+#    where:
+#    - variable:
+#        name: MetaData/latitude
+#      minvalue: 4
+#      maxvalue: 30
+#    - variable:
+#        name: MetaData/longitude
+#      minvalue: 32
+#      maxvalue: 56
```
AndrewEichmann-NOAA, line 140: Is there expectation of using this again?

AndrewEichmann-NOAA (commented):
Nothing critical - I think I'd need a diagram for some of the static B comments

parm/jcb-gdas/model/marine/marine_background_error_static_diffusion.yaml.j2
guillaumevernieres: yes, for now. But is also needs to be commented out and replaced by the soca balance. I'll change this.

parm/jcb-gdas/observations/marine/adt_rads_all.yaml.j2
guillaumevernieres: I hope not. I'll delete also, we can always recover this from the history.

parm/jcb-gdas/model/marine/marine_background_error_static_diffusion.yaml.j2
```diff
@@ -27,26 +156,48 @@ saber outer blocks:
...
-linear variable change:
-  input variables: {{ marine_det_increment_variables }}
-  output variables: {{ marine_det_increment_variables }}
-  linear variable changes:
-  - linear variable change name: BalanceSOCA
+- saber block name: TorchBalance
+  save jacobians: true
+  surface emulators:
+  - name: sea_ice_area_fraction
+    path: /scratch3/NCEPDEV/da/Guillaume.Vernieres/sandboxes/saber-pytorch/runs/ml_aice/ml_aice_dual.ts
```
JohnSteffen-NOAA: This can only be used on Ursa then since these paths are hardcoded?

parm/marine/fields_metadata.yaml
AndrewEichmann-NOAA, line 154: Delete duplicate entry

```suggestion
```

AndrewEichmann-NOAA, line 179: Delete duplicate entry
```suggestion
```

AndrewEichmann-NOAA (commented):
Couple of duplicate entries

shlyaeva:
@guillaumevernieres: is `SeaIceFreeboard` obs operator in ufo develop? I don't see it there, is it possible that it's still in your branch?

guillaumevernieres:
> @guillaumevernieres: is `SeaIceFreeboard` obs operator in ufo develop? I don't see it there, is it possible that it's still in your branch?

Not yet, I'll try to prepare the freeboard UFO PR this week.

AndrewEichmann-NOAA:
Cycles with C384mx025_3DVarAOWCDA, assuming g-w 93d5916805e6cc5b85c3dca40a727fe800f9cb33, the proper combo of dates, ICs, and obs, and with torch balance turned off per Guillaume

PR #2257, opened 2026-09-21
Bugfix: Fix elevation based error inflation code for snow var DA config/yaml file

### LINKED ISSUE #2251: Div by zero error in snow elevation adjustment
A div by zero error in JEDI snow CI test (in Global Workflow) caused by snow elevation adjustment

Image

Likely after recent (do not know which one) updates in UFO.

Not sure about the cause. Testing by adding a guardrail for when elvation diff < small value did not fix it. Need a thorough look and fix.

tsga:
# Description

This PR fixes the bug in the station snow DA yamls, which had erroneous name used for coefficients in ObsFunction/Arithmetic, that led to a div/0 error.  Also added the option "abort if invalid operation: true" (default is false) so that any invalid operation like Div/0 lead to early program exit. And lower and upper bounds to the final inflation factor were added.

# Companion PRs

# Issues

Resolves #2251 

- [ ] atm_jjob 
- [ ] C96C48_ufs_hybatmDA 
- [x] C96C48_hybatmsnowDA 
- [ ] C96_gcafs_cycled 
- [ ] C48mx500_3DVarAOWCDA 
- [ ] C48mx500_hybAOWCDA 
- [ ] C96C48_ufsgsi_hybatmDA 
- [ ] C48_ufsenkf_atmDA 
- [ ] C96C48_hybatmDA


parm/jcb-gdas/observations/snow/snomad.yaml.j2
```diff
@@ -150,49 +150,10 @@
...
-      filter variables:
-      - name: totalSnowDepth
-      where:
-      - variable:
-          name: ObsValue/totalSnowDepth
-        minvalue: 0.
-      action:
-        name: inflate error
-        inflation variable: DerivedMetaData/elevation_diff_exp_inv # 2.0
+## obs error inflation for obs-model elevation difference
```
ClaraDraper-NOAA, line 153: Using the same yaml each time will be great for reducing errors / code.

parm/jcb-gdas/observations/snow/station_obs_error_inflation.yaml.j2
```diff
@@ -0,0 +1,57 @@
...
+- filter: Variable Assignment
+  assignments:
+  - name: DerivedMetaData/elevation_infl_factor
+    type: float
+    function:
+      name: ObsFunction/Exponential
+      options:
+        variables:
+        - DerivedMetaData/elevation_diff_func
+        coeffA: 1.0
```
ClaraDraper-NOAA, line 26: I diffed the old and new code - it looks like the coefficients here have changed from [1 -1 0 1] to [1 1 0 3], and we no longer do an explicit inverse step. Can you please provide some details of what this change does?

parm/jcb-gdas/observations/snow/station_obs_error_inflation.yaml.j2
```diff
@@ -0,0 +1,57 @@
...
+        coeffB: 1.0
+        coeffC: 0.0
+        coeffD: 3.0
+- filter: Variable Assignment
```
ClaraDraper-NOAA, line 30: The other change is limiting the error inflation to be [1,3]. 1 makes sense to me, but 3 seems low? If we have obs that are very far from the model, I'd want to downweight them by a large amount.

parm/jcb-gdas/observations/snow/station_obs_error_inflation.yaml.j2
```diff
@@ -0,0 +1,57 @@
...
+      name: ObsFunction/Arithmetic
+      options:
+        variables:
+        - MetaData/stationElevation
+        - GeoVaLs/filtered_orography
+        coefs:
+        - 1.0
+        - -1.0
+        total exponent: 2
+        total coefficient: 1.5625e-06
```
ClaraDraper-NOAA: Can you add an explanation in the comments for this number? (1/(800*800), right?)

ClaraDraper-NOAA (changes requested):
Looks good. I like this solution of having a separate yaml with the inflation details.

parm/jcb-gdas/observations/snow/station_obs_error_inflation.yaml.j2
tsga: Done

parm/jcb-gdas/observations/snow/station_obs_error_inflation.yaml.j2
tsga: I have now changed to 5, but with comment that it is unlikely to matter in this case, because the elevation_diff filter is excluding obs with >800 diff so the values would be limited to ~2.7 (e)

parm/jcb-gdas/observations/snow/station_obs_error_inflation.yaml.j2
tsga: I have now added a comment there.

ClaraDraper-NOAA: Sorry I should have been more explicit. we don't really need the comment here, I was more just looking for an explanation in the github discussion of what the initial bug was and how it's been fixed.

parm/jcb-gdas/observations/snow/station_obs_error_inflation.yaml.j2
```diff
@@ -0,0 +1,57 @@
...
+      name: ObsFunction/Arithmetic
+      options:
+        variables:
+        - MetaData/stationElevation
+        - GeoVaLs/filtered_orography
+        coefs:
+        - 1.0
+        - -1.0
+        total exponent: 2
+        total coefficient: 1.5625e-06    # 1/(800*800: assuming vertical (exponential decay) scale of 800
```
ClaraDraper-NOAA: add unit. 800 m right?

parm/jcb-gdas/observations/snow/station_obs_error_inflation.yaml.j2
ClaraDraper-NOAA: Up to you, but 3 was fine - perhaps more intuitive, since is't ~2.7.

parm/jcb-gdas/observations/snow/station_obs_error_inflation.yaml.j2
tsga: I have added a sentence above, but this change is not really about the bug. It tries to simplify the yaml coding by reducing one block (because div by exp(-1) is equivalent to multiply by exp(1))

